PR #42, opened 2015-10-11
New image and display implementations with much improved RAM use.

markshannon:
This PR has three main goals.

1. Improve the memory impact of Images by a large factor, by
    * not using the DAL heap when manipulating images or the display
    * minimizing the size of images.
2. Adding missing functionality to image and display.
3. Use a more Pythonic (and mostly functional) style for the display and image API.

To achieve these goals this PR:

1. Reduces memory use by:
    * Making no reference to the DAL's MicrobitImage class in the implementation of our Image class and does not use the DAL's ManagedString when displaying strings. This reduces the  interference between the DAL's C++ heap and MicroPython's GC heap.
    * Reducing the size of constant 5x5 images to a mere 8 bytes, and the size of other images to 8 bytes plus 1 byte for every two pixels.
2. Image gains crop and new constructors for creating an abritrary size blank image and creating an image for a single character. The display interface remains much the same, except that it looses the set_brightness() and set_mode() methods. This are now useless (and confusing) as the brightness is now controlled on a per-pixel basis.
3. Most methods on Image become "functional" in that they return a new image, rather than modifying the current image. Apart from being a better interface, this allows immutable images and mutable images to have the same interface.
   E.g. Image.crop() returns a new image which means that `Image(...).crop(...)` and `Image.HEART.crop(...)` act alike.
   The obvious exception to this is `set_pixel()`, which raises an exception for immutable images.

It is worth noting that all the more fancy functionality of display, such as display.animate and display.scroll is based around the concept of `display.animate()` taking an arbitrary iterable of images and/or single character strings.
For example `display.print()` for long strings is easily implemented by taking advantage of the fact that a string is an iterable of single character strings,
and passing it to animate.
   
#### Performance.

The following program which creates 100 000 images while keeping a rolling 100 of them live (to stress the GC a bit) takes about 90 seconds.
```python
l = [None] * 100                        
for i in range(1,1001):                 
    for j in range(100):                
        l[j] = Image.HEART.crop(0,0,5,5)
    print(i*100)                        
```
Given that, previously we couldn't even create more than 45 images while keeping one alive, this has to be considered an improvement :smile:


examples/conway.py
```diff
@@ -10,9 +10,11 @@
 arena2 = bytearray(7 * 7)
 
 def show():
+    img = microbit.Image(5,5)
     for y in range(5):
         for x in range(5):
-            microbit.display.image.set_pixel(x, y, arena1[8 + y * 7 + x])
+            img.set_pixel(x, y, arena1[8 + y * 7 + x]*9)
+            microbit.display.print(img)
```
markshannon: Note to self. Don't edit code when half asleep :disappointed: 
Fixed.

markshannon:
OK, after much experimentation, it would appear that if you set a `Timeout` with a very short delay (32 or 64 us) and don't subsequently increase the delay,  then it does strange things to interrupts. I suspect what happens is there is not long enough to detach the `Timeout` from the event Q, before the timer interrupt triggers again, even though the `TImeout` function is not called again.

So, I've reverted to the multiple, exponentially increasing delay interrupt model.
I've shaved off as many cycles as I can from the interrupt, it now feels fairly responsive.
The REPL seems unaltered by greyscale mode.

It is still impossible to have full speed UART and greyscale rendering at the same time (but that is also true of the DAL-based master branch).

If we do want greyscale rendering & fast comms at the same time, then I think that we will need to play with interrupt priorities.

markshannon:
GitHub doesn't seem to be picking up my latest changes. Probably due to making the repos public.

PR #46, opened 2015-10-20
Create the basic configuration for docs using Sphinx

deshipu:
Also copy the "API Design Heuristics" from the wiki as an example.


deshipu:
Sorry, I fail at using github, this is a duplicate pull request.

PR #47, opened 2015-10-20
Add LED dance example

jrmhaig:


dpgeorge:
Thanks @jrmhaig!  Merged in 1904e1987114d9c57602b5c1fdb41a8725cdb090 with small change to use "set_pixel" instead of "set_pixel_value" (the method has been renamed).

The "random" function will hopefully be fixed soon, so it's properly random.

PR #52, opened 2015-10-20
New image and display implementations with much improved RAM use

markshannon:
Rebased and reopened version of:
https://github.com/dpgeorge/microbit-micropython/pull/42


dpgeorge:
> It is still impossible to have full speed UART and greyscale rendering at the same time (but that is also true of the DAL-based master branch).

I liked your solution where the display logic only used 1 interrupt per frame if the pixels were either "0" or "9".  Does the new code have this property?

Maybe grayscale is too CPU/power intensive and shouldn't be enabled by default?  But then we'd be back to having to provide B&W and grayscale modes, and a function to switch between them, which is not ideal from the user perspective.

markshannon:
If all pixels are 0 or 9, then there is only one interrupt per 6ms.
It is impossible to get pyboard to work (unless you drop the bitrate) otherwise. The same applies for master; if you turn on greyscale by default, then pyboard doesn't work.

dpgeorge:
I'm getting an error: uBit.display.renderTimer is private.  Maybe the latest version of the DAL has changed?  (yt update)

Would be good not to need to make a new instance of this timer, since it's RAM hungry.  One hack to access the private renderTimer is to take the address of uBit.display and add the correct offset to get to renderTimer.  It's a hack, but I think a clean hack.

dpgeorge:
The way you can combine images is very cool, and I think in general the API is clean and simple, yet powerful.

The only thing I don't agree with is the str/repr of images.  I thing `str(img)` should return a string that's valid input to `Image(...)`, ie the constructor.  So that means removing the border around the str/repr.

markshannon:
Generally `repr(x)` is supposed to return a string that is a valid input for the initializer, but there are many exceptions.There is no such rule for `str(x)` which should return a human readable form.

Does micropython support both `repr` and `str` or they treated the same?

I have found the border quite helpful for looking at unusual sized images with blank pixel near the edge.
Without the border it is not clear where the image stops.

dpgeorge:
> There is no such rule for str(x) which should return a human readable form.

With the newlines I think it's quite readable.  I take your point about having borders to distinguish blank pixels, but that can be fixed by printing "0" (or ".") for blank pixels.

> Does micropython support both repr and str

Yes!  The `mp_print_kind_t kind` variable is either PRINT_STR or PRINT_REPR (well, there's also PRINT_JSON and PRINT_EXC, but ignore those; if it's not PRINT_STR assume it's PRINT_REPR).

I think @ntoll will have an opinion here, because he was advocating for str to print exactly what the kids would need to enter to create such an image, so they could make their own by looking at existing (eg builtin) images (and learn the format of the image string constructor by exploration).

markshannon:
Ok, you've convinced me that the constructor should take the same format as print produces.
With quotes and implicit string concatenation, I think I can still produce a readable form.

markshannon:
I've reimplemented __str__ and __repr__ as below:
```python
>>> Image.HEART
Image(' 9 9 \n99999\n99999\n 999 \n  9  \n')
>>> print(Image.HEART)
Image(
    ' 9 9 \n'
    '99999\n'
    '99999\n'
    ' 999 \n'
    '  9  \n'
)
```
Both are identical as far as the parser is concerned, the only difference is ease of comprehension.

markshannon:
I've added an unpleasant hack to work around the DAL API breakage :disappointed:

dpgeorge:
> I've added an unpleasant hack to work around the DAL API breakage :disappointed:

It's strange that it originally worked, because it does seem that renderTimer has always been private...??

markshannon:
ping

dpgeorge:
There are a lot of self-inconsistencies in the coding style, and also inconsistencies with the existing code-base.  We aim for https://github.com/micropython/micropython/blob/master/CODECONVENTIONS.md

For example, in microbit_display_obj_t::advanceRow:

- some comments use /* */, some use //
- some comments have a space after //, some don't
- some comments end in a fullstop, others don't
- some comments start with a capital, some lowercase
- should have a space between "if" and the opening parenthesis
- the "for" block should have the opening brace on the same line as the for
- the condition in the "for" block should have spaces around the "<"
- trailing whitespace on some lines
- why a blank line at the end of the function?

markshannon:
The wonky formatting is a result of pulling in code from various sources.
I've tidied it up.

I'd like to get this merged soon, so that it gives people a chance to use the new API.

ntoll:
Can't wait... it'll mean we can also merge in @marichkakorolyuk's wonderful image based work too. :-) Rabbits FTW!

markshannon:
Where are the rabbits?

ntoll:
https://github.com/dpgeorge/microbit-micropython/files/3187/more-images.txt :-)

dpgeorge:
Merged, first commit 57a4b19dd09fa9e84a12ecb40ccbf4acd09b02fd, last commit 66dec5b24177d5382ea6810ddce70c4afbcfe115.

Thanks @markshannon!

markshannon:
@dpgeorge  How come Github doesn't register your merges, as such? Don't you use the big green button?

dpgeorge:
No, I do it by hand and rebase (it's a habit to keep the git history nice and linear).  And so the hashes change and github can't determine that it was merged.  In this case a rebase was not really needed, but I did take the chance to format your commit messages to fit in the 73 char (soft) limit.  Hope you don't mind :)

PR #54, opened 2015-10-21
Add toy compass example

markshannon:
Very simple compass example.

### DIFF
--- a/examples/compass.py
+++ b/examples/compass.py
@@ -4,18 +4,25 @@ needles = (Image.CLOCK12, Image.CLOCK1, Image.CLOCK2, Image.CLOCK3,
            Image.CLOCK4, Image.CLOCK5, Image.CLOCK6, Image.CLOCK7, 
            Image.CLOCK8, Image.CLOCK9, Image.CLOCK10, Image.CLOCK11)
 
-#You need to spin the microbit about a few times to help calibrate the compass
+# You need to spin the microbit about a few times to help calibrate the compass
+
+# Scroll a message to encourage user to help calibrate the compass
+# "Spin" "Me!" reads better than "Spin Me!" due to lack of kerning
+display.scroll("Spin")
+display.scroll("Me!")
+
+# Start calibrating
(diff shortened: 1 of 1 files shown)


dpgeorge:
That's cool, it actually works! (ie points north)

markshannon:
If it works, then perhaps it should be merged :wink:

examples/compass.py
```diff
@@ -0,0 +1,21 @@
+from microbit import *
+
+needles = (Image.CLOCK12, Image.CLOCK1, Image.CLOCK2, Image.CLOCK3,
+           Image.CLOCK4, Image.CLOCK5, Image.CLOCK6, Image.CLOCK7, 
+           Image.CLOCK8, Image.CLOCK9, Image.CLOCK10, Image.CLOCK11)
+
+#You need to spin the microbit about a few times to help calibrate the compass
```
willingc: @markshannon Cool example. Perhaps for user friendliness in the example, display a message to indicate "Calibrating..." before the while loop output.

examples/compass.py
```diff
@@ -0,0 +1,21 @@
...
+while compass.is_calibrating():
+    for i in range(12):
+        display.print(needles[i])
+        sleep(80)
+display.clear()
+
+#Try to keep needle pointed in (roughly) the same direction
+while True:
+    sleep(500)
+    needle = ((15-compass.heading())//30)%12
```
willingc: Since new python users might use this example, it might be good to be more pythonic on the line spacing:
``needle = ((15 - compass.heading()) // 30) % 12``

Also on the comments, add a space after the #

markshannon:
Done.

dpgeorge:
Merged in cf309e98c7bd0a09b2358ddfa33692a52dacdefd.

ntoll:
Apparently the correct way to calibrate the compass is to turn it flat on the spot (without tilting it otherwise that will add inaccuracies). We should mention this somewhere.

PR #55, opened 2015-10-23
Update music example in tune.py to make it work correctly.

ntoll:
This branch does exactly what it says so it works with the current music API.


dpgeorge:
tune is now named play, and pin arg is second and optional.

ntoll:
OK... updated.

dpgeorge:
Squashed and merged in a240daa9ecc590760f6a29ec3953e1582e76b813.

PR #56, opened 2015-10-25
Adds python styling to README

frankcash:
README didn't have python formatting so I added it to the markdown!


markshannon:
I'm not sure about the highlighting. The prompt (`>>>`) is rendered the same as `import`.
I'm guess that the prompt is rendered as three `>`s. Is it possible to mark them in bold or something?

If not, then it is better than highlighting at all IMO.

PR #58, opened 2015-10-29
Drop 'fibers' and convert to conventional MicroPython polling model.

markshannon:
This PR:
Adds `mp_hal_sleep` which uses `__WFI` not `fiber_sleep`
Converts all uses of `uBit.sleep` to `mp_hal_sleep`
Only updates the accelerometer and compass when read and only if needed.

This means that the following code now works correctly:
```python
while True:
     print(accelerometer.get_x())
```
Printing varying values as the micro:bit is moved about.

This should save power, as sleep does much less work on interrupt, and the  accelerometer and compass are polled far less often. However, I lack the equipment to verify this.

As a bonus, I've added `accelerometer.get_values()` so you can type
`x, y, z = accelerometer.get_values()`


dpgeorge:
Ok, so now we are really moving away from the DAL.  On the one hand we need to be resource conscious and fine tune the event handler exactly to our needs.  On the other hand it's tricky to debug this stuff and that's already been done in the DAL.

Instead of `mp_hal_sleep` we should use `mp_hal_delay_ms` since this is what uPy is now using, eg https://github.com/micropython/micropython/commit/404dae80a9d5cb6315071fe6206b2a6026be0a09

We should decide if we want to keep the accelerometer accessors as `get_x()`, or switch to something like `x()`.

source/microbit/microbitaccelerometer.cpp
```diff
@@ -36,28 +36,57 @@ typedef struct _microbit_accelerometer_obj_t {
     MicroBitAccelerometer *accelerometer;
 } microbit_accelerometer_obj_t;
 
+bool accelerometer_up_to_date = false;
+
 mp_obj_t microbit_accelerometer_get_x(mp_obj_t self_in) {
     microbit_accelerometer_obj_t *self = (microbit_accelerometer_obj_t*)self_in;
+    if (!accelerometer_up_to_date) {
+        self->accelerometer->idleTick();
```
dpgeorge: What happens if a systemTicker interrupt occurs while idleTick is running, and then systemTicker calls idleTick?  I don't think idleTick is reentrant.

larryhastings:
How about just .x ?  Does MP support descriptors in C?

dpgeorge:
> Does MP support descriptors in C?

Yes.

> How about just .x ?

We discussed this at length at PyCon UK: we use functions for everything.  It's more consistent (and setters like obj.attr = 1 can't be used in a lambda, while obj.set_attr(1) can).

ntoll:
I seem to remember having a conversation about `get_foo()` v `foo` or `foo()` at PyCon UK. IIRC we agreed to keep such things as method calls since that's a consistent pattern in the way the API is written. I also remember a discussion about using explicit verbs in the method name so kids immediately understood what it was they were calling. I.e. `get_foo()` is more obvious than simply just `foo()`.

Or am I mis-remembering..???

ntoll:
Aha... thanks @dpgeorge - I wasn't imagining things.

larryhastings:
Let me get this out of my system once and then I'll (try to remember to!) button my lip about it.

I'm all for consistency, but in this case I think the group has chosen to be consistent in the wrong way.  Things like this are *exactly* what the descriptor protocol is for.  The spelling "accelerometer.x" couldn't be clearer, not to mention easy to read and pleasingly short to type.  I would absolutely design my API that way for adults and I think it's better for kids too.

It's true that you can't use an assignment statement in a lambda function.  But:
a) one shouldn't warp one's API design to accommodate lambda functions,
b) accelerometer.x would be read-only anyway,
c) I doubt the curriculum for 11-year-olds will include lambda, and
d) I propose we strenuously avoid callback-based APIs.  ("I don't like callback interfaces, I always get them wrong" --GvR (quote is approximate))  Iteration > callbacks.

I'd *greatly* rather see us make properties of objects behave like simple attributes, using descriptors where necessary.  I suggest that's far more Pythonic--accessor functions always feel like Java or C++ to me.

So, if it were me, I'd use descriptors nearly everywhere.  I'd reserve .set_x() accessors for APIs that are actively changing something (e.g. microbit.display.set_display_mode).  And I still might use a read-only descriptor instead of a .get_x() method by convention.  Though I'd have to think about that one.

markshannon:
Please, can we not bury this PR in a discussion about `.get_x()` vs `.x()` vs `.x`
The mailing list is the place for such a discussion.
(And updating the API belongs in another PR)

larryhastings:
Apologies, I'll shift to the mailing list as you suggest.

source/microbit/microbitaccelerometer.cpp
markshannon: Good point, I've made the accelerometer and compass update functions safe against interrupt.

markshannon:
I've made the compass and accelerometer methods safe for interrupts and renamed `mp_hal_sleep` to `mp_hal_delay_ms`

source/microbit/main.cpp
```diff
@@ -30,3 +30,37 @@ void app_main() {
...
+    microbit_display_tick();
+}
+
+extern "C" {
+
+void microbit_init(void) {
+    microbit_display_init();
+
+    // Hijack the DAL system ticker.
+    uBit.systemTicker.attach(ticker, MICROBIT_DISPLAY_REFRESH_PERIOD);
```
dpgeorge: Is the DAL systemTick still being called?  If so, what does it do?  Does it update the buttons as well?

markshannon: With this PR we no longer calling the DAL's `systemTick`. 
The entry points to the DAL fiber scheduler were `systemTick()`, fiber_sleep()` and `schedule()`. This PR drops all references to those functions, ensuring that MicroPython, and not the DAL, now controls what runs and when.

dpgeorge:
Merged!

Issue #59, opened 2015-10-30
Document the correct way to callibrate the compass


ntoll:
Apparently, the correct way to calibrate the compass is to turn it flat on the spot (without tilting it otherwise that will add inaccuracies). We should document this in an obvious place.

willingc:
@ntoll Perhaps the easiest way to document initially is in an introductory docstring block:

```
"""
    compass.py
    ~~~~~~~~~

    Creates a compass. To calibrate the compass, lie it flat and spin it horizontally. For 
    best results during calibration, try not to tilt the compass up and down.

"""
```

ntoll:
+1 - also there should be something in the api so `help(microbit.compass.callibrate)` returns something useful that an 11yo kid would understand.

ntoll:
Docstring added, the `help` related text still doesn't reference how to calibrate it. This will be fixed by #125.

PR #60, opened 2015-10-30
Add a general info docstring to compass example #59

willingc:
This adds a user friendly docstring to the top of the compass example (nicely done @markshannon) #59 #54. 

As an FYI, it follows the clean style of the Flask and Requests projects.

Minor change: removed a trailing space

### DIFF
--- a/examples/compass.py
+++ b/examples/compass.py
@@ -4,7 +4,7 @@
 
     Creates a compass.
 
-    To calibrate the compass, lie it flat and spin it horizontally. For best
+    To calibrate the compass, lay it flat and spin it horizontally. For best
     results during calibration, try not to tilt the compass up and down.
 """
 from microbit import *
--- a/examples/compass.py
+++ b/examples/compass.py
@@ -1,7 +1,16 @@
+"""
+    compass.py
+    ~~~~~~~~~~
+
+    Creates a compass.
+
+    To calibrate the compass, lie it flat and spin it horizontally. For best
+    results during calibration, try not to tilt the compass up and down.
+"""
 from microbit import *
 
 needles = (Image.CLOCK12, Image.CLOCK1, Image.CLOCK2, Image.CLOCK3,
-           Image.CLOCK4, Image.CLOCK5, Image.CLOCK6, Image.CLOCK7, 
+           Image.CLOCK4, Image.CLOCK5, Image.CLOCK6, Image.CLOCK7,
            Image.CLOCK8, Image.CLOCK9, Image.CLOCK10, Image.CLOCK11)
 
 # You need to spin the microbit about a few times to help calibrate the compass


examples/compass.py
```diff
@@ -1,7 +1,16 @@
+"""
+    compass.py
+    ~~~~~~~~~~
+
+    Creates a compass.
+
+    To calibrate the compass, lie it flat and spin it horizontally. For best
```
blmorris: Minor grammatical question - should this be "lie it flat" or "lay it flat"? The latter sounds better to my ears, but I can't say I'm certain which is correct.

willingc: @blmorris You are absolutely correct. It is "lay it flat". Updating. Great catch.

sparkslabs: http://blog.oxforddictionaries.com/2012/03/lie-or-lay/ suggests "lay it
flat" is correct. I'd probably say "lie it flat". I suspect an 11 year old
might too.

Michael.

willingc: @sparkslabs An opportunity to share grammar with 11 year olds ;)

markshannon:
LGTM

willingc:
Thanks @markshannon. Would you mind if I added a header docstring to some of the other examples?

markshannon:
Please do. But don't forget that memory is somewhat precious, so try to keep them short.

dpgeorge:
> But don't forget that memory is somewhat precious, so try to keep them short.

It would actually be beneficial to make them comments starting with # rather than doc strings, since comments are completely ignored by the compiler, whereas doc strings must be copied into RAM.  If the doc strings are short (like < 64 chars) then it's not really a problem.  But longer than that and it can start to have a detrimental effect.

I know it's more Pythonic to have doc strings, and for these examples it won't stop them working, but it's something to keep in mind / think about.

[Lexer would need to be context sensitive to discard doc strings straight away.  And even if it was context sensitive (which is really hard) it would still not be possible to discard without breaking strict Python semantics; eg, file might start with : `"""doc string""" + 1` which should lead to a TypeError at runtime.]

willingc:
@dpgeorge Thanks for a bit more insight on lexer. How would you feel about a compromise approach that mixes the # comment with docstrings for the header comment? For example,

```
# """
#     compass.py
#     ~~~~~~~~~
#
#     Creates a compass.
#
#     To calibrate the compass, lay it flat and spin it horizontally. For best
#     results during calibration, try not to tilt the compass up and down.
# """
```

Since one of the nice things about the docstrings are the Sphinx tools and extensions that are helpful with documentation. It would likely possible to write a custom extension for Sphinx that could convert the examples to documentation that will be helpful to end users.

By creating a simple script that could strip out the `#` marks from the `# """` strings and the `#` marks in between, someone using the examples on a computer other than the microbit could easily convert to a more common Python program. This may be helpful for education as the students progress beyond the microbit. 

Thoughts?

dpgeorge:
@willingc I think comment + doc string is pretty ugly.  It teaches people the wrong thing (those new to Python) and confuses those familiar with it.

I'd rather just keep it as a concise doc string (like you have it).

willingc:
@dpgeorge Thanks for the feedback. OK, concise doc string will be the approach (with a goal of conserving code space) :sunny:

ntoll:
@willingc there is also a bunch of `help(foo)` related things that'll need updating in `source/microbit/help.c`. You'll see I'd made a start from about a month or so ago (but this work is now out of date). I suggest nothing be done until the API is "frozen" / finished and then use the same modus operandi: reading age of 11, short, simple and direct explanations. :-)

dpgeorge:
Merged in 31c0259a5392bbf5b129ed072bf34c56649f5428.  Thanks @willingc!

willingc:
Thanks @dpgeorge :ocean:

PR #62, opened 2015-10-31
Improve legibility of scrolling text.

markshannon:
Use proportional spacing and slowing down scrolling.
In combination, this means that text takes about the same time to scroll, but is more legible.

Add `monospace` option to `display.scroll()` for those who like ugly mono-spaced text :wink:


dpgeorge:
Awesome!  Merged in 1437aeaa08f915d8750b47a5bc68ad51e6aba90c.

Issue #21, opened 2015-09-18
Help text is longer than 25 lines


dpgeorge:
We want the help text to be minimal and fit on a small terminal screen (eg 25 lines).

We can shorten the current help text and provide pointers to further places to look for help.

One way of doing this is allow the user to type `help('some_topic')` and then it prints specific help for that topic.  The main help would tell you what topics are available.  This is how CPython works.

Eg we could have `help('modules')` to list available modules (same as CPython).  And `help('demo')` to list some quick demo scripts.

willingc:
I have started to work on this issue in my fork of this repo.

My initial approach is to break up the long initial welcome message into smaller messages that could be combined as desired later. I will begin working from @dpgeorge's approach of ``help('some_topic')``.

willingc:
@dpgeorge @ntoll @markshannon I want to clarify some semantics when it comes to the words 'screen' and 'display' since the words pop up in the API, examples, and issues. My current understanding, and please do correct me, is that typically people refer to:

display => the LED grid on the microbit
screen => the higher level term for content on the LED grid on the microbit
terminal screen => an external monitor or output display; info is sent from microbit via serial connection or BLE to an external display/mobile device/computer

It's pretty easy in docs, comments, and conversations to use 'screen' and 'display' interchangeably. From the API perspective, is 'display' limited to the hardware LED grid on the microbit? Where 'screen' is informally used to refer to any visual output device?

larryhastings:
Personally I haven't been consistent about this, and your comment suggests to me that we should definitely be consistent.

Do you think we need this conceptual distinction between "display" and "screen"?  My suspicion is, no.  We should have one term, for the 5x5 LED matrix, and I nominate "display" as superior to "screen" here.

A side note: is MicroPython on the micro:bit going to permit use of BLE?  If not, we probably shouldn't bother to mention it (except to say that it isn't supported).  Also, I hadn't realized BLE allowed connecting up a "display" like that.  Does it permit a REPL?  Is it conceptually a serial device, or something else?

sparkslabs:
Hi,


For the microbit - I've always used the term 5x5 LED Display - from the
original implementation and the reference specification docs (display
management etc) so definitely display. Before that in the original
requirements we always described it as a display.

Content displayed on the display is an Image.

(These are concepts which are valid and used across all the interfaces, and
should be consistent across those IMO - otherwise we're being different for
the sake of being different)

[...]

PR #65, opened 2015-11-03
Add consistency and minimize jargon in help text

willingc:
Relates to #21.

This is a rough draft pass at editing the individual help messages. I'm open to suggestions from others that: 1) would benefit teaching an 11 year old or 2) offer greater precision in user friendly language.

Some tips on teaching and jargon.
http://csteachingtips.org/browse-all?search_api_views_fulltext=help+messages

While common and familiar to adult developers, some of the existing help text is imprecise and is a bit heavy on jargon vs. real world language.


ntoll:
This is good! :-) Many many thanks. I like the link for teaching jargon too.

There are some phrases I'd change to simplify even more. I'm at "real work" [tm] at the moment, but will fill in suggestions over my lunch time.

I think it wonderful that this is progressing. We should also aim for completeness too.

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
-    {&microbit_running_time_obj, "running_time() -- get the number of milliseconds since reset.\n"},
-    {&microbit_accelerometer_obj, "A toolbox of code to detect the device's position (orientation).\n"},
-    {&microbit_accelerometer_get_x_obj, "get_x() -- get the device's X axis reading.\n"},
-    {&microbit_accelerometer_get_y_obj, "get_y() -- get the device's Y axis reading.\n"},
-    {&microbit_accelerometer_get_z_obj, "get_z() -- get the device's Z axis reading.\n"},
-    {&microbit_button_a_obj, "Represents button A. Use its is_pressed() method to discover its status.\n"},
-    {&microbit_button_b_obj, "Represents button B. Use its is_pressed() method to discover its status.\n"},
-    {&microbit_button_is_pressed_obj, "is_pressed() -- True if the button is pressed, else False.\n"},
-    {&microbit_compass_obj, "A toolbox of code to detect the device's heading.\n"},
+    {&microbit_module, "A group of tools to control the micro:bit hardware.\n"},
```
ntoll: I'm in two minds about "tools" etc. How about just "Useful stuff to control the micro:bit hardware." (I'm intentionally using "stuff" since any UK11yo will understand how I'm using it).

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    // System state objects
+    {&microbit_panic_obj, "Begin panic() mode (requires a reset).\n"},
```
ntoll: I like how the usage is now part of the sentence rather than at the start. :-)

Trying to work out how we can explain why one would need to go into panic mode. How about... "Go panic() when something is wrong (requires a reset)."

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_random_obj, "Return a random(number) between 0 and 'number - 1'.\n"},
+    {&microbit_sleep_obj, "Wait for sleep(time) 'time' milliseconds (1 second = 1000 ms).\n"},
+    {&microbit_running_time_obj, "Return running_time() in milliseconds since micro:bit's last reset.\n"},
+    // Accelerometer 3D orientation
+    {&microbit_accelerometer_obj, "Detect the microbit's position (3D orientation).\n"},
+    {&microbit_accelerometer_get_x_obj, "Return micro:bit's X axis position.\n"},
```
ntoll: "the micro:bit's" or just "micro:bit's"... consistency issue perhaps???

I like dropping the "the" because it makes the device sound more friendly (to my ears at least). :-)

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_accelerometer_get_y_obj, "Return micro:bit's Y axis position.\n"},
+    {&microbit_accelerometer_get_z_obj, "Return micro:bit's Z axis position.\n"},
+    // Pushbutton
+    {&microbit_button_a_obj, "micro:bit's 'A' button. is_pressed() is True when button is pressed down.\n"},
+    {&microbit_button_b_obj, "micro:bit's 'B' button. is_pressed() is True when button is pressed down.\n"},
```
ntoll: Typo: please remove full-stops between "button. is_pressed()" -> "button is_pressed()" for both button's A and B.

ntoll: Ahha... my confusion. Ignore the above. :-)

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_button_is_pressed_obj, "is_pressed() -- True if the button is pressed down, else False.\n"},
+    // Compass 3D direction heading
+    {&microbit_compass_obj, "Detect the micro:bit's compass heading (direction faced).\n"},
```
ntoll: "direction faced" sounds clunky. How about, "the direction to which the micro:bit is pointing"

source/microbit/help.c
willingc: @ntoll Actually, I think your confusion makes me think we should change to:

``micro:bit's 'A' button. When button is pressed down, is_pressed() is True.\n``

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
-    {&microbit_compass_is_calibrating_obj, "is_calibrating() -- True if the compass is adjusting itself for accuracy.\n"},
-    {&microbit_compass_clear_calibration_obj, "clear_calibration() - re-sets the compass to an un-calibrated state.\n"},
-    {&microbit_compass_get_x_obj, "get_x() -- get the device's X axis reading.\n"},
-    {&microbit_compass_get_y_obj, "get_y() -- get the device's Y axis reading.\n"},
-    {&microbit_compass_get_z_obj, "get_z() -- get the device's Z axis reading.\n"},
-    {&microbit_display_obj, "A toolbox of code to work with the device's 5x5 display.\n"},
-    {&microbit_display_print_obj, "print(s) -- print the string 's' to the display. E.g. print('bit').\nprint(s, i) -- print string 's', one character at a time with a delay of 'i'.\n"},
-    {&microbit_display_scroll_obj, "scroll(s) -- scroll the string 's' from left to right on the display.\nscroll(s, i) -- scroll string 's' with delay 'i' between characters.\n"},
-    {&microbit_display_clear_obj, "clear() -- clear the display.\n"},
+    {&microbit_compass_calibrate_obj, "calibrate() -- Starts the steps to adjust the micro:bit's compass for accuracy\n"},
```
ntoll: OK... so usage is now back at the start of the definition.

source/microbit/help.c
willingc: I considered using "stuff" but had wondered if too informal for UK ;)

source/microbit/help.c
ntoll: How about, "You must calibrate() micro:bit's compass to make it accurate."

source/microbit/help.c
willingc: Let me think on this one. I want to evoke the computer, not the programmer, into panic mode.

source/microbit/help.c
willingc: Agreed. Will drop "the"

source/microbit/help.c
```diff
@@ -81,57 +81,64 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_compass_is_calibrating_obj, "is_calibrating() -- True if the micro:bit is adjusting the compass for accuracy.\n"},
+    {&microbit_compass_clear_calibration_obj, "clear_calibration() - Reset the micro:bit's compass. Run calibrate() to improve accuracy.\n"},
+    {&microbit_compass_get_x_obj, "Return the compass' X axis direction that the micro:bit is facing.\n"},
+    {&microbit_compass_get_y_obj, "Return the compass' Y axis direction that the micro:bit is facing.\n"},
+    {&microbit_compass_get_z_obj, "Return the compass' Z axis direction that the micro:bit is facing.\n"},
```
ntoll: why not just "Return the compass' Z axis direction."

source/microbit/help.c
sparkslabs: Stuff is great IMO.

That said, I also use doohickey, whatsamahij, thingymabob, and several
similar do-da-mewhatsit, and similar, so pinch of salt.


Michael.

source/microbit/help.c
willingc: Why not, indeed! Much better :)

source/microbit/help.c
willingc: Will continue to think on this one to make it more precise. Will use your suggestion if needed. Was trying to keep the message short.

ntoll:
OK... so this is wonderful. I think we can collaboratively nail this so it's consistent and friendly. I'd really love people to say, "just use help(foo) it's really great". This needs to be teacher's secret weapon of choice??? As in, "Miss, why doesn't foo work?", "Why not ask help(foo)?" etc... so important to encourage autonomous exploration. This stuff ^^^ is key.

Right... off to lunch. :-)

source/microbit/help.c
ntoll: That's great! :-)

source/microbit/help.c
ntoll: I want a "hair on fire" animation we could play. ;-)

Totally get that we need to make the distinction between computer/programmer obvious!

source/microbit/help.c
ntoll: @willingc despite bowler hatted stereotypes - we're also the land that gave rise to Monty Python's Flying Circus. Be as informal as you like. Kids will love it. ;-)

Seriously, I'd love it if we can encourage a playful, exploratory, fun and creative "outlook" associated with Python.

dpgeorge:
It's nice.  The function declaration is now inline with the text, which in some cases might be a bit awkward to grok (even for a seasoned Pythonista) but I understand the reason for doing it and it is very refreshing to read this way.

willingc:
@ntoll and @dpgeorge, Here's another iteration of the help messages.

Thoughts on how long each message should be before a ``\n``? Especially those that have multiple lines?

Tried to incorporate an example for some of the commands like sleep(time).

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_accelerometer_get_y_obj, "Return micro:bit's Y axis position.\n"},
+    {&microbit_accelerometer_get_z_obj, "Return micro:bit's Z axis position.\n"},
+    // Pushbutton
+    {&microbit_button_a_obj, "micro:bit's 'A' button. When button is pressed down, is_pressed() is True.\n"},
+    {&microbit_button_b_obj, "micro:bit's 'B' button. When button is pressed down, is_pressed() is True.\n"},
+    {&microbit_button_is_pressed_obj, "If the button is pressed down, is_pressed() is True, else False.\n"},
+    // Compass 3D direction heading
+    {&microbit_compass_obj, "Use the compass to see what direction micro:bit is facing.\n"},
+    {&microbit_compass_is_calibrated_obj, "If micro:bit's compass is_calibrated() and adjusted for accuracy, return True.\nIf the compass hasn't been adusted for accuracy, return False.\n"},
+    {&microbit_compass_calibrate_obj, "Micro:bit, I command you to calibrate() and adjust your accuracy.\nWhere are my manners, please calibrate()"},
```
willingc: Looking at this line now. "Where are my manners? Please calibrate() now."

I'm trying to avoid punctuation directly after a closing paren since I think that can be confusing for learners.

ntoll:
WRT line length... <=79 + use your judgement for aesthetics..?

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
-    {&microbit_p19_obj, "Represents pin 19 on the gold edge connector.\n"},
-    {&microbit_p20_obj, "Represents pin 20 on the gold edge connector.\n"},
-    {&microbit_pin_write_digital_obj, "write_digital(v) -- write a digital value v to the pin. Use 0 (lo) or 1 (hi).\n"},
-    {&microbit_pin_read_digital_obj, "read_digital() -- read the current digital value from the pin: 0 (lo) or 1 (hi).\n"},
-    {&microbit_pin_write_analog_obj, "write_analog(v) -- write an analog value v between 0..255 to the pin.\n"},
-    {&microbit_pin_read_analog_obj, "read_analog() -- read the current analog value from the pin: 0 - 65535.\n"},
-    {&microbit_pin_is_touched_obj, "is_touched() -- True if the pin is touched, otherwise False.\n"},
+    {&microbit_module, "Useful stuff to control the micro:bit hardware.\n"},
+    // System state objects
+    {&microbit_panic_obj, "Begin panic() mode (requires a reset).\n"},
```
ntoll: So a common question I get in workshops is "What's panic for?". We should address that. How about...

"Use panic() to show something broke (requires a reset)."

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_random_obj, "Return a random(number) between 0 and 'number - 1'.\nrandom(10) returns a number between 0 and 9.\n"},
+    {&microbit_sleep_obj, "Put micro:bit to sleep(time) for some milliseconds (1 second = 1000 ms) of time.\nsleep(2000) gives micro:bit a 2 second nap.\n"},
+    {&microbit_running_time_obj, "Return running_time() in milliseconds since micro:bit's last reset.\n"},
+    // Accelerometer 3D orientation
+    {&microbit_accelerometer_obj, "Detect micro:bit's position (3D orientation).\n"},
+    {&microbit_accelerometer_get_x_obj, "Return micro:bit's X axis position.\n"},
```
ntoll: Can we address _how_ the X position is represented so kids grok what it means?

How about...

"Return a number showing micro:bit's X axis position." 

(Same of Y and Z)

TBH, not sure about this... comments..?

source/microbit/help.c
ntoll: Hmm... we need to mention *what* we're calibrating. ;-)

Also, this is the only time we're addressing the micro:bit itself. I like the humour but perhaps we shouldn't address the micro:bit..?

"Oi micro:bit, calibrate() yourself to improve the accuracy of your compass!"

Would be more colloquial British.... ;-)

How about...

"If micro:bit is lost, calibrate() it to improve the accuracy of its compass!"

Perhaps "lost" could be replaced by "confused", "fuzzy" or "dazed".

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_compass_get_x_obj, "Return the compass' X axis direction.\n"},
+    {&microbit_compass_get_y_obj, "Return the compass' Y axis direction.\n"},
+    {&microbit_compass_get_z_obj, "Return the compass' Z axis direction.\n"},
+    // Display 5x5 LED grid
+    {&microbit_display_obj, "micro:bit's 5x5 LED display.\n"},
+    {&microbit_display_print_obj, "Use print(s) to print the string 's' to the display. Try print('Hello!').\nUse print(s, i) to print string 's', one character at a time with a delay of 'i' milliseconds.\n"},
+    {&microbit_display_scroll_obj, "Use scroll(s) to scroll the string 's' from left to right on the display.\nUse scroll(s, i) to scroll string 's' with a delay of 'i' milliseconds after each character.\n"},
+    {&microbit_display_clear_obj, "Use clear() to clear micro:bit's display.\n"},
+    {&microbit_display_animate_obj, "Use animate(img, delay, stride, start=0, async=False, repeat=False) to animate\n    image 'img' with 'delay' milliseconds and 'stride' pixels offset between\n    frames. Optional: 'start' offset from left hand side, 'async' to run in the\n    background, 'repeat' to loop the animation.\n"},
```
ntoll: This has changed.

We'll need to go through the API to ensure it's both complete and accurate. ;-)

ntoll:
@willingc thanks for your efforts with this! I'm having lots of fun reviewing the changes. :-)

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_button_b_obj, "micro:bit's 'B' button. When button is pressed down, is_pressed() is True.\n"},
+    {&microbit_button_is_pressed_obj, "If the button is pressed down, is_pressed() is True, else False.\n"},
+    // Compass 3D direction heading
+    {&microbit_compass_obj, "Use the compass to see what direction micro:bit is facing.\n"},
+    {&microbit_compass_is_calibrated_obj, "If micro:bit's compass is_calibrated() and adjusted for accuracy, return True.\nIf the compass hasn't been adusted for accuracy, return False.\n"},
+    {&microbit_compass_calibrate_obj, "Micro:bit, I command you to calibrate() and adjust your accuracy.\nWhere are my manners, please calibrate()"},
+    {&microbit_compass_is_calibrating_obj, "If micro:bit's compass is_calibrating() its accuracy, it sends True.\nIf it's not busy calibrating things, it sends False.\n"},
+    {&microbit_compass_clear_calibration_obj, "Reset micro:bit's compass using clear_calibration() command. Run calibrate() to improve accuracy.\n"},
+    {&microbit_compass_get_x_obj, "Return the compass' X axis direction.\n"},
```
carlosperate: is `direction` the right word here? What would the returned direction be? (same of y and z)

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_compass_get_y_obj, "Return the compass' Y axis direction.\n"},
+    {&microbit_compass_get_z_obj, "Return the compass' Z axis direction.\n"},
+    // Display 5x5 LED grid
+    {&microbit_display_obj, "micro:bit's 5x5 LED display.\n"},
+    {&microbit_display_print_obj, "Use print(s) to print the string 's' to the display. Try print('Hello!').\nUse print(s, i) to print string 's', one character at a time with a delay of 'i' milliseconds.\n"},
+    {&microbit_display_scroll_obj, "Use scroll(s) to scroll the string 's' from left to right on the display.\nUse scroll(s, i) to scroll string 's' with a delay of 'i' milliseconds after each character.\n"},
```
carlosperate: wouldn't the direction be from right to left?

source/microbit/help.c
```diff
@@ -81,58 +81,65 @@ STATIC const mp_doc_t help_table_types[] = {
...
+    {&microbit_display_clear_obj, "Use clear() to clear micro:bit's display.\n"},
+    {&microbit_display_animate_obj, "Use animate(img, delay, stride, start=0, async=False, repeat=False) to animate\n    image 'img' with 'delay' milliseconds and 'stride' pixels offset between\n    frames. Optional: 'start' offset from left hand side, 'async' to run in the\n    background, 'repeat' to loop the animation.\n"},
+    {&microbit_display_get_pixel_obj, "Use get_brightness(x, y) to return the display's brightness at LED pixel (x,y).\n"},
+    {&microbit_display_set_pixel_obj, "Use set_brightness(x, y, b) to set the display at LED pixel (x,y) to brightness 'b'.\n"},
```
carlosperate: Is it worth point out here the min and max values required for the brightness?

deshipu:
Sorry for being late, but I only just noticed this, and I'm wondering if we shouldn't generate this and the docs from the same source? If it was python code, we would simply use docstrings for both, I wonder what we could use here?

ntoll:
@deshipu the intention wrt the help strings is very different to the dev docs. This work is written for a reading age of 11 and aimed at kids. The dev docs are for, er, tech savvy devs. ;-)

deshipu:
I see. I just see a lot of overlap with what I wrote in the docs, and I'm afraid that keeping it all in sync may be difficult.

source/microbit/help.c
willingc: @carlosperate Great suggestion.

willingc:
@deshipu @ntoll The key with the help messages are to be technically helpful to 11 year old children as well as teachers with varying degrees of technical experience. For the help messages, we're balancing tone, message length, and technical accuracy. Once the API is finalized, we should revisit the technical accuracy of the help messages.

@deshipu You are doing a nice job on the developer documentation :D

source/microbit/help.c
willingc: @carlosperate Probably would be best and more child friendly to remove the direction and mention ``character by character across the display``

source/microbit/help.c
willingc: Perhaps ``value`` or ``reading`` work better here.

source/microbit/help.c
carlosperate: I agree that removing direction might be a good edit, although `character by character` might be a bit redundant.

source/microbit/help.c
carlosperate: Whichever word is used, it'd be useful to have short description of what it represents, as most people would expect a heading from a compass rather than an x, y, z value.

source/microbit/help.c
whaleygeek: Strictly speaking, the value returned from the accelerometer is not position. It's the force in milli-g's in that plane, but that is a bit of a mouthfull too! If you tilt it in the X axis, this affects a different force on the X reading. If you shake it for example, you can get up to a +/-2g displacement in any axis.

Z 'feels' very different to X and Y in use. Holding it level on the table, you get a -1g force due to gravity. If you turn the bit upside down, you get a +1g force due to gravity. If you shake it up and down, you'll get up to +/- 2g force due to the movement. 

The best I have found to date is to describe the X/Y reading as 'tilt', as that is what it will ultimately be used for, although still not strictly correct, it might be more accurate and more obvious to an 11 year old reader. z is not tilt in any way at all, it generally is usable to work out if the micro:bit is face up or face down. The +/-2g can be used to indicate it is moving (rapidly) in the up and down plane.

source/microbit/help.c
willingc: Brightness, b, can take values 1 (min) to 9 (max). When b = 0, the display pixel is turned off.

carlosperate: great, so how does `Use set_brightness(x, y, b) to set the display at LED pixel (x,y) to brightness 'b', which can be set between 0 (off) to 9 (full brightness).\n` sound?

willingc:
As preparation for release 0.9, I've updated the help text based on the suggestions made by others, as well as information available on the BBC:microbit website. The most significant changes are in the help sections for compass and accelerometer.

willingc:
Rebased to master. Please review.

Happy to squash for merge.

ntoll:
Covered by #116.

PR #78, opened 2015-11-07
Add initial documentation for the compass

deshipu:


deshipu:
This is an initial version, to get started.

It misses the descriptions of the format in which get_?() and heading() functions return what they return.

docs/compass.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+.. py:function:: clear_calibrarion()
+
+    Undoes the calibration, making the compass uncalibrated again.
+
+
+.. py:function:: get_x()
+
+    Gives the reading on the ``x`` axis of the compass.
```
ntoll: ... as an integer.

docs/compass.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+.. py:function:: get_y()
+
+    Gives the reading on the ``y`` axis of the compass.
```
ntoll: ... as an integer.

docs/compass.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+.. py:function:: get_z()
+
+    Gives the reading on the ``z`` axis of the compass (vertical).
```
ntoll: ... as an integer.

docs/compass.rst
```diff
@@ -0,0 +1,60 @@
...
+
+
+.. py:function:: heading()
+
+    Gives the compass heading, calculated from the above readings.
```
ntoll: Gives the compass heading as an integer representation of degrees (0-359) with north as zero. Calculated from the x, y and z readings.

ntoll:
Hope my comments add a bit more context for revisions.

docs/compass.rst
deshipu: What is the range? What does it represent?

docs/compass.rst
deshipu: Great, I will add that.

deshipu: Hmm, clockwise or counter-clockwise?

docs/compass.rst
ntoll: I have no idea of the range but it returns +/- values and represents the magnetic force measured along the referenced axis.

docs/compass.rst
ntoll: Clockwise (I just checked). 

Strangely enough, the range is 0-360. :-/ This is probably a bug in the upstream device abstraction layer.

PR #79, opened 2015-11-07
Add initial documentation for the buttons

deshipu:


ntoll:
This is a great start.

However, the ``Button`` class is never available to the end user. Rather two instances representing buttons A and B are available in the ``microbit`` namespace. I wonder if this documentation is a little misleading as a result.

Thoughts..?

deshipu:
Well, it does seem like a weird API, and Sphinx obviously wasn't designed for such a case. It's a little bit similar to how True and False objects work in Python, I suppose? They are singletons, and you never get to create any more of them.
To be honest, I have no idea how to solve it.

docs/button.rst
```diff
@@ -0,0 +1,36 @@
+Buttons
+*******
+
+.. py::module:: microbit
+
+There are two buttons on the board, called ``button_a`` and ``button_b``.
+
+
+Classes
```
ntoll: Not sure if this is the correct thing to do, but suggesting it all the same for the sake of something to discuss. We should remove the Class section and replace it with something along the lines of:

```
.. py:attribute:: button_a

    Represents button A.

    .. py:method:: is_pressed()

        Returns ``True`` if the button is pressed, and ``False`` otherwise.
```

...etc...

deshipu: But then we would have some repetition with the same method described on both instances of the Button class.
Perhaps we could just mention in the class description, that it's not actually available?

ntoll: +1 That's certainly the path of least resistance. ;-)

dpgeorge:
The button instances (and all other object instances) are pre-created so that it's easy for the kids to use.

You can use `type(button_a)` to get the type, but it can't construct new instances.

These docs are for the more tech savy reader, so I think it's ok to just note that the button instances are sigletons.

deshipu:
Do we need something more here, or can it be merged?

dpgeorge:
It looks ok, and should go in before #101.  Can you please rebase to latest master?

PR #85, opened 2015-11-08
Pin docs

deshipu:


docs/pin.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    .. py:method:: write_analog(value)
+
+        Output a PWM signal on the pin, with the duty cycle proportional to
+        the provided ``value``. The ``value`` may be either an integer or a
+        floating point number between 0 (0% duty cycle) and 1023 (100% duty).
+
+    .. py:method:: set_analog_period(period)
+
+        Set the period of the PWM signal being output to ``period``.
```
deshipu: Not sure what the unit and allowed range is here.

docs/pin.rst
```diff
@@ -0,0 +1,112 @@
...
+
+    .. py:method:: set_analog_period_microseconds(period)
+
+        Set the period of the PWM signal being output to ``period`` in
+        microseconds.
```
deshipu: Not sure what is the allowed range.

docs/pin.rst
```diff
@@ -0,0 +1,112 @@
...
+    .. py:method:: read_analog()
+
+        Read the voltage applied to the pin, and return it as an integer
+        between 0 (meaning 0V) and 1023 (meaning 3.3V).
+
+    .. py:method:: write_analog(value)
+
+        Output a PWM signal on the pin, with the duty cycle proportional to
+        the provided ``value``. The ``value`` may be either an integer or a
+        floating point number between 0 (0% duty cycle) and 1023 (100% duty).
```
deshipu: We probably need a paragraph explaining what PWM is, with pictures.

docs/pin.rst
```diff
@@ -0,0 +1,112 @@
...
+|  16 | Digital |          |
++-----+---------+----------+
++-----+---------+----------+
+|  19 | Digital | I2C SCL  |
++-----+---------+----------+
+|  20 | Digital | I2C SDA  |
++-----+---------+----------+
+
+The above table summarizes the pins available, their types (see below) and what
+they are internally connected to.
```
deshipu: We will probably need some explanations about the functions of those pins, with examples etc.

docs/pin.rst
dpgeorge: Unit is milliseconds.  Allowed range is hard to figure out.  It will need testing on actual hardware.

docs/pin.rst
dpgeorge: I've found that the minimum is about 35us, but don't know what the maximum is.

dpgeorge:
Nice work adding the pinout graphic!

ntoll:
@deshipu where did you get the pinout graphic? I seem to remember that I've seen two versions and want to make sure we have the right one..! 

The people to ask would be @finneyj and/or @jaustin. Or perhaps I'm mis-remembering.

Guys..?

remay:
I believe that this is correct: https://developer.mbed.org/platforms/Microbit/#pinout

deshipu:
The graphics is from https://developer.mbed.org/platforms/Microbit/#pinout as @remay noticed. I mentioned it in https://github.com/deshipu/micropython-1/commit/567a268ce08c21db427b9da8a76ff8a749a6edc3 but I guess commit messages are not the most visible place to put such information.

I added the units and minima (I assumed that since 1ms is larger than 35µs, ti will work).

I started to wonder if only the analog pins are PWM-capable? This seems like an arbitrary restriction? Maybe ``analog_write`` should be available on the digital pins too?

docs/pin.rst
```diff
@@ -0,0 +1,172 @@
...
+Note that this works well with devices such as motors, which have huge inertia
+by themselves, or LEDs, which blink too fast for the human eye to see the
+difference, but will not work so good with generating sound waves. This board
+can only generate square wave sounds on itself, which sound pretty much like
+the very old computer games -- mostly because those games also only could do
+that.
+
+
+Hobby Servomechanisms
+---------------------
```
dpgeorge: This subsection is probably better suited to go in a pin examples section.

carlosperate: I was just about to suggest the same.

Is there going to be example documentation stored in this repository as well then? It might be hard to define a scope of how many examples might be needed or what would be outside of the scope of the documentation.

dpgeorge: There is an examples/ directory which should have a large number of example scripts (as long as they each provide something new or different).

I think for each class there should be a few examples, possibly just linking to a script in examples/ dir.

carlosperate: I assumed from your original comment that you meant to have examples as documentation, but yes having practical scripts with general examples and their explanations as part of the comments seems to be the best place for this kind of information.

deshipu:
Do we need to change something here, or can we merge it?

dpgeorge:
@deshipu it looks good to me.  Can you please rebase to latest master, and if possible squash "Remove the servo example from pin docs" into "Add information about PWM and servos"?

deshipu:
Squashed.

dpgeorge:
Please rebase against latest master and I'll merge this next (or I could just do the rebase myself...).

deshipu:
rebased

dpgeorge:
Thanks!

PR #86, opened 2015-11-08
Add initial I²C documentation

deshipu:


docs/i2c.rst
```diff
@@ -0,0 +1,29 @@
...
+.. py:module:: microbit.i2c
+
+The ``i2c`` module lets you communicate with devices connected to your board
+using the I²C bus protocol.There can be multiple devices connected at the same
+time, and each one has its own unique address, that is either fixed for the
+device or configured on it. How exactly you should communicate with those
+devices, that is, what bytes to send and how to interpret the responses,
+depends on the device in question and should be described separately in that
+device's documentation.
+
```
carlosperate: Are there internal pull ups in the I2C lines? knowing if additional components are needed to connect to the I2C bus should be specified somewhere.

carlosperate:
Are there internal pull ups in the I2C lines? knowing if additional components are needed to connect to the I2C bus should be specified somewhere. Also it would be good to know if the address size can be selected and how.

dpgeorge:
> Are there internal pull ups in the I2C lines?

I don't think the answer to this is yet public knowledge.

deshipu:
The pin api doesn't have anything for controlling pullups, so I would assume no.

dpgeorge:
> The pin api doesn't have anything for controlling pullups, so I would assume no.

But there could be actual resistors on the board doing the pulling.

deshipu:
>> The pin api doesn't have anything for controlling pullups, so I would assume no.
> But there could be actual resistors on the board doing the pulling.
We can theoretize here, but wouldn't that make the pins unsuitable for pretty much anything else?

dpgeorge:
> but wouldn't that make the pins unsuitable for pretty much anything else?

I guess.  But that doesn't mean they can't have pullups.

At this point we can't have anymore information, so will need to mark this as TODO.

remay:
The nRF51 reference manual is a good place to start for what the hardware
should be capable of:
https://www.nordicsemi.com/eng/nordic/download_resource/20337/12/59738211

Working out what the various nordic/mbed and DAL defaults are is hard work
though.

All nRF51 GPIO Pin are capable of being configured pullup/pulldown/none by
software.   The I2C implementation can be software configured to use any 2
GPIO Pins. (From what I can see).

deshipu:
> At this point we can't have anymore information, so will need to mark this as TODO.
I created Issue #88 for this.

carlosperate:
In case anybody was wondering as I was, after looking at the datasheet @remay posted, it looks like the nRF51 can only act as a master, and [this](https://devzone.nordicsemi.com/question/5134/i2ctwi-slave-with-nrf51822/) discussion at the nordic forum also seems to confirms it.

docs/i2c.rst
```diff
@@ -0,0 +1,43 @@
+I²C
+***
+
+.. py:module:: microbit.i2c
+
+The ``i2c`` module lets you communicate with devices connected to your board
+using the I²C bus protocol. There can be multiple slave devices connected at
+the same time, and each one has its own unique address, that is either fixed
+for the device or configured on it. Your board acts as the I²C master.
```
carlosperate: This paragraph assumes the reader is familiar with the definition of what an I2C master and slave are. Perhaps it could be rephrase to initialy say you can connect devices to the microbit, and then define that the master would be the "controller" and the slaves are the the "controlled" devices.

deshipu: I think that it is safe to assume that a user who attempts to use an I²C device already has some familiarity with the protocol, and if not, should familiarize with it before actually connecting anything. Renaming it to "controller" and "controlled" will only add confusion, because the official documentation for the device is going to use the established and standard terms "master" and "slave", and the user will have to guess that this is what we mean.

docs/i2c.rst
```diff
@@ -0,0 +1,43 @@
...
+.. py:function:: write(addr, buf, repeat=False)
+
+    Write bytes from ``buf`` to the device with address ``addr``. If ``repeat``
+    is ``True``, no stop bit will be sent.
+
+
+Connecting
+----------
+
+Connecting an I²C device to the board is simple -- you have to connect the pin
```
willingc: @deshipu For user friendliness replace "board" with "micro:bit". I would also highlight ``SCL`` and ``SDA``. 

*Suggestion*: 
You should connect the device's ``SCL`` pin to micro:bit pin 19, and the device's ``SDA`` pin to micro:bit pin 20. You may need to connect the device to the micro:bit ground (pin ``GND``). You may need to power the device using an external power supply or the micro:bit.

deshipu:
I updated the docs with the suggestions, can we merge this?

dpgeorge:
It's good.  Please rebase against master.

I think eventually we should add i2c.init() so that the pins can be reassigned, but that's a task for another day.

deshipu:
rebased

dpgeorge:
Well, now that other commits have gone in, this needs rebasing again... please?

@markshannon take note.

PR #87, opened 2015-11-08
Initial documentation for uart

deshipu:


deshipu:
I need some help here, as I can't really decode the parameters to those functions, and also I'm not entirely sure how exactly they behave.

Does initializing the UART disable the REPL on the USB? What happens when you use "print" afterwards? Are we alone in the universe?

dpgeorge:
UART is a preliminary feature.

There is a single UART peripheral and it is configured on start up to be connected to the USB serial convertor.

If you just start using the `uart` object then it'll read/write data from/to the USB serial device.

If you init the `uart` object then it'll be connected to the pins that you supply in the init method.  They default to P0/P1 for tx/rx.  So if you do `uart.init()` then you lose the REPL.  This needs to be well documented!

deshipu:
Excellent. Is there a way to connect it back to USB?

dpgeorge:
> Is there a way to connect it back to USB?

Not really.  Not unless the script itself has code to automatically reconnects (eg when you press both buttons at the same time).  But actually I don't think that can even be done by the script because the pins that go to the USB serial device are not exposed on the Python side.

docs/uart.rst
```diff
@@ -0,0 +1,65 @@
...
+        * 19200
+        * 28800
+        * 38400
+        * 57600
+        * 115200
+
+    The ``bits`` defines the size of bytes being transmitted, it can be either
+    7 or 8. The ``parity`` parameter defines how partity is checked, and it can
+    be ``None``, ``microbit.uart.ODD`` or ``microbit.uart.EVEN``. The ``stop``
+    parameters tells the number of stop bits.
```
carlosperate: I assume the board can probably support 1, 2,and possibly 1.5 stop bits, whatever the case, this should probably be specified here.

deshipu:
> But actually I don't think that can even be done by the script because the pins that go to the USB serial device are not exposed on the Python side.

Maybe it would make sense to add such a function to the library.

docs/uart.rst
dpgeorge: As far as I know, it only supports 1 stop bit and 8 data bits.

dpgeorge:
> Maybe it would make sense to add such a function to the library.

We would just need to add the pins, pin_usb_tx and pin_usb_rx.

docs/uart.rst
deshipu: Then why specify it?

dpgeorge: Code was copied from normal MicroPython, which has this capability.  So, would need to check properly if it is supported or not by the hardware, and if it's not then probably best to remove these arguments (bits and stop).

carlosperate: It's probably not a bad idea to maintain it as an optional parameter, useful for self-explanatory code, and if future versions of the board support other values there is no need to change the interface.

dpgeorge: Yes, either way.  If these args are not in this version then it still provides a path for upgrading to include new keyword args which is backwards compatible.  If the args are in this version then they'd need to be checked to be valid values (ie bits=8, stop=1).

carlosperate:
From what I see, there seems to be a willingness to differ a little from upstream micropython in order to simplify the interface (not quite sure how or if the bbcmicro micropython and micropython proper are synchronised). If that is the case, might I suggest to change the pin argument interface? passing a tuple forces the user to remember the TX and RX pin order, so having something along the lines of:

```python
init(baudrate=9600, bits=8, parity=None, stop=1, tx_pin=microbit.pin0, rx_pin=microbit.pin1)
```

is more intuitive than:

```python
init(baudrate=9600, bits=8, parity=None, stop=1, pins=(microbit.pin0, microbit.pin1))
``` 

Perhaps this is not the best place to discuss this, as this is a documentation pull request, we can always open a different issue if you find it worth to consider.

dpgeorge:
> From what I see, there seems to be a willingness to differ a little from upstream micropython in order to simplify the interface

The aim with the microbit API is to be really simple to do basic stuff, but still have a gradual path for more advanced features.  Keeping in sync with upstream MicroPython is nice, but not a top priority.

> not quite sure how or if the bbcmicro micropython and micropython proper are synchronised

The core of MicroPython is occasionally copied across from upstream to this repo see eg 24e16bf3b78e014bedb4626f5767a05b3ae441f3

But that's just the core.  The microbit-specific modules and the API are independently developed here.

> might I suggest to change the pin argument interface? ...  Perhaps this is not the best place to discuss this

The uart was added in a hurry just to land features and no real discussion was had on the API, so feel free to open a proper discussion!  You can use a github issue, or maybe better to post to microbit@python.org since there are a lot of people there with passionate feelings about how APIs should be :)

carlosperate:
Thanks @dpgeorge, I've subscribed to the mailing list and started reading it, there seems to be an ongoing discussion about several aspects of the API already, it'll take me a little while to catch up with the conversation.

sparkslabs:
-1 on the use of a tuple.
+1 on named parameters

Having named values is IMO clearer - especially considering that the values
you're most likely to want to change are these.

Michael

docs/uart.rst
```diff
@@ -0,0 +1,65 @@
...
+        pin on your board needs to be connected with the RX pin on the device,
+        and the RX pin -- with the TX pin on the device. Also make sure the
+        ground pins of both devices are connected.
+
+.. py:function:: any()
+
+    Return ``True`` if there are any bytes to be read in the input buffer.
+    Return ``False`` when the buffer is empty.
+
+.. py:function:: read(buf_in, size)
```
dpgeorge: It should be `read(size=-1)`, where -1 means read all.  It returns the bytes object containing the read data.

carlosperate: Maybe this is another portion of the API that might need further discussion? `-1`, even if it is just used as a default value, might not be all that obvious of what it represents.

dpgeorge: It's standard Python: https://docs.python.org/3.4/tutorial/inputoutput.html#methods-of-file-objects

carlosperate: Indeed it is, but it doesn't make it any more intuitive nor easy to explain with something else than "that's the way it's always been". However, as a optional parameter that can be left out to retrieve all data, I don't see any real need to change it, and there is also a clear benefit to use standard conventions.

deshipu: All of those functions need proper descriptions and proper parameters. Unfortunately I don't know the Micropython C API well enough to guess them from the source code, and there is no other source of information.

deshipu:
I need help from a knowledgable person to fill in the details for the ``readall``, ``readinto`` and ``readline``.
Also, do those return bytes or strings?

dpgeorge:
This might help with documentation: http://docs.micropython.org/en/latest/library/pyb.UART.html (you could even copy the sphinx code directly).

The uart is in binary mode and returns bytes objects.

deshipu:
I assume, "newline character" there is "\n", not "\r" or "\r\n" or anything like that? Is it also included as part of the bytes that readline returns?

dpgeorge:
It splits on '\n' only.  '\n' is included as part of the line, as per CPython: https://docs.python.org/3.5/library/io.html#io.IOBase.readline

deshipu:
Thanks! I will include that.

deshipu:
I included the docs from micropython's``pyb.UART``, added a note about the newline character being included in the returned bytes, and added a warning that initializing UART will disable REPL. Also squashed and rebased.

dpgeorge:
Thanks!

PR #89, opened 2015-11-09
Add initial documentation for display

deshipu:


docs/display.rst
```diff
@@ -0,0 +1,54 @@
...
+
+    Set the brightness of all LEDs to 0.
+
+
+.. py:function:: print(string_or_image), delay=400)
+
+    If ``string_or_image`` is a string, it will be displayed one letter at a
+    time, with ``delay`` milliseconds delay between each pair of letters.  If
+    it is an ``Image``, it will simply show that image.  This function blocks
+    until it is finished.
```
deshipu: I changed the name of the parameter here, in an attempt to make it a little clearer, but this is a very unpythonic interface. Perhaps it would make sense to have a separate ``blit(image)`` (or ``show``) function.

deshipu: I don't know if this function actually blocks. I assumed it does here, to get the attention of whoever can test it.

docs/display.rst
```diff
@@ -0,0 +1,54 @@
...
+
+
+.. py:function:: scroll(string, delay=400)
+
+    Similar to ``print``, but scrolls the ``string`` horizontally instead. The
+    ``delay`` parameter controls how fast the text is scrolling. This function
+    blocks until it is finished.
```
deshipu: I don't know if this function actually blocks. I assumed it does here, to get the attention of whoever can test it.

docs/display.rst
```diff
@@ -0,0 +1,54 @@
...
+
+
+.. py:function:: animate(image_or_iterable, delay, stride, start, wait=True, loop=False)
+
+    If ``image_or_iterable`` is a single ``Image``, you also have to specify
+    ``delay``, ``stride`` and ``start``, and it will animate that image
+    somehow.
```
deshipu: I have no idea what this actually does with the image, and no way to test it.

deshipu: This is again an unpythonic interface, and could perhaps be split into two separate functions.

deshipu:
I have much more trouble documenting this module, but I'm sharing what I have anyways, so that we can improve it.

docs/display.rst
```diff
@@ -0,0 +1,54 @@
...
+
+This module controls the 5×5 LED display on the front of your board. It can
+be used to display images, animations and even text.
+
+Functions
+=========
+
+.. py:function:: get_pixel(x, y)
+
+    Return the brightness of the LED at column ``x`` and row ``y``.
```
deshipu: Unlcear what is the range here.

docs/display.rst
```diff
@@ -0,0 +1,54 @@
...
+
+
+.. py:function:: set_pixel(x, y)
+
+    Set the brightness of the LED at column ``x`` and row ``y``.
```
deshipu: Argh, forgot ``value``. Will fix that when filling the rest of missing information.

docs/display.rst
deshipu: A typo here with the extra ")".

dpgeorge:
It's good to see documentation being written witouth a device.  It really shows how clear (or not) the API is.  Great work!

docs/display.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. py:function:: clear()
+
+    Set the brightness of all LEDs to 0.
+
+
+.. py:function:: print(string, delay=400)
+
+    Display the ``string`` one letter at a time, with ``delay`` milliseconds
+    between each pair of letters.
```
deshipu: Not related to the documentation itself, just a comment on the choice of name. I think this is a horrible idea. You are basically setting yourself up for people confusing this with the build-in print. Not to mention that "printing" an image doesn't really sound like a good idea.

Possible other names: ``flash``, ``show``, ``write``.

By the way, if I was designing a general API for this, I would make this function check for a magical attribute, similar to ``__repr__``, that would return an iterable of images. Then you could have user objects that can also be displayed, and instead of "animated" function below, you would just have an "Animation" iterable.

deshipu: On a second thought, it could just work with any iterable of images or characters.

docs/display.rst
```diff
@@ -0,0 +1,67 @@
...
+    Similar to ``print``, but scrolls the ``string`` horizontally instead. The
+    ``delay`` parameter controls how fast the text is scrolling. This function
+    blocks until it is finished.
+
+
+.. py:function:: animate(image, delay, stride, start, wait=True, loop=False)
+
+    Display the ``image`` starting with its ``start`` column, and scroll
+    it every ``delay`` milliseconds to the right by ``stride`` columns, until
+    the last column of the image is displayed.
```
deshipu: After talking with @ntoll I clarified what it does exactly, although I'm still not sure this is accurate.

deshipu:
Do we need something more here, or can we merge this?

docs/display.rst
```diff
@@ -0,0 +1,67 @@
...
+
+
+.. py:function:: show(image)
+
+    Display the ``image``.
+
+
+.. py:function:: scroll(string, delay=400)
+
+    Similar to ``print``, but scrolls the ``string`` horizontally instead. The
```
dpgeorge: Print will need to become show.

deshipu:
Fixed and squashed.

PR #90, opened 2015-11-09
Initial documentation for the Image class

deshipu:


docs/image.rst
```diff
@@ -0,0 +1,108 @@
...
+=======
+
+.. py:class:: Image(width_or_string=None, height=None, buffer=None)
+
+    If ``width_or_string`` is a string, creates an image from that string.
+
+    If it's an integer, a height has to also be specified, and it creates
+    an image with ``width_or_string`` columns and ``height`` rows. Optionally
+    ``buffer`` can be an array of ``width_or_string``×``height`` integers to
+    initialize the image.
```
deshipu: Again, this is awkward, it would be much easier to explain if there was a ``from_string`` classmethod.

docs/image.rst
```diff
@@ -0,0 +1,108 @@
...
+    .. py:method:: shift_down(n)
+
+        Same as ``image.shift_up(-n)``.
+
+
+Attributes
+==========
+
+The ``Image`` class also has the following instances of itself included as its
+attributes:
```
deshipu: Not sure a class including instances of itself is sane...

docs/image.rst
dpgeorge: You can write it as 2 separate constructors:
```
Image(str)
Image(width, height, buffer=None)
```
I think that's pretty clear.

deshipu: However, I can see no way to do that with Sphinx, because the py:class is a block. Unless I describe it as two separate classes, but then it's not clear that it's the same class with the same methods (should I duplicate the method descriptions?).

Finally, and I know it's a matter of taste to some degree, this is unpythonic and can lead to surprising results when the parameter you are passing is not the type you expected it to be.

markshannon: IMO, the string constructor is the "natural" one. If we were targeting adult developers, then I would like to turn the buffer constructor into a `from_buffer` static method.

Also, given that brightnesses are in `range(10)`, the buffer constructor has a rather odd interface.
Ideally I would like `Image.bw_from_data(width, height, data)` and `Image.grey_from_data(width, height, data)`
To create a 2x2 black and white image, one would write `Image.bw_from_data(2, 2, 0b0101)`
To create a 2x2 greyscale image, one would write `Image.grey_from_data(2, 2, 0909)`

docs/image.rst
markshannon: IMO, not only sane, but reasonably sensible  for immutable singleton objects.
My concern with the ever increasing list of images is that they drown out the methods on Image when tabbing for auto-completion.
Moving them to an `images` module would address both of our concerns.

deshipu: It's not a singleton if:
a) there are multiple instances of it,
b) users can create more instances,
as it is for the Image objects. Also, are they really immutable?

docs/image.rst
deshipu: I'm not sure about from_string being the default. I think the default should be an empty image with optional dimensions (defaulting to 5×5) -- that seems like the most basic thing. Then you could have:

``Image.from_string(string)``
``Image.from_data(width, height, data, mono=False)``

and also ``image.load_data(data, mono=False)`` for changing an existing image.

dpgeorge: Having 2 constructors is fine, since they use different types for the args as well as different number of args, so it's very unlikely to make a mistake that doesn't raise an exception.

I'm happy to remove the buffer argument.  I added it at the start thinking it would be useful but I never actually used it.  Initialising with a string is the main way to create an image, but also being able to make a blank image of a given w*h is useful (and I use it).

We must remember that the target audience is children and things must stay simple.

docs/image.rst
dpgeorge: The images are constants, so having them as they are makes sense to me (it's like select.POLL_IN, select.POLL_OUT, for example).

Putting them in a different namespace would just make it harder for the kids to use.  Lowest barrier to usage is as it is now: display.print(Image.HEART).

deshipu: And it lets you do stuff like display.print(Image.HEART.HAPPY.ANGRY) ;)

dpgeorge: > And it lets you do stuff like display.print(Image.HEART.HAPPY.ANGRY) ;)

That's true.  Image is not a module, it's a class as you point out.

> Does Image.HEART.set_pixel(0, 0, 0) raise an exception?

Not sure... maybe .set_pixel is not even an attribute of a constant image.

deshipu: Hmm.. so I should probably document two classes, Image and the constant image.

markshannon: There is one `Image` class. `set_pixel` is a method on it.
`Image.HEART.set_pixel(0, 0, 0)` raises an exception.
All images of the form `Image.XXX` are immutable (and black and white)

docs/image.rst
```diff
@@ -0,0 +1,126 @@
...
+*****
+
+.. py:module:: microbit
+
+Classes
+=======
+
+.. py:class::
+    Image(string)
+    Image(width=None, height=None, buffer=None)
```
deshipu: I found a way to specify two constructors in Sphinx.

docs/image.rst
```diff
@@ -0,0 +1,126 @@
...
+
+    If ``string`` is used, it has to consist of digits 0-9 arranged into
+    lines, describing the image, for example::
+
+        image = Image("""\
+        90009
+        09090
+        00900
+        09090
+        90009""")
```
deshipu: Are the spaces ignored? If they are ignored, this could be formatted much nicer.

docs/image.rst
```diff
@@ -0,0 +1,126 @@
...
+    * ``Image.SAD``
+    * ``Image.CONFUSED``
+    * ``Image.ANGRY``
+    * ``Image.ASLEEP``
+    * ``Image.SURPRISED``
+    * ``Image.YES``
+    * ``Image.NO``
+    * ``Image.CLOCK12``, ``Image.CLOCK11``, ``Image.CLOCK10``, ``Image.CLOCK9``,
+      ``Image.CLOCK8``, ``Image.CLOCK7``, ``Image.CLOCK6``, ``Image.CLOCK5``,
+      ``Image.CLOCK4``, ``Image.CLOCK3``, ``Image.CLOCK2``, ``Image.CLOCK1``
```
deshipu: Perhaps it would make sense to put those in a tuple, so that you could select them easily with a variable, instead of making huge if-elsif-else statements?

docs/image.rst
```diff
@@ -0,0 +1,126 @@
...
+    * ``Image.ARROW_N``, ``Image.ARROW_NE``, ``Image.ARROW_E``,
+      ``Image.ARROW_SE``, ``Image.ARROW_S``, ``Image.ARROW_SW``,
+      ``Image.ARROW_W``, ``Image.ARROW_NW``
```
deshipu: Same here.

docs/image.rst
markshannon: We treat spaces as zeroes. So `"9   9\n 9 9 \n  9  \n 9 9 \n9   9\n"` is a cross.
Also, I think `\n` is not a great line separator. `"90009:09090:00900::09090:90009"` might be better.

docs/image.rst
markshannon: :+1:

ntoll: Yeah, but we should make sure ``Image.CLOCK[0]`` is the same as ``Image.CLOCK[12]`` is the same as ``Image.CLOCK12``. ;-)

docs/image.rst
ntoll: Or a dict in this case, viz:

```
>>> Image.ARROW['N']
```

Feels a lot more natural than:

```
>>> Image.ARROW[0]
```

markshannon: I think that a sequence is better:
```python
def spin():
    display.animate(ARROWS, repeat=True)
```

docs/image.rst
markshannon: Having `CLOCK12` at both ends of the sequence is a bit odd.
To me, the "natural" sequence for a clock program is a 12 element sequence with `CLOCK12` at the 0<sup>th</sup> position.

docs/image.rst
markshannon: On second thoughts, maybe the dict is better. It is easy enough to add an ancillary list:
```python
compass = [ 'N', 'NE', ...]
compass_points = [ ARROW[p] for p in compass ]
```

ntoll: I've changed my mind. :-) 

Kids can get the correct arrow directly via ``Image.ARROW_FOO`` so the benefit of the sequence works out. Good call @markshannon

docs/image.rst
ntoll: @markshannon you're right. I agree that ``Image.CLOCK[0]`` should be the same as ``Image.CLOCK12``.

Not sure what I was thinking. I should go make a cup of tea or something to wake up my brain. ;-)

docs/image.rst
deshipu: You could treat both \n and : the same, then you get a choice of both.

If we changed how whitespace is interpreted, we could have the images in multiline strings.

docs/image.rst
deshipu: I added issue #92 to discuss this.

docs/image.rst
deshipu: I added issue #93 to discuss it.

deshipu:
I guess we have to wait for #102 with this? Or should we merge it now, add docs update to #102?

markshannon:
I think we should merge it now. I'm happy to add a docs update to #102.

dpgeorge:
Let's do this one next... @deshipu if you don't mind rebasing again...

deshipu:
Certainly, although github says this branch is up-to-date and can be merged?

dpgeorge:
Thanks!

> although github says this branch is up-to-date and can be merged?

Yes, it might be mergable, but I don't think github cares that it's going to be a pure fast-forward (which is important to keep the history clean).

PR #97, opened 2015-11-18
Reduce the size of initial help text.

ntoll:
Does what it says on the tin. This could probably be compacted even more.


source/microbit/help.c
```diff
@@ -33,40 +33,27 @@
...
+"  sleep(1000)\n"
+"  button_a.is_pressed()\n"
 "What do these commands do? Can you improve them? HINT: use the up and down\n"
 "arrow keys to get your command history. Press the TAB key to auto-complete\n"
-"unfinished words (so 'mi' becomes 'microbit' after you press TAB). These\n"
+"unfinished words (so 'di' becomes 'display' after you press TAB). These\n"
 "tricks save a lot of typing and look cool!\n"
 "\n"
 "Explore:\n"
 "Type 'help(something)' to find out about it. Type 'dir(something)' to see what\n"
```
inglesp: Unless `something` is available in the global namespace, would it be sensible to give a concrete example here?  Or is this an initiative test?

deshipu: ``help(help)`` ;)

carlosperate: I'd argue that "something" is just fine, otherwise I'd be like explain what "any key" is, users might be young but perfectly capable to understand what this means.

ntoll: I'm with @carlosperate - experience tells me your average 11yo in the UK will understand the intention here.

Are we good to merge this?

willingc: @ntoll I agree with you and @carlosperate that an 11 yo will understand the intent. Python would use ``help(object)`` and ``dir(object)`` with the option to not pass in an object ``help()`` and ``dir()``.

From the message context, it seems ``help(command)`` could work. Both ``command`` and ``something`` seem more user friendly than ``object`` for an 11 yo. I would say it's good to merge as is or with a wording tweak.

ntoll:
OK... I'm going to rebase from master and merge this.

Issue #69, opened 2015-11-07
Add was_pressed(), get_presses() and reset() methods to buttons.


ntoll:
Put simply, pull request #53 needs to be fixed / merged.

willingc:
To clarify for others that are new to these issues, new methods for buttons should be added to the code base. The methods needed are:
- [ ] was_pressed()
- [ ] get_presses()
- [ ] reset()

I believe that @markshannon is working on this per message on #53.

dpgeorge:
I think it should be called "reset_presses()".

willingc:
@dpgeorge +1

markshannon:
Fixed by https://github.com/bbcmicrobit/micropython/issues/101

PR #101, opened 2015-11-22
Implement was_pressed, get_presses and reset for buttons.

markshannon:
This PR implements https://github.com/bbcmicrobit/micropython/issues/69.

As the DAL does not expose the functionality we require and I don't want any more hacks like:
https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/microbitdisplay.cpp#L209
I've reimplemented the pin de-bouncing logic from the DAL and put it in a class of its own.


dpgeorge:
If we wanted to use the DAL, couldn't we just listen on the message bus for wanted events and then set flags based on those?

Of course, what you've implemented here is much more efficient with memory (ie doesn't need to allocate on the heap whenever a button is pressed...) but I'm just thinking how to minimise divergence from the DAL.

dpgeorge:
> As the DAL does not expose the functionality we require

Can you elaborate on this?

Note that if you set up the message bus events correctly, then the DAL does not allocate any heap memory on button event.

markshannon:
I probably should have said "does not expose the functionality we require, without dragging in lots of unwanted dependencies." 
However, as it looks like we might be making some progress on https://github.com/lancaster-university/microbit-dal/issues/40, I'll revert to using events. At least, for now.

markshannon:
And rebased.

markshannon:
Rebased again...

dpgeorge:
I think it's much better using the DAL.

Issue #92, opened 2015-11-11
Make series of build-in images easier to use


deshipu:
Right now we have two series of images in ``microbit.Image``, clock faces and arrows, and they all have names like ``CLOCK11``, which makes it awkward to access them automatically based on a value of a varibale, or to iterate over them.

It would be nice to have some way of refering to those collections by the collection name and the number in sequence, or, in case of arrows, maybe even name of of the direction as a string.

Perhaps it would be also nice if we could refer to all the other build-in images through a string with their name.

ntoll:
Given discussion elsewhere, how about something like...

```
Image.CLOCKFACE = (Image.CLOCK12, Image.CLOCK1...., Image.CLOCK11)
```

and 

```
Image.ARROWS = (Image.ARROW_N, Image.ARROW_NE..., Image.ARROW_NW)
```

These would be relatively simple to implement. Ergo, I'll add to milestone 0.9. ;-)

dpgeorge:
Making CLOCKFACE/ARROWS/etc a list of images might be confusing because then you have some constants which are images and some which are lists and it's not clear which are which just by the name.  Maybe prefix the name by `ALL_`, ie `ALL_CLOCKS` and `ALL_ARROWS`?

> Perhaps it would be also nice if we could refer to all the other build-in images through a string with their name.

How about: `getattr(Images, "CLOCK1")`?

deshipu:
It's already confusing, because you also have some that are methods of the Image class and some that are python internal stuff in there, and there is no way to just list the ones that are images.

> How about: ``getattr(Image, "CLOCK1")``?
How about ``exec("image = Image.{}".format("CLOCK1"))"``? Because that is what you will see used in this situation.

markshannon:
I think we may need less confusing name than `CLOCKFACE` and `ARROWS`.
So what are those names? I don't know. Anyone have any good suggestions?

deshipu:
How about removing the images from inside the ``Image`` class and putting them in separate collections?
Then there could be ``IMAGE.HEART``, ``ARROW[2]`` or ``ARROW.NE``, and ``CLOCK[12]``.

Other ideas:

``ARROW_COLLECTION``, ``ARROW_SEQUENCE``, ``ARROW_SEQ``, ``ARROW_FRAMES``, ``ARROW_IMAGES``, ``ROTATING_ARROW`` (it can be passed to ``display.show`` as is), ``SPINNING_ARROW``

[...]

PR #102, opened 2015-11-22
Add CLOCKFACE and ARROWS tuples of images.

markshannon:
Implements https://github.com/bbcmicrobit/micropython/issues/92


dpgeorge:
I still think we should have a naming scheme for these constants that makes it clear if something is a single image or a list/tuple of images.

source/microbit/microbitconstimagetuples.c
```diff
@@ -0,0 +1,58 @@
+/*
+ * The MIT License (MIT)
```
dpgeorge: @markshannon You should put a copyright header here with your name.

markshannon: Ownership conflicts with my trotskyist beliefs :wink: 
Seriously, could we give the copyright to the PSF, and not have to worry about the legal crap?

markshannon:
W.r.t. naming, I think that a plural name is sufficient for tuples. But what is a good plural name for `CLOCKFACE`; `CLOCK_HANDS`?

source/microbit/microbitconstimagetuples.c
dpgeorge: Someone needs to "own" it so they can decide on how the code is used.  I'm ok to give it to the PSF, but someone from there would need to approve it.

Otherwise you could make this file public domain, but it might be a bit strange to have some files MIT, some PD...

dpgeorge:
> I think that a plural name is sufficient for tuples

But ARROWS could be an image with lots of little arrows in one picture.  And CLOCKFACE could be a single clock face with markings at 12,3,6 and 9.  Without an obvious way to tell if a constant is an image or a list of them you'll always need to go to the docs to find out (or just try using it, but that seems wrong).

source/microbit/microbitconstimagetuples.c
markshannon: Probably best to be consistent. I don't mind putting my name on it.

markshannon:
The problem with changing the names is that we need better ones to change them to. See https://github.com/bbcmicrobit/micropython/issues/92.

markshannon:
Updated to reflect https://github.com/bbcmicrobit/micropython/issues/92.
Rebased

ntoll:
LGTM.

These ease-of-use collections and the re-factored display API are a wonderful step forward for teaching and learning about this stuff. We have simple methods with ready made examples that are easy for kids to understand.

Great stuff!

dpgeorge:
@markshannon if you want to rebase, I can merge this next.

markshannon:
Rebased.

PR #120, opened 2015-11-29
Uncomment help code. Ensure modlove and modthis use 'C' linkage.

markshannon:
@ntoll This fixes your linker problems.


ntoll:
@markshannon many many thanks. As you probably realise, my level of C/C++ is 12yo undergraduate level C101. Would love some book suggestions for C/C++ so I can skill up. ;-)

Issue #119, opened 2015-11-29
microbit.i2c.write() does nothing


deshipu:
So, today I tried to connect the micro:bit to my quadruped robot (http://tote.rtfd.org) using I²C. I have a firmware (https://bitbucket.org/thesheep/tote/src/tip/i2cslave/) that makes the Arduino Pro Mini on that robot behave like a simple servo controller -- it basically just listens for the servo positions on the I²C bus.

I have previously tested this using Raspberry Pi and the following code:
```
import smbus
import struct
import math

bus = smbus.SMBus(1)
ADDRESS = 0x04

def servo_move(servo, radians=None, degrees=None):
    if radians is None:
        radians = math.radians(degrees)
    bus.write_block_data(ADDRESS, servo,
        [ord(b) for b in struct.pack('f', radians)])

servo_move(0, degrees=45)
```
so I know it is working properly. Now, I tried this with micro:bit:
```
import ustruct
import math
from microbit import i2c

i2c.write(0x04, ustruct('bf', 0, math.radians(45)))
```
The effect? Nothing at all. So I thought that maybe I don't understand how ``write_block_data`` works. I decided to make a simpler test. I wrote the following Arduino sketch:
```
#include <Wire.h>

void setup() {
  Serial.begin(115200);
  Wire.begin(8);
  Wire.onReceive(receiveEvent);
  Serial.println("Listening...");
}

void loop() {
  delay(100);
}

void receiveEvent(int howMany) {
  while (0 < Wire.available()) {
    char c = Wire.read();
    Serial.print(c, HEX);
  }
  Serial.println();
}
```
And I tested it with the following Python code on the Raspberry Pi:
```
import smbus

bus = smbus.SMBus(1)
bus.write_byte(8, 10)
```
It correctly printed "A". Now, trying that with micro:bit:
```
from microbit import i2c

i2c.write(8, b"X")
```
does nothing. Note, that it does nothing no matter if the Arduino is connected or not -- it doesn't even raise any exceptions if the I²C bus is not connected at all.

I suppose that concludes my work on getting this robot to work with micro:bit.

markshannon:
There are two issues here.
1. We should raise an exception if the underlying DAL method fails.
2. If the underlying write method doesn't work, then it needs to be reported as a DAL issue to 
https://github.com/lancaster-university/microbit-dal/issues

markshannon:
Could you set the pins that communicate with the robot directly?

deshipu:
[...]

PR #121, opened 2015-11-29
Raise an exception if i2c read/write fails.

markshannon:
Partial fix for https://github.com/bbcmicrobit/micropython/issues/119

### DIFF
--- a/source/microbit/microbiti2c.cpp
+++ b/source/microbit/microbiti2c.cpp
@@ -51,10 +51,10 @@ STATIC mp_obj_t microbit_i2c_read(mp_uint_t n_args, const mp_obj_t *pos_args, mp
     // do the I2C read
     vstr_t vstr;
     vstr_init_len(&vstr, args[1].u_int);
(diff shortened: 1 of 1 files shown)


carlosperate:
We are currently looking at return error values for the UART module, and throwing an exception for the pin module  (if a digital pin write value is incorrect) and now the I2C. We should probably agree on a single way to deal with these kind of issues. Coming from an embedded background I am used to deal with return states, but throwing exceptions for erros is definitely more pythonic.

dpgeorge:
> We should probably agree on a single way to deal with these kind of issues. Coming from an embedded background I am used to deal with return states, but throwing exceptions for erros is definitely more pythonic.

Raising an exception is, I think, exactly the right thing to do.  That's what Python is.

I am happy to merge this.

markshannon:
Rebased.

carlosperate:
> Raising an exception is, I think, exactly the right thing to do. That's what Python is.

I agree, we just need to remember to change the UART module as well then.

PR #123, opened 2015-11-29
Fix many misspellings in comments

kwadrat:


markshannon:
Quite a few of these are in the MicroPython code base. Perhaps they should be applied upstream.
@dpgeorge?

dpgeorge:
We cannot change anything from upstream (eg in source/py, inc/py) because then it becomes a nightmare when upgrading that part of the code.

Fixing spelling mistakes in the docs and examples is important.  In the code, less so, but I'm fine with it.

ntoll:
Most of these corrections that a "public facing" have been addressed by #190. Closing as a result. Many thanks for your contribution @kwadrat..!

kwadrat:
Thank you :)

Issue #36, opened 2015-09-28
Consider some form of event handling execution model


dpgeorge:
The excution model for uPy on the microbit is currently a single thread executing with full control.  There are idle and system "threads" behind the scenes, but these are used to update low-level things like the display and are not exposed to the Python API (except as background animations).

TD has event handlers for things like button presses, and there have been requests for such things in uPy.  Eg:

``` python
def click_handler():
    print("you clicked the button!")
button_a.onclick(click_handler)
```

A "click" is defined as a short press down and release of the button.  It's not easy to write a polling loop to detect a click event, and even more difficult to do other things while checking for such a click (basically you need to write your own cooperative scheduler).

There are a few ways I can think to make the above code work.  They vary significantly in complexity and use.

**Simple events**
Queue events in the background and expose them using simple methods.  Eg: `button_a.was_clicked()` -- this will check to see if the button was clicked since the last time this function was called.  If it was then it returns True and resets the event state.  Otherwise it returns False.  In this case there are no event handlers, it's really just polling (but polling for button events, not simple button status).

**Explicit event handler**
Allow event handlers to be registered but force the user to periodically execute an "event check and handle" function.  Eg: 

``` python
def click_handler():
    print("you clicked the button!")
button_a.onclick(click_handler)

while True:
    microbit.event_handler() # check for pending events and execute the handlers if needed
    # do other stuff
```

This is cooperative "multitasking".  It makes the user aware that they need to do something special to get the event handlers to run, and gives control to the user when such callbacks are executed.

**Interrupt-like handler**
Allow event handlers to be registered and execute them like an interrupt (or Linux signal): when the event fires a short time later the main thread is paused and execution is passed to the relevant event handler.  Only when the event handler is finished does execution return to the main thread.  (The event handler would execute on the same stack as the main thread and no task switching is needed.)

This leads to less problems with data locking and deadlocks, but does expose a degree of preemptive issues to the kids.  Eg an event handler can modify a list that you are currently iterating through.

Technically this scheme would be implemented by the VM periodically checking for pending events and pausing the current bytecode when an event comes up.  That means event handling can't preempt a C/C++ function, but I think that's a good thing (keeps things simpler) and don't think it'll be a problem since all C/C++ functions eventually return (except panic!).

**Full monty**
Have actual threads that execute concurrently.  This is way too complex (for the kids, and for us to implement), and I don't think we even need the power/flexibility that this offers.

Please raise your opinion on this very interesting topic!

[...]

PR #129, opened 2015-12-05
Add an events module

inglesp:
NOT READY FOR MERGING!

This PR introduces an `events` module for reacting to events from Python code.  Events include things like a button being pressed, the accelerometer or compass taking on a certain value, or a timer popping.  There's some related discussion in #36.

The module can be used like this:

```
import microbit
import events

button_a_pressed = events.when_button_a_pressed()
button_b_pressed = events.when_button_b_pressed()

for event in events.events():
    if event == button_a_pressed:
        microbit.display.show(microbit.Image.HAPPY)
    elif event == button_b_pressed:
        microbit.display.show(microbit.Image.SAD)
```

A more complex example (an implementation of tetris) can be found in [this gist](https://gist.github.com/inglesp/78a492083303eda82af3), which can be compared with an implementation of the same game without an event loop in [this gist](https://gist.github.com/inglesp/2a25e597b25eef6e0129).

Implementation notes can be found in the comments at the top of modevents.cpp.

The code is not yet ready for merging.  At the very least, the following need to be addressed:
- If no event has occurred, `next(event.events())` yields `None`.  It would be much better if it instead blocked until it could yield something, but I'm not sure how to do this.
- ~~It should be possible to stop listening for an event.~~
- ~~There should be an event for a timer popping once, rather than just repeatedly.~~
- The bouncing behaviour from @larryhastings's [microbit_game](https://bitbucket.org/larry/microbit_game) could be implemented.
- The implementation could be made more efficient.  For instance, if there are several events that check the value of the accelerometer, the code only needs to query `uBit.compass.heading()` once.
- There should be help text for each of the available events.
- The code needs a thorough review from somebody who speaks C -- there's some pretty heavy cargo culting going on here.

However, I wanted to to make this available as soon as possible so that I could start to get feedback.


deshipu:
Have you considered having ``events.get()`` and ``events.wait()`` instead of a single ``events.events()``? This is how PyGame does it, and it lets you choose whether you want to do something else while waiting for the events, or just block. It's also consistent with the rule that function names should be verbs.

deshipu:
Another note, using constants for identifying the events, instead of some weird objects returned from factory functions, that you have to keep all the time in your program, would save some memory, which we don't really have much.

inglesp:
Hi @deshipu, thanks for your feedback.

`events.get()` is certainly doable -- in fact I think I originally had something like that.  A blocking function like `events.wait()` is probably also doable, but I'd require some advice from somebody who knows the code better -- I had problems getting blocking code to work properly.

As for these "weird objects"... they're actually just normal Python integers.  (I should've made that clear, sorry).  I'm not quite sure what you mean about using constants to identify the events.  Could you provide some example Python code?

deshipu:
I was thinking about something like:

```
if event == events.BUTTON_A_PRESSED:
    ...
```

of course that won't work with events that require parameters, for those you actually need variables:
```
tilt = events.register(source=events.ACCEL_Y, min=-2000, max=-500)
if event == tilt:
    ...
```

I'm not sure if having two different ways to do it is actually a good idea, but it does make the simple case simpler...

sparkslabs:
Sounds like you're after feedback :-) To me this looks really nice,
simple and cool :-)


I think having some mechanism to go:
import microbit
import events

button_a_pressed = events.when_button_a_pressed()
button_b_pressed = events.when_button_b_pressed()

while True:
    while events.no_events_ready():
         micro_sleep_of_some_kind()
    for event in events.events():
        if event == button_a_pressed:
            microbit.display.show(microbit.Image.HAPPY)
        elif event == button_b_pressed:
            microbit.display.show(microbit.Image.SAD)

Or events.wait_for_event() or similar would be useful. (The
no_events_ready() option though is useful because it allows you to do
other stuff - like animations - while waiting for the user)

That said if you exit the loop with no events, that tells you there
weren't any events :-)

@deshipu The "when" approach was discussed elsewhere, and the reason
for that rather than a word like register was simply because it's a
phrasing for events that kids are likely to be aware of due to how
Scratch works :-) (And while doing things a certain way "because
that's how something else does it" isn't always a good argument I
think it's useful here because it works *really well* in scratch)

inglesp:
Hi @deshipu -- I can see the argument for making the simple case simpler, but as far as I can see, it is only the button events that you could treat like this.  However, we've now got #101, which has added `. was_pressed()` for buttons, and so we already have a simple way of handling button events.

inglesp:
Hi @sparkslabs -- Thanks for your feedback.  I hadn't really thought about the something like `no_events_ready`, but at the moment iterating over `events.events()` yields `None` if there are no events ready.  That means you could do:

````
for event in events.events():
    if event is None:  # or maybe: if event == events.NOTHING_HAPPENED
        micro_sleep_of_some_kind()
    elif event == button_a_pressed:
        microbit.display.show(microbit.Image.HAPPY)
    elif event == button_b_pressed:
        microbit.display.show(microbit.Image.SAD)
````

I'm not sure though what you mean for `micro_sleep_of_some_kind()` though -- do you actually mean some kind of sleep?  If so, I'm not sure what the purpose of it is.

**Comment by ntoll** [comment]
@inglesp this is still a WiP? Care to rebase. Perhaps @dpgeorge has comments on it too. Not sure it'll make 1.1 unless it's properly finished, documented etc...

**Comment by inglesp** [comment]
Hi @ntoll!

Current status is that I'm waiting for guidance from project leadership (is that you?  @dpgeorge?) about whether this is:

 * the right kind of API;
 * the right kind of implementation;
 * worth doing at all.

Happy to pick this up again if it's likely to be accepted.

**Comment by dpgeorge** [comment]
We did want something like this, especially since it's a synchronous event manager (as opposed to async).  But I don't have the cycles at the moment for a full evaluation of it.

**Comment by inglesp** [comment]
This is now rebased.

**Comment by dpgeorge** [comment]
Thanks @inglesp.  This code takes up about 2000 bytes, which is not bad considering how much functionality it brings.

I'd say we should merge it, but consider it preliminary for now.  Ie the API can change, or even the whole module can be removed.  We'll let people play with it to see how it fits, and if it's useful, and then we can write docs and a tutorial and make it official.

What do people think about that?  @ntoll @markshannon ?

File: source/microbit/modevents.cpp [hunk 1]
```diff
@@ -0,0 +1,582 @@
...
+    return result;
+}
+
+STATIC mp_obj_t microbit_events_iter_next(mp_obj_t o_in) {
+    // Is there a way to mark o_in as unused?
+    uint16_t event_id;
+    microbit_events_obj->state = MICROBIT_EVENTS_LOOP_RUNNING;
+    event_id = event_queue_dequeue(microbit_events_obj->event_queue);
+    if (event_id == MICROBIT_EVENTS_NO_EVENT) {
+        // Is there a way to indicate that the iterator should not yield?
```
markshannon: `__WFI();` or `__WFE();`, it depends on interrupt level. Probably `__WFE();` in this case.

markshannon:
I'd like to see the API, ideally with an example or two.

What happens when the consumer of events does not keep up with the production of events, will the event queue consume all of the memory?
A fixed size buffer would save quite a lot of memory and do the "right thing" when there are too many events, which is to drop the oldest events.
I'm also worried that it does a lot of memory allocation in interrupt handlers, and is likely to cause GC pauses which cannot be interrupted.

Why not push events onto the queue when they occur, rather than scanning for them later?
For example, button events could be added to the queue here:
https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/microbitbutton.cpp#L72

dpgeorge:
Closing due to lack of activity.

PR #130, opened 2015-12-10
Allow music.play() to use different tempos

nceder:
It seems to me that part of the fun in playing music is being able to play around with the tempo of tunes. The previous implementation had in effect hard coded **all** tunes to be played with a tempo of 120, 4.

With this patch, I've added an optional keyword parameter to music.play() called reset, with a default value of True. 

Now, if the tempo is reset, and play() is called with reset=False the new tempo will be used, often with comic results. I promise you, once a room full of 11 year-olds discovers this feature, there will be shrieks of delight and the most splendid chaos. :-)

(I might add that it took a room full of adults in Chicago less than 10 minutes of playing with the music module to suggest this feature. :-) )


ntoll:
Hahahaha... you're a very naughty girl Naomi. That's awesome. If the code's OK I say merge it. :-D

deshipu:
Wouldn't it be more intuitive to have a ``tempo`` parameter, to which it would get set while playing?

ntoll:
@deshipu no... that's a global state set with (IIRC) `set_tempo`.

deshipu:
Right. Global state makes everything so much easier. Imagine I'm a new user exploring the API. I find the ``set_tempo`` function and I want to try it. I call it, and try to play a song. No change. I scan the available functions and maybe even parameters, but see nothing related to tempo. I give up.

ntoll:
So, I can play music in the background and use `set_tempo` to change the speed of the performance in reaction to some event. As a robot speeds up the music gets faster etc... :-)

That's a really awesome effect and you can't do that with a field passed in at the start. 

Also, it removes complexity from the `play` method and cleanly separates the notion of playing music from tempo. Finally, setting the tempo requires two fields anyway (see the docs).

Hope that makes it clear! :-D

deshipu:
> That's a really awesome effect and you can't do that with a field passed in at the start.

Of course you can. You just return an object from the play function, and set tempo on that object. Globals are evil.

sparkslabs:
On 10 December 2015 at 09:43, Radomir Dopieralski <notifications@github.com>
 wrote:

> That's a really awesome effect and you can't do that with a field passed
> in at the start.
>
> Of course you can. You just return an object from the play function, and
> set tempo on that object. Globals are evil.
>
If you can only play one piece of music at a time, you have a *logical*
global.  How you access it/store it is another matter.

That said, I don't see an issue with having:

.play(tempo=initial value)

As well as .set_tempo()

After all, doesn't music normally specify a tempo? (and speeding up/slowing
down seems reasonable)

Not a musician, though I can bang keys and blow through things, and make
horrible noises :-)


Michael.

ntoll:
Strictly speaking, the module *is* a (certain sort of) object. What Michael says about a logical global state is correct. I also think there should be one obvious way to do something (like change the tempo). That's what we have. ;-)

nceder:
Well, I was hoping to keep the changes minimal and simple. I had considered the notion of passing in an initial tempo, but decided it wasn't worth the trouble. 

I guess I viewed my approach as a reasonable compromise between being pythonic ("Explicit is better than implicit") and doing as little as possible and keeping the basic function as simple as possible. TBH I'm not personally offended by the module global music state. Now, if it becomes possible to play different tunes on different pins simultaneously, I would probably *ahem* change my tune. ;-)

However it's done, I'd still strongly argue that playing tunes with different tempos should be possible.

ntoll:
I'm a +1 on this change as it stands but @dpgeorge is, as always, final arbiter of this sort of thing.

deshipu:
Sorry, I'm used to Python code on big computers, that generally tends to grow and have problems with such things as global state. After some thought, I agree that it totally make sense on such a small device, where you can only ever play one thing at a time and are never going to write tests.

dpgeorge:
> Imagine I'm a new user exploring the API. I find the set_tempo function and I want to try it. I call it, and try to play a song. No change. I scan the available functions and maybe even parameters, but see nothing related to tempo. I give up.

I agree with this.

In fact, I think set_tempo is useless because the bpm/ticks that are set with set_tempo are only ever used by play, and are reset at its beginning.

I think the intention of the reset was to make sure the octave and duration was reset (so the tune is in the same key), but not to reset the bpm and ticks.

So I'd propose to simple make it so that play resets only the current octave and duration, and not the bpm or ticks.

nceder:
Okay, following Damien's suggestion, I'm closing this PR. 

If I get a chance I'll patch as he says.

PR #131, opened 2015-12-10
doco improvements

Nathanator:
Fixing monospacing
Cleaned up some whitespace
Added base section on running code

### DIFF
--- a/docs/devguide/flashfirmware.rst
+++ b/docs/devguide/flashfirmware.rst
@@ -32,10 +32,11 @@ some convenience targets.
 Preparing firmware and a Python program
 ---------------------------------------
 
-Try these::
+A script in the following location, called using the following syntax, should help::
 
-  tools/makecombined
-  hexlify
+  tools/makecombinedhex.py <firmware.hex> <script.py> [-o <combined.hex>]
+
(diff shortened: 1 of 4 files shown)


docs/devguide/flashfirmware.rst
```diff
@@ -33,15 +32,17 @@ some convenience targets.
...
 ---------------------------------------
 
-tools/makecombined
-
-hexlify
-
+Try these::
 
+  tools/makecombined
+  hexlify
```
dpgeorge: I think with makecombined.py you don't need hexlify.py, so the latter doesn't need to be mentioned.  You could actually mention in more detail how to use makecombined, eg: ./tools/makecombined.py build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython.hex myscript.py

Nathanator: Cheers, didn't realise that existed.  I'll update.

dpgeorge:
@Nathanator thanks for fixing, but can you please rebase to the latest master so that the commit is clean?  Thanks!

Nathanator:
Will do.  Git amateur, unfortunately, but I'm sure I can work it out :-)

PR #135, opened 2015-12-13
Remove bpm, ticks from music.reset()

nceder:
Following Damien's proposal this patch simply removes bpm and ticks from the music.reset() function. Therefore, any changes made using set_tempo() will be reflected when the next tune is played.

### DIFF
--- a/docs/music.rst
+++ b/docs/music.rst
@@ -105,8 +105,6 @@ Functions
 
     Resets the state of the following attributes in the following way:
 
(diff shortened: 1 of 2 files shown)


dpgeorge:
Actually, what I meant (sorry if it wasn't clear!) was that music.reset() should remain as-is, but that music.play() should not call reset(), but instead should just reset last_octave and last_duration.

PR #136, opened 2015-12-13
reset only octave and duration to defaults before play()

nceder:
Reset only duration and octave in music.play() to allow different tempos.


dpgeorge:
Thanks!  @ntoll any concerns?

ntoll:
None... LGTM :-)

PR #141, opened 2015-12-22
Add working neopixel module

dpgeorge:
This PR adds a neopixel module, with tested and working neopixel support.  To get reliable control of a long chain of neopixels, interrupts must be disabled while outputing the bit stream (this is done in the neopixel_show function).

To use:
```python
import neopixel
neopixel.init(pin0, 8) # 8 leds driven on pin0
for i in range(8):
    neopixel.set_colour(i, 0, 0, i * 32) # make pixel blue, increasing brightness
neopixel.show()
```

I've implemented the neopixel as a singleton object with global state.  This kind of follows what's done in the music module.  But probably it's much better to make it a true object, so you can have multiple chains of neopixels on different pins (entirely possible to do).  Then the API might be something like:
```python
import neopixel
np0 = neopixel.NeoPixel(pin0, 8)
np1 = neopixel.NeoPixel(pin1, 16)
for i in range(8):
    np0[i] = (0, 0, i * 32)
np0.show()
```

@whaleygeek You can use this branch if you'd like to try out the neopixel support.

### DIFF
--- a/source/microbit/modneopixel.cpp
+++ b/source/microbit/modneopixel.cpp
@@ -29,6 +29,7 @@
 
 extern "C" {
 
+#include "py/runtime0.h"
 #include "py/runtime.h"
 #include "lib/neopixel.h"
 #include "microbitobj.h"
@@ -56,6 +57,14 @@ STATIC mp_obj_t neopixel_make_new(mp_obj_t type_in, mp_uint_t n_args, mp_uint_t 
     neopixel_init(&self->strip, pin, num_pixels);
 
     return self;
+}
+
+STATIC mp_obj_t neopixel_unary_op(mp_uint_t op, mp_obj_t self_in) {
+    neopixel_obj_t *self = (neopixel_obj_t*)self_in;
+    switch (op) {
+        case MP_UNARY_OP_LEN: return MP_OBJ_NEW_SMALL_INT(self->strip.num_leds);
+        default: return MP_OBJ_NULL; // op not supported
+    }
 }
 
 STATIC mp_obj_t neopixel_subscr(mp_obj_t self_in, mp_obj_t index_in, mp_obj_t value) {
@@ -114,7 +123,7 @@ const mp_obj_type_t neopixel_type = {
     .print = NULL,
     .make_new = neopixel_make_new,
     .call = NULL,
-    .unary_op = NULL,
+    .unary_op = neopixel_unary_op,
     .binary_op = NULL,
     .attr = NULL,
     .subscr = neopixel_subscr,


ntoll:
@dpgeorge I have a neo-pixel strip arriving tomorrow. ;-) Looking forward to testing this. I think the latter API (i.e. instantiate an object for each pin) gives us the most flexibility and potential for amazing effects. ;-)

Great stuff.

dpgeorge:
Yes, I much prefer the latter API with subscription to set the pixel.  Let's try to get this API worked out soon se we can merge this :)

Let me know how the tests go!  You may need to power from 5v, but 3v from the microbit edge connector could work (I tested only 5v power, maximum 8 pixels).

Neon22:
I never considered just removing the 50microsecond delay. I spent
    ages trying to work out how to make one at the C level :( Sigh. So
    glad you got it working so easily Damien.
    
    FWIW there is one disadvantage to the current API. You can't have
    long arrays of LEDs as memory for each one has to be initially
    allocated. I had considered adding a second API (maybe
    Another useful way to define it is:
        - Define the max number of simultaneously active leds based on
    your memory capability.
        - Store the index to each led in the LED strip.
    This way you could have long string (256 max ?) that you would not
    light up more than 32(say) at a time.
    
    Using this API you could send a "pulse" along the string by updating
    the index of the led you were writing to, as well as its rgb value.
    This interface is also suited to stationary displays like scrolling
    a short message along a long strip folded over to form a matrix
    display. Controlled by shifting indexes and using the same rgb
    values.
    
    startpos = 12
np0 = neopixel.NeoPixel_indexed(pin0, 8, 240) # max 8 leds in 240 led strip - driven on pin0
for i in range(8):
  pos = startpos + i  
  np0[i] = (pos, 0, 0, i * 32) # index, R,G,B
np0.show()

    
    
    Cheers, Mark...
    
     
    On 12/23/2015 5:39 AM, Damien George
      wrote:
    
    
      Yes, I much prefer the latter API with subscription to set the
        pixel. Let's try to get this API worked out soon se we can merge
        this :)
      Let me know how the tests go! You may need to power from 5v,
        but 3v from the microbit edge connector could work (I tested
        only 5v power, maximum 8 pixels).
      —
        Reply to this email directly or view
          it on GitHub.
      
        
          
          
        
        
      
      No virus
        found in this message.
        Checked by AVG - www.avg.com
        Version: 2016.0.7294 / Virus Database: 4489/11233 - Release
        Date: 12/22/15

deshipu:
Speaking of memory use, I wonder if it really needs to store the pixel values, and not just iterate over provided object to get them -- then we can have an iterator that calculates them on the fly, with almost no memory used for some of the patterns.

ntoll:
@deshipu s/iterator/generator ???

deshipu:
Well, if you accept any iterator, then you can use a generator function. Or an iterator comprehensions. Or  a custom iterator class.

ntoll:
Good point... :-) My thinking was that the generator is the memory efficient thing to use for obvious on-the-fly reasons.

deshipu:
Another use case for this is if you build a screen of neopixels, you may want to keep your pixel data in a list of lists. Then a simple generator function can feed that to the display routine.

dpgeorge:
@Neon22 the neopixel branch that I pushed some time ago was actually working, so I'm not sure what problems you had with it...

The ideo to support just a chunk of pixels within the chain is neat, but I think complicates things for the user.  I'd say 64 neopixels is already a lot, and that only needs 64*3=196 bytes of RAM to store the colours in the buffer.  We have enough RAM for that!

@deshipu there is no way the microbit can generate pixel colours fast enough on the fly... even writing a generator in C is not possible (well, for anything non-trivial).  If you look at the driver, it uses assembly to do the timing, and each pixel has at best a run of 8 nops per bit (with 24 bits per pixel).  There's not much you can do in 8 instructions.  You could hard code a specific generator in assembler, but not a generic function call, let alone a generic Python function call.

If we need to reduce RAM, first thing would be to compress the colours to use 16 bits instead of 24.  Then uncompress on the fly (should be capable of that).

Neon22:
Alas I never got any signals appearing at the pins.
I agree having the indices will complicate things for the user which is why I suggested a differently named interface so the user could choose either the regular neopixel or the neopixel_indexed class.

If I have three 5m long strips of neopixels for my game I can't allocate enough RAM, but as I'm only ever writing to about 32 of them at one time - I could do so if I had indexed access.

I think the length of the string is probably not critical as long at the routine can count as high as the highest index. So the extra arg in the init function may not be required but not certain about this.
Could also constrain the python array to force all indices to be > previous one so that C code counting would be simply be counting until current index is reached then stepping to next entry in array. Halting when end of array, or index, is lower than current count.
It would be up to the user to sort() before calling show() or for show() to sort before calling passing control to C.

ntoll:
Hi @dpgeorge, unsurprisingly, I can get it to work. Looking forward to the updated API. ;-)

See: 

https://www.youtube.com/watch?v=qFtLLDEa6Mw

...with this code (inspired by @Neon22 ~ also includes sound effects :smiley:):

```
# Play a game of squeeky pong with a 1m (30 led) NeoPixel strip. :-)
from microbit import *
import music
import neopixel


neopixel.init(pin1, 30)


def show_pixel(position, wait):
    neopixel.set_colour(position, 0, 0, 64)
    neopixel.show()
    music.pitch(440 + position * 16, wait)
    neopixel.set_colour(position, 0, 0, 0)


def run():
    pos = 0
    score = 0
    pixels = [x for x in range(30)]
    pause = 100

    while True:
        if button_a.is_pressed():
            for pos in pixels:
                show_pixel(pos, pause)
                pos += 1
            for pos in pixels[::-1]:
                show_pixel(pos, pause)
                pos -= 1
            score += 1
            pause -= 5
            display.show("{}".format(score))
        else:
            display.scroll("Score: {}".format(score))
            break

while True:
    if button_a.is_pressed():
        run()
```

Neon22:
def show_pixel(position, wait):
    neopixel.set_colour(max(position-2,0), 5, 0, 64/4)
    neopixel.set_colour(max(position-1,0), 10, 0, 64/2)
    neopixel.set_colour(position, 0, 0, 64)
    neopixel.show()
    music.pitch(440 + position * 16, wait)
    neopixel.set_colour(max(position-2,0), 0, 0, 0)
    neopixel.set_colour(max(position-1,0), 0, 0, 0)
    neopixel.set_colour(position, 0, 0, 0)
Then try thresholding the accelerometer instead of looking for the button press. (attach microbit to board/card. hit card with pencil).

ntoll:
@dpgeorge any more movement on this..?

Combined with the new gestures work would make it very easy to do @Neon22's 1 dimensional pong. ;-)

gbaman:
After playing with it today, I have a few issues (and a request) to throw in.
First, the below code causes pixel 1 to go to full brightness. Pretty sure it should either throw an exception or be ignored.
```python
import neopixel
neopixel.init(10)
neopixel.set_colour(1, -1, -1, -1)
neopixel.show()
```
Next, I have hit am assuming memory issues. I have [quite easily been able to control 300 pixels](https://www.youtube.com/watch?v=ZDGsjfmoIH4) at a time and have them cycling, but when I add other stuff into the mix that uses some RAM, the behavior of the pixels gets extremely random. For example [using this twinkle generator code](https://gist.github.com/gbaman/2f7da17c0145cde3d108) works [great up to about 120 pixels](https://www.youtube.com/watch?v=Kl5sPIPOZLI), but after going above that I start getting [random pixels usually near the end that appear random colours very bright](https://www.youtube.com/watch?v=DRIJOPQToI8).

Finally then, a request.
Any chance of getting some way to read the current RGB value of a pixel? Would have made stuff like the twinkle program far easier to be able to get the colour value of a pixel, given it is being stored anyway.

ntoll:
@gbaman wow +100000 on the amazingly cool videos. 

I guess @dpgeorge is the one to comment about potential memory issues.

Neon22:
If I may so bold as to suggest that:
- the class vars are not used and 
- change lower() to have an amount parameter.
The subtraction may be overflowing to 255.
You can read apixel value by asking pixels for their rgb values, or make a new class function called (say) read_pixel() which might return an rgb list...

Like:
total_pixels = 120 #Total number of pixels
bright = 60 #Max brightness
max_pixels = 20 #Number of allowed twinkles at any one time

neopixel.init(pin0, total_pixels)

class pixel():

    def __init__(self, id, r, g, b):
        self.r = r
        self.g = g
        self.b = b
        self.id = id
    def lowerer(self): #Decrease each colour brightness of the pixel
        self.r = make_lower(self.r)
        self.g = make_lower(self.g)
        self.b = make_lower(self.b)
    def check_finished(self): #Check if pixel needs to be removed
        if (self.r < 1) and (self.g < 1) and (self.b < 1):
            return True
        else:
            return False

def make_lower(data, amount=1): #Decrease pixel brightness if can be done
    if data > amount-1:
        data = data - amount
    return data

def reduce_pixels(pixels): #Decrease the brightness of each pixel and delete dead ones
    for p in pixels:
        p.lowerer(2)
    for p in pixels:
        if p.check_finished():
            neopixel.set_colour(p.id, 0,0,0)
            pixels.remove(p)

def write_pixels(pixels): #Output the pixels
    for i in range(0, len(pixels)):
        neopixel.set_colour(pixels[i].id, pixels[i].r, pixels[i].g, pixels[i].b)
    neopixel.show()

def add_twinkle_pixel(pixels): #Add new twinkle pixel
    if len(pixels) < max_pixels:
        r = random(bright)
        g = random(bright)
        b = random(bright)
        id = random(total_pixels)
        pixels.append(pixel(id, r, g, b))
    return pixels



def twinkler():
    pixels = []
    while True:
        pixels = add_twinkle_pixel(pixels)
        reduce_pixels(pixels)
        write_pixels(pixels)

twinkler()

dpgeorge:
@ntoll I'll get to this as soon as I can :)

@gbama thanks for the feedback!  Passing out-of-range rgb values probably should raise an exception, just like all the other methods (eg random, pin.write_analog).  Personally I'd consider this unnecessary bloat, but for something like the microbit we want to help the kids as much as possible to write correct code, and raising exceptions helps that.

As part of the refactoring to use objects and [] access I can easily add functionality to read back a pixel colour.

gbaman:
@dpgeorge Excellent. I think as a minimum if they put in a minus number it would be ignored (instead of blinding them), but I think an exception is a better idea to help them find errors faster.

@Neon22 The code was thrown together very very quickly, in parts simply to test with. I am more than happy to admit it isn't great. 
The code you provided unfortunately does not actually work though as the pixels never actually die. I will have a look over it later when I have some free time.

dpgeorge:
I updated the neopixel module.  New API:
```python
import neopixel

# construct driver on a pin
np = neopixel.NeoPixel(pin0, 8)

# clear all pixels
np.clear()

# show pixels
np.show()

# write colour (doesn't update display, use show() for that)
np[0] = (255, 0, 128) # first element
np[-1] = (0, 255, 0) # last element

# read colour
print(np[0])
```
It raises ValueError if the rgb input values are not in the range [0..255].

I didn't test multiple drivers on different pins at the same time, but it should work :)

The NeoPixel class could probably do with a method to query the number of pixels, eg NeoPixel.num_pixels()?

There is an issue with the driver not working sometimes... I think it's because the nrf51 changes its clock frequency from 8MHz to 16MHz at some point but I'm not sure.....

dpgeorge:
> The NeoPixel class could probably do with a method to query the number of pixels, eg NeoPixel.num_pixels()?

Of course, it should be `len(np)` to give the number of pixels.

ntoll:
Care to rebase and squash? I'll check and merge..!

dpgeorge:
Rebased.  I didn't squash because the commits themselves are important to separate out (well, except maybe the last 2, but it's nice history).  Feel free to disagree :)

dpgeorge:
Hang on, should we implement the len function to return number of pixels?

ntoll:
If it's quick to do... I'm on a test/review/merge bender this evening... ;-)

dpgeorge:
Ok, just pushed a commit to add len(neopixel) support.  It compiles but is untested!  Please test :)

inc/lib/neopixel.h
```diff
@@ -0,0 +1,133 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "nrf_delay.h"
+#include "nrf_gpio.h"
+
+
+//These defines are timed specific to a series of if statements and will need to be changed
+//to compensate for different writing algorithms than the one in neopixel.c
+#define NEOPIXEL_SEND_ONE	NRF_GPIO->OUTSET = (1UL << PIN); \
```
markshannon: This macro should take `pin` as a parameter, rather than trusting that `PIN` is defined correctly.

inc/lib/neopixel.h
```diff
@@ -0,0 +1,133 @@
...
+				" NOP\n\t" \
+				" NOP\n\t" \
+				" NOP\n\t" \
+				" NOP\n\t" \
+				" NOP\n\t" \
+				" NOP\n\t" \
+			); \
+		NRF_GPIO->OUTCLR = (1UL << PIN); \
+
+#define NEOPIXEL_SEND_ZERO NRF_GPIO->OUTSET = (1UL << PIN); \
```
markshannon: Ditto

source/lib/neopixel.c
```diff
@@ -0,0 +1,181 @@
...
+	const uint8_t PIN =  strip->pin_num;
+	NRF_GPIO->OUTCLR = (1UL << PIN);
+	nrf_delay_us(50);
+	uint32_t irq_state = __get_PRIMASK();
+	__disable_irq();
+			for (int i = 0; i < strip->num_leds; i++)
+			{
+				for (int j = 0; j < 3; j++)
+				{
+					if ((strip->leds[i].grb[j] & 128) > 0)	{NEOPIXEL_SEND_ONE}
```
markshannon: Since each macro expansion accounts for 30-40 bytes it would save a few hundred bytes to use a loop here.
Something like 
```C
bit = 128;
do {
     if(strip->leds[i].grb[j] & bit) ...
    bit >>= 1;
} while (bit);
```
should be a tight enough loop to not throw out the timing too much.

source/microbit/modneopixel.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+STATIC mp_obj_t neopixel_unary_op(mp_uint_t op, mp_obj_t self_in) {
+    neopixel_obj_t *self = (neopixel_obj_t*)self_in;
+    switch (op) {
+        case MP_UNARY_OP_LEN: return MP_OBJ_NEW_SMALL_INT(self->strip.num_leds);
+        default: return MP_OBJ_NULL; // op not supported
+    }
+}
+
+STATIC mp_obj_t neopixel_subscr(mp_obj_t self_in, mp_obj_t index_in, mp_obj_t value) {
```
markshannon: Should we be consistent with `display` and use `get_pixel`, `set_pixel`?

inc/lib/neopixel.h
dpgeorge: I want to minimise changes to the neopixel driver because it's from an upstream project (see comment at top of .h file).  You can see in the commits that I added the pristine source files, then made baby-step modifications, just enough so that it would work for us.  No need to make unnecessary changes for code style.  It works as it is :)

source/lib/neopixel.c
dpgeorge: As above, I don't want to change anything that doesn't need changing.  It works.  Timing is critical and it'll need some effort to get it right and test it.

source/microbit/modneopixel.cpp
dpgeorge: See the discussion at the very start of this PR.

dpgeorge:
I will rebase this once we come to a consensus on above comments.  Otherwise the comments disappear.

markshannon:
For reference:

Datasheets:
https://learn.adafruit.com/downloads/pdf/adafruit-neopixel-uberguide.pdf
https://www.adafruit.com/datasheets/WS2812.pdf

Allowable jitter is plus/minus 150ns which is only 2.5 cycles, so getting the timing right is tricky.

markshannon:
I'm OK with merging this. We can always improve it later if there is a need.

dpgeorge:
I am open to changing the pixel accessor from [] to set_pixel/get_pixel methods.  It is more consistent with the display.  On the other hand, neopixels are advanced things and [] does feel more natural/Pythonic.

ntoll:
Let's leave this as it is. I'm at "real work" right now so don't have my neopixels to hand to test / play. I'll merge this once I've done so this evening.

dpgeorge:
I rebased.  I also checked that len(neopixel) works.

ntoll:
Lots of fun to test. Lots of "oooOOOoohs" from my kids. :-)

ntoll:
@gbaman now that neopixels are in master... I've created my own "twinkler"....

```
# NeoPixel Twinkler
# (c) 2016 Nicholas H.Tollervey
from microbit import *
import neopixel


def fade(pixel, fade_by):
    # Fade a given pixel by fade_by amount. A pixel is a tuple of
    # three numeric values representing RGB.
    new = []
    for i in pixel:
        dimmed = i - fade_by
        if dimmed < 0:
            new.append(0)
        else:
            new.append(dimmed)
    return tuple(new)
    

def set_pixels(np, pixels, fade_by):
    # Iterate over the pixels, fade them by fade_by and update and show
    # their state on the neopixel strip.
    for i, p in enumerate(pixels):
        pixels[i] = fade(p, fade_by)
        np[i] = pixels[i]
    np.show()
      
      
def add_pixel(pixels, brightest):
    # Given a list of pixels and a max starting brightness add a new 
    # pixel to an empty slot in the list.
    
    # Get the indexes of the empty slots.
    empty = [i for i in range(len(pixels)) if not sum(pixels[i])]
    if empty:
        # If there are empty slots add a new pixel of random colour
        # to one of the empty slots.
        new_position = empty[random(len(empty))]
        pixels[new_position] = (random(brightest), random(brightest), 
                                random(brightest))
    

def run(pause=20):
    # The length of the neopixel strip.   
    size = 30  
    # The max value for brightness.
    brightest = 255
    # Represents the neopixel strip attached to the specified pin.
    np = neopixel.NeoPixel(pin0, size)
    # Reset to blank.
    np.clear()
    # Create an empty array of "off" tuples for each pixel.
    pixels = [(0, 0, 0) for i in range(size)]  
    while True:
        # Add a new pixel in a random empty slot at a random colour.
        add_pixel(pixels, brightest)
        # Fade the pixels on the neopixel strip by a factor of 8.
        set_pixels(np, pixels, 8)
        # Wait for pause milliseconds.
        sleep(pause)

run()
```

ntoll:
https://www.youtube.com/watch?v=ZEd9LNlw3ZI :-)

(I added button and gesture control just now)

PR #144, opened 2015-12-28
Just a simple random student name picker to use for questioning in the classroom.

marcotranchino:


deshipu:
Actually, I wonder if the examples should have proper, importable, python module names -- with underscores.

marcotranchino:
OK, sounds good to me.

ntoll:
Hi @marcotranchino the API has changed since this PR was created. For example, we've removed `microbit.random` and have a `random` module instead, including a `random.choice` method. Also, animations are a lot nicer / easier. Full docs here: http://microbit-micropython.readthedocs.org/en/latest/ Happy to help...

marcotranchino:
No problem, I will change it. 
I have been using this program with 5 classes, with a pupil per class in charge of changing the names in the code and re-deploy. It works well, the students like it.

PR #145, opened 2015-12-31
Add quaver image and correct crotchet note image.

ntoll:
I noticed the note image was mis-named. Corrected it and used the old image as the new "quaver" image.


dpgeorge:
Merged!

Issue #147, opened 2016-01-09
BUG - random(-n) always returns -1001


ntoll:
This is what I see:

```
random(-3)
-1001
```

I was expecting a number between 0 and -3.

whaleygeek:
I *think* -1001 is a DAL error code (all the error codes were re-mapped to
-1000 and below in a change before Christmas).


___________________________________________________________
David Whale, B.Sc (Hons), MIET
*Software Engineer and IET Schools Liaison Officer, Essex*

email:  dwhale@theiet.org
twitter: @whaleygeek
blog:  blog.whaleygeek.co.uk

Co-author of global bestselling book "Adventures in Minecraft"
<http://amzn.to/ZGfxZG> - lets get kids coding!
Now in English, Spanish, Chinese and Russian.

ntoll:
That's what I wondered. Personally, I think we should only accept positive integers and you negate the result if you want a minus number. Anything < 0 results in Python's `ValueError`.

dpgeorge:
Yes, it's due to the DAL rejecting inputs 0 or below.

I agree we should raise a ValueError for inputs 0 or below.

markshannon:
Fixed.

PR #150, opened 2016-01-13
Various fixes

### LINKED ISSUE #137: write_analog doesn't accept a float
The [docs say](https://microbit-micropython.readthedocs.org/en/latest/pin.html#microbit.MicroBitAnalogDigitalPin.write_analog):
> write_analog(value)
> Output a PWM signal on the pin, with the duty cycle proportional to the provided value. The value may be either an integer or a floating point number between 0 (0% duty cycle) and 1023 (100% duty).

But:
```python
>>> microbit.pin0.write_analog(0.5)
TypeError: can't convert float to int
```

Also just to really confuse things:
```python
>>> help(microbit.pin0.write_analog)
micro:bit, write_analog(value) to the pin. You can use a whole number between
0 and 255 as the value.
```

Let's decide if floats should be allowed or not (I'd say *why not?*), and fix either the docs (which I can do a patch for) or the implementation.

dpgeorge:
There are 4 small commits here to fix various problems.  See message of each one for more detail.

Fixes #137 and #147.

@ntoll if you don't mind merging :)

### DIFF
--- a/source/microbit/microbitpin.cpp
+++ b/source/microbit/microbitpin.cpp
@@ -59,7 +59,7 @@ MP_DEFINE_CONST_FUN_OBJ_1(microbit_pin_read_digital_obj, microbit_pin_read_digit
 mp_obj_t microbit_pin_write_analog(mp_obj_t self_in, mp_obj_t value_in) {
     microbit_pin_obj_t *self = (microbit_pin_obj_t*)self_in;
     int set_value;
-    if (MP_OBJ_IS_TYPE(value_in, &mp_type_float)) {
+    if (mp_obj_is_float(value_in)) {
         mp_float_t val = mp_obj_get_float(value_in);
         set_value = val+0.5;
     } else {
--- a/source/microbit/main.cpp
+++ b/source/microbit/main.cpp
@@ -63,6 +63,12 @@ static void ticker(void) {
 
 extern "C" {
 
+// We need to override this function so that the linker does not pull in
+// unnecessary code and static RAM usage for unused system exit functionality.
+// There can be large static data structures to store the exit functions.
+void __register_exitproc() {
+}
+
 void microbit_init(void) {
     microbit_display_init();
 
--- a/source/microbit/modmicrobit.cpp
+++ b/source/microbit/modmicrobit.cpp
@@ -28,6 +28,7 @@
 
 extern "C" {
 
+#include "py/nlr.h"
 #include "py/obj.h"
 #include "py/mphal.h"
 #include "modmicrobit.h"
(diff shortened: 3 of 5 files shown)


ntoll:
Any guidance on how I could/should test these?

dpgeorge:
The float and random patch are easy, just try them on the microbit :)

The others are very hard to test:

- To test the __register_exitproc patch you need to install arm-none-eabi-gcc v5.3.0 (see https://github.com/lancaster-university/microbit-dal/issues/72).
- To test the memset GC patch you need a test program from the mailing list (waves.py) and run it for a few hours before and after the patch.  I also checked that it work using a lot of other heavy test machinery that I have for upstream MicroPython, include some stuff done for the ESA project.

I don't know how to deal with these situations, merging stuff like this.  I put a *lot* of work into bug fixes like this, which end up being a couple of lines of code, so may not look like much.  But believe me I do a lot of proper testing.  I think it's a burden having to review small patches, since pretty much all of us do this in our spare time and we spread our resources really thin this way.  @markshannon your opinion is appreciated here.

markshannon:
I prefer small patches to review, but I wouldn't count 28 lines as large :smile:
I'll take a look at the GC and __register_exitproc commits.

dpgeorge:
> I prefer small patches to review,

The PR is broken into 4 separate commits, each of 10 or so lines or less, with proper commit messages.

ntoll:
@dpgeorge totally agree with time/effort put into these fixes. Your work is truly appreciated..! I'll check the two easy fixes in a moment.

ntoll:
I've just tested the random and float fixes and can confirm "they work for me!" :-)

Over to you @markshannon

source/microbit/microbitpin.cpp
```diff
@@ -59,7 +59,7 @@ MP_DEFINE_CONST_FUN_OBJ_1(microbit_pin_read_digital_obj, microbit_pin_read_digit
 mp_obj_t microbit_pin_write_analog(mp_obj_t self_in, mp_obj_t value_in) {
     microbit_pin_obj_t *self = (microbit_pin_obj_t*)self_in;
     int set_value;
-    if (MP_OBJ_IS_TYPE(value_in, &mp_type_float)) {
+    if (mp_obj_is_float(value_in)) {
```
markshannon: OOI, is this necessary as we now use tagged floats, rather than heap-allocated ones?

dpgeorge: Yes.  mp_obj_is_float is the proper way to check if an obj is a float.

markshannon:
I've verified that [waves.py](https://gist.github.com/markshannon/cfaeb4d8a6b72e46cb24
) doesn't fail with the fix.
However, it doesn't fail on master either :confused:

The code looks good and the GC fix appears sound, even if doesn't fix the root cause of the memory leak.

I'm inclined to merge this anyway, but it would be nice to have a test case.

dpgeorge:
Some small variations on waves did and did not fail on master.  Changing the size of the script by simply adding a comment can do it.  Fortunately I found something rather reproducible and that was fixed.  I also inspected the heap (on host not microbit) and found the root cause was exactly that there were stray pointers in previously allocated strings that were not being cleared to zero.

markshannon:
OK. Merging.

dpgeorge:
Great, thanks!

Issue #148, opened 2016-01-12
pin0.is_touched() always returns False


ntoll:
If I'm touching the pin, it should return True.

ntoll:
This also applies to the other pins too.

ntoll:
Looking at https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/microbitpin.cpp#L98 it's just a wrapper around a function in the DAL. :-/

dpgeorge:
I think you need to touch ground of the microbit with your other hand.  I'm pretty sure I got is_touched() working before.

whaleygeek:
Yes, we were using this at the club the other week, it does work, but you
do have to hold GND.

It's resistive touch, (not capacitive touch as the original spec blog page
says), so it needs a path to ground in order to work.

This just about works between two people if the grounds are close and on
the same arms as touching (see the Love-ometer resource that the BBC
created using this).

https://www.microbit.co.uk/td/lessons/love-meter

The front picture on the video in the above link shows how you have to make
the circuit for it to work.



___________________________________________________________
David Whale, B.Sc (Hons), MIET
*Software Engineer and IET Schools Liaison Officer, Essex*

email:  dwhale@theiet.org
twitter: @whaleygeek
blog:  blog.whaleygeek.co.uk

Co-author of global bestselling book "Adventures in Minecraft"
<http://amzn.to/ZGfxZG> - lets get kids coding!
Now in English, Spanish, Chinese and Russian.

ntoll:
@dpgeorge @whaleygeek I've been touching GND while holding pin0. Can't get anything other than False out of it.

whaleygeek:
What happens if you follow the touch develop resource I linked you to -
does that work? That worked for me a couple of weeks before Christmas,
although there has been another drop in the last few days that is
percolating it's way up.

i.e. is it a runtime problem, or a python problem?

Also, can you confirm what version of the runtime microPython is built
with, and what version of the runtime the latest TouchDevelop is built
with, in case of a difference there.

___________________________________________________________
David Whale, B.Sc (Hons), MIET
*Software Engineer and IET Schools Liaison Officer, Essex*

email:  dwhale@theiet.org
twitter: @whaleygeek
blog:  blog.whaleygeek.co.uk

Co-author of global bestselling book "Adventures in Minecraft"
<http://amzn.to/ZGfxZG> - lets get kids coding!
Now in English, Spanish, Chinese and Russian.

[...]

PR #152, opened 2016-01-16
Fix issue 148

markshannon:
Update all components in ticker, not just buttons a and b.
This ensures that touch pins (which are button objects under the covers) are updated.
Fixes https://github.com/bbcmicrobit/micropython/issues/148

### DIFF
--- a/source/microbit/main.cpp
+++ b/source/microbit/main.cpp
@@ -50,9 +50,12 @@ static void ticker(void) {
     compass_up_to_date = false;
     accelerometer_up_to_date = false;
 
-    // Update buttons
-    uBit.buttonA.systemTick();
-    uBit.buttonB.systemTick();
+    //update any components in the DAL's systemComponents array
+    for (int i = 0; i < MICROBIT_SYSTEM_COMPONENTS; i++) {
+        if (uBit.systemTickComponents[i] != NULL) {
+            uBit.systemTickComponents[i]->systemTick();
+        }
+    }
 
     // Update the display.
     microbit_display_tick();


dpgeorge:
Doesn't this then call the DAL's display update method?  Isn't that exactly what we don't want to do?

markshannon:
It does, but it is benign as we have already disabled it by this point:
https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/microbitdisplay.cpp#L468

dpgeorge:
Ok.  I guess then we could call MicroBit::systemTick() to reduce code duplication.  But in that case we do a bit more unnecessary processing in the idle components loop.

markshannon:
What I really want to avoid is the DAL's fiber/scheduler getting involved. Also, if we use the DAL's ticker then there is no guarantee that the accelerometer or compass will get updated as the DAL only updates them in the idle thread.

dpgeorge:
> Also, if we use the DAL's ticker then there is no guarantee that the accelerometer or compass will get updated as the DAL only updates them in the idle thread.

Yes, we would still need to explicitly update the compass and accel as we currently do.

Well I guess this is good for merging as-is then!

dpgeorge:
I verified that the touch functionality now works.

ntoll:
@dpgeorge it doesn't work for me. Believe me, I've tried (both with and without holding GND too). :-(

Viz the following test typed into the REPL:

```
while True:              
    if pin0.is_touched():
        print('touched') 
    else:                
        print('')        
    sleep(20)
```

In this instance `pin0.is_touched()` always returns False. Am I missing something obvious here..? This is with a hex built from the current master.

whaleygeek:
>
> 20ms is pretty fast, won't you flood the serial link with that and get
> screenfulls of spaces and newlines filling up the transmit buffer? Try a
> longer delay (250ms for example).
>


D

ntoll:
@whaleygeek you beat me to it. I've just managed to get it to work since I found that comment. Re-closing an apologies to everyone for the noise on the line. ;-)

tgauchet:
Hello I have the same issue with latest firmware.. Even this code is returning 0 (https://makecode.microbit.org/courses/ucp-science/electricity/setup-procedure). 
pin0 seems out of order.

dpgeorge:
> I have the same issue with latest firmware.. Even this code is returning 0

@tgauchet that link is for Make Code, but this repository here is about MicroPython on the micro:bit.  If you have a problem with MicroPython code then please copy that code here (if possible) and describe the issue you have.

tgauchet:
Thank you @dpgeorge , you are right it is working with MicroPython... My problem seems to be a Make Code issue.

Issue #104, opened 2015-11-22
Rationalise the display API.


markshannon:
Currently, to show an image, series of images or a string we have the following:
```python
display.show(string, delay=400) # prints the string to the display one character at a time
display.show(image, delay=400) # shows the image on the screen.
display.scroll(string, delay=400) # scrolls a string across the display (more exciting than display.print)
display.animate(image, delay, stride, start, wait=True, loop=False)
display.animate(iterable, delay, wait=True, loop=False)
```
There is some overlap between these and it is not always obvious which to use. The current interface is also inconsistent. `display.show()` doesn't take keyword arguments, but `display.animate` does, even though both take `delay`. `display.show("X", 500)` shows "X" forever. `display.show("XX", 500)` shows "X" for a second.

I propose to the following API:
Keep only `display.show()` and `display.scroll()`.
```python
display.show(image, delay=0, wait=True, loop=False, clear=False)
display.show(iterable, delay=400, wait=True, loop=False, clear=False)
display.scroll(string, delay=400)
```
The new version of `display.show()` should operate as `animate` does now and would take an iterable of images and/or single characters and display them sequentially. The single image variant would differ only in the default value for `delay`. Since strings are already sequences of single characters, we don't need to treat them specially.

`display.show` will take all the optional arguments that `animate` does now. The compulsory `delay` argument of `animate` would become optional and a sensible default provided. `display.show` will also gain a new optional parameter `clear` which would default to `False` and determine whether the display should be cleared at the end of the iteration.

`display.scroll()` would continue to work as it does now, creating a special iterator that is passed to `display.show()`: `display.scroll = lambda txt: display.show(scroller(txt))`.

This API change would entail some loss of functionality as `display.animate(image, delay, stride, start, ...)` will be gone. I don't see this as a problem as it can be easily reimplemented using `display.show` and `Image.crop`. In fact, it might make a nice example to do just that.

sparkslabs:
In case useful, +1 on the change :-)

(incidentally I've always thought of print as an internal thing, and
scrolling IMO is easier to read.)

Also, my understanding is this: since display.show when used with  a string
already displays a sequence of images, and you're building on that ? ie
display.show will show whatever is passed in. If it's passed in a string,
it shows one character at a time, and if what's passed in is an iterable
that returns images, it will show the images.

That strikes me as quite neat and tidy - since it allows something
logically like this:

    display.show("Hello")  # Not sure you're allowing this :-)

And

[...]

PR #154, opened 2016-01-16
Implement the new (and final) display API.

markshannon:
In summary the new API is 
```python
display.show(image, delay=0, wait=True, loop=False, clear=False)
display.show(iterable, delay=400, wait=True, loop=False, clear=False)
display.scroll(string, delay=400)
```
See for https://github.com/bbcmicrobit/micropython/issues/104 for full discussion about the new API.


dpgeorge:
It looks like display.animate still exists, is that correct?  If it were removed could one also remove the microbit_display_animate helper function, or is that needed by other functions?

ntoll:
Just to help me plan my work over the coming days... is there an ETA when this will land? I'll need to find time to revise the documentation and downstream editors that rely on there being a recent .hex file to flash the device. The release of the actual device is (at a guess) likely to be a matter of a few weeks at the most.

markshannon:
Yes, `display.animate` still exists. 
It could be removed, I don't care much either way. 
What do you think? 
@ntoll?

It won't allow us to remove the `microbit_display_animate` function as that sets up animations in general, but we will able to remove the image slice iterator as this is the only use of it.

markshannon:
I'll have time to fix up any comments tomorrow evening. I'd guess it should be in by the weekend?

ntoll:
@markshannon for `display.animate` we should follow the Zen of Python so that there is one obvious way to do a thing. Ergo please remove the `animate` function. Also, perfect timing so I can do a documentation blitz over the weekend. :-)

dpgeorge:
> we should follow the Zen of Python so that there is one obvious way to do a thing. Ergo please remove the animate function

+1.  Also remove the image slice stuff.

markshannon:
Done. And I've updated the docs as well.

ntoll:
@markshannon care to squash/rebase..? I'll test it works but leave it to @dpgeorge to pass judgement on the C (and hopefully merge)... ;-)

markshannon:
Rebased.

ntoll:
OK... this was also a lot of fun to test and it "feels" nice and simple. Really great work @markshannon, top notch stuff. :-) @dpgeorge waiting on you to check the C and merge. :-D

dpgeorge:
I tested it and it's really nice: simple yet powerful, and nice and consistent.  I can confirm that it fixes #157.

I'll make a few comments in-line.  Then if @markshannon can please rebase I think it's good to merge!

source/lib/utils/iters.c
```diff
@@ -0,0 +1,74 @@
...
+} repeat_iterator_t;
+
+STATIC mp_obj_t microbit_repeat_iter_next(mp_obj_t iter_in) {
+    repeat_iterator_t *iter = (repeat_iterator_t *)iter_in;
+    mp_obj_t result = mp_iternext(iter->iterator);
+    if (result == MP_OBJ_STOP_ITERATION) {
+        iter->iterator = mp_getiter(iter->iterable);
+        result = mp_iternext(iter->iterator);
+        if (result == MP_OBJ_STOP_ITERATION) {
+            nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError, "cannot repeat empty sequence."));
```
dpgeorge: No need for fullstop at end of message.

source/microbit/microbitimage.cpp
```diff
@@ -440,16 +440,6 @@ mp_obj_t microbit_image_invert(mp_obj_t self_in) {
 }
 MP_DEFINE_CONST_FUN_OBJ_1(microbit_image_invert_obj, microbit_image_invert);
 
-mp_obj_t microbit_image_slice_func(mp_uint_t n_args, const mp_obj_t *args) {
```
dpgeorge: There is still a lot of other image slice functionality that can be deleted (eg microbit_image_slice function).

dpgeorge:
The inc/lib/utils and source/lib/utils directory is there to mirror upstream uPy, and has source code from there.  I think we should keep our custom, generic libraries just in inc/lib and source/lib.

markshannon:
All done, and rebased.

ntoll:
Merged :tada:

PR #155, opened 2016-01-16
Update module.json to point to microbit-dal version 1.4.6.

dpgeorge:
This updates the dependencies to use the latest version of the DAL.

v1.4.6 brings:
- gestures
- fix to pwm
- fix to touch pins deinitialising
- servo API on pins
- compass auto-calibration

Some of these fix bug for us, some we need to expose at the Python level.


ntoll:
Merged.

PR #160, opened 2016-01-19
Add gesture methods to MicroBitAccelerometer class.

dpgeorge:
This PR adds basic gesture methods to the accelerometer object.  They are:
```python
accelerometer.is_gesture('gesture name') # is this gesture currently active?
accelerometer.was_gesture('gesture name') # get and reset if this gesture was active since last call
accelerometer.get_gestures('gesture name') # get number of times that this gesture occurred
accelerometer.reset_gestures('gesture name') # reset the number of times that this gesture occurred
```
See #9 for background.

Things I don't like:

1. The name get_gestures is not intuitive at all.  It seems like it should return a list of gestures or something.

2. I don't think the get_gestures/reset_gestures functionality is even needed.  Does one need to know how many times the device was tilted left?

3. get_gestures can't tell you the order in which pending gestures occurred, just how many of a certain kind.

4. There's no easy way to tell if any gesture occurred, only if a specific one did.

One way to improve/fix these things is to remove get_gestures and reset_gestures as they are now, and repurpose the get_gestures method to return a list of all pending gestures.  This could have an upper limit (eg 10) so we don't need heaps of RAM to store them.  Eg:
```python
>>> print(accelerometer.get_gestures())
['left', 'face up', 'shake', 'face up']
>>> print(accelerometer.get_gestures())
[]
```
So calling get_gestures() would reset the pending status of *all* gestures.  Thus was_gesture('xx') would return False after calling get_gestures (until a gesture actually occurred again).

The other problem is that for the gestures to be recognised, the underlying DAL update function must be called regularly.  Usually it's called by the DAL idle handler, but we never call that (because uPy is never idle).  So in the PR it's called only when an accelerometer method is called (this is how it's been previous to this PR, nothing has changed here).  That means that if you actually want to detect gestures you need to sit in a loop calling one of the accelerometer methods, eg:
```python
while True:
    if accelerometer.was_gesture('shake'):
        print('shake!')
    sleep(10)
```
There's currently no way to do "other things" and for a gesture to be detected and recorded in the background.


dpgeorge:
BTW, this PR relies on #155.

ntoll:
I'll have a play this evening.

ntoll:
It works, but not before I got caught out by the the get/reset gestures because I've had a long day and wasn't thinking. Basically, I was expecting @dpgeorge's revised suggestion rather than the currently implemented version. I prefer the proposed version as described above.

Regarding putting things in an explicit loop - from an educational point of view this is a "good thing" [tm] IMHO. ;-)

Happy for this to be merged once revised so I can document it properly.

dpgeorge:
Ok, I'll change this PR to use the revised behaviour.

dpgeorge:
Updated.  The new API is:
```python
# return a string for the current active gesture (eg "face up")
accelerometer.cur_gesture()

# is the named gesture currently active? (the same as doing cur_gesture() == 'gesture name')
# (maybe we don't need this?)
accelerometer.is_gesture('gesture name')

# get and reset if the named gesture was active since last call
accelerometer.was_gesture('gesture name')

# get a tuple of gesture history, each element is a string naming the gesture
# first element is tuple is oldest in history, last element is most recent
# does not reset the history, so repeated calls returns the same tuple or bigger
accelerometer.get_gestures('gesture name')

# clear the gesture history
# after calling this, get_gestures will return the emtpy tuple until another gesture occurs
accelerometer.reset_gestures('gesture name')
```

It works quite well now, except that you need to remember to call an accelerometer function (even get_x() will work) to update the gestures.

@finneyj I think we need to reduce the values for these constants: MICROBIT_ACCELEROMETER_GESTURE_DAMPING and MICROBIT_ACCELEROMETER_SHAKE_DAMPING.  Currently they are 10 and that means you need 10 idle calls to accel::update for a new gesture to be recognised.  I think it could easily be cut down to 4, to get better response.  Can we make these constants configurable via MicroBitCustomConfig.h?

ntoll:
Looks great... one request, can we use `current_gesture` rather than a (potentially confusing) shortening..? Perhaps we could just shorten it to `gesture`..? Thoughts..?

dpgeorge:
> can we use current_gesture rather than a (potentially confusing) shortening..?

Sure.  I think current_gesture makes it obvious what it does.

ntoll:
Cool. If you update and rebase I'll check and merge. :-) (Infinitely more fun than trying to get CSS to work on IE, Safari, Firefox and Chrome on TD).

dpgeorge:
@ntoll done!

ntoll:
This is a *lot* of fun to test. 

https://www.youtube.com/watch?v=rv9_wucCpgU

:-)

Issue #98, opened 2015-11-18
music.pitch needs to be non-blocking


ntoll:
See here:

https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/microbitmusic.cpp#L344

ntoll:
Fixed by #161

PR #161, opened 2016-01-20
Allow music.pitch function to run in the background.

dpgeorge:
This patch implements non-blocking mode for music.pitch.  Fixes issue #98.

Also, the len parameter is renamed to duration.


ntoll:
Not sure what happened with git during rebase, but this work has been merged via 6a2e05a433b710c58a5f646459cedaa305792ac4

PR #167, opened 2016-01-22
Add random module

dpgeorge:
Includes:
```
getrandbits()
seed()
randint()
randrange()
choice()
random()
uniform()
```
Works, except that we need some changes to the DAL to get the seeding working properly: need to expose uBit.seedRandom() and uBit.randomValue as public (or something to this effect).  I'll post upstream about it.


dpgeorge:
Note that microbit.random() has been removed.

ntoll:
I guess this should only be merged when upstream DAL is fixed?

dpgeorge:
> I guess this should only be merged when upstream DAL is fixed?

I submitted https://github.com/lancaster-university/microbit-dal/issues/79 and https://github.com/lancaster-university/microbit-dal/issues/80.  But these might not be acted upon for some time...

ntoll:
Understood. 

@finneyj just so you know, this is a blocker for us releasing MicroPython for the micro:bit 1.0. As far as I can tell, @dpgeorge is happy to do the leg work if you're willing to swiftly merge his PR. 

Thoughts..?

finneyj:
@dpgeorge @ntoll - noted. We'll merge and tag a commit as soon as it's ready.

ntoll:
Given the upstream bug is fixed, what needs to be done to get the PR into a merge-able state?

dpgeorge:
> Given the upstream bug is fixed, what needs to be done to get the PR into a merge-able state?

We need the DAL tagged with a new version.  Other option is to reference "master" but that would only be a temporary solution.

jamesadevine:
@dpgeorge Will do!

jamesadevine:
lancaster-university/microbit-dal#1.4.9

ntoll:
Awesomeness on stilts! :-)

I'll check and merge when I get home from work.

dpgeorge:
Ok, this should now work.  There's an extra commit to update the DAL to v1.4.9.  Please check that it builds correctly (may need to delete yotta_* and refresh).

ntoll:
@dpgeorge I compiled the module as per the instructions from @finneyj to solve the BLE disabled problem.

I'm seeing errors. See the following REPL transcripts:

It appears `getrandbits` is broken for values of >= 32:

```
>>> import random
>>> random.getrandbits(32)
ValueError: 
```

I was expecting an integer. Contrast the above with Python 3:

```
>>> import random
>>> random.getrandbits(32)
3280387012
```

Also, `seed()` doesn't appear to do anything. I believe this is quite an important omission since I imagine being able to set a seed will be important for funky stuff like procedural generation. ;-)

Notice how I set the seed, then reset the board and try again "just in case" (of what, I know not, but it doesn't work). :-)

```
>>> import random
>>> random.seed(1)
>>> random.random()
0.8989272
>>> random.random()
0.7826762
>>> random.seed(1)
>>> random.random()
0.3477952
>>> random.random()
0.2056807
>>> MicroPython v1.5.2-76-g6a2af82 on 2016-01-21; micro:bit with nRF51822
>>> import random
>>> random.seed(1)
>>> random.random()
0.5387158
>>>
```

I was expecting reproducibly deterministic randomness. Contrast the above with Python 3:

```
>>> random.seed(1)
>>> random.random()
0.13436424411240122
>>> random.random()
0.8474337369372327
>>> random.seed(1)
>>> random.random()
0.13436424411240122
>>> random.random()
0.8474337369372327
```

...etc...

Ahh... I just noticed this: 

https://github.com/bbcmicrobit/micropython/pull/167/files#diff-c99d7100296fc412d0320ee27658acfeR43

and this:

https://github.com/bbcmicrobit/micropython/pull/167/files#diff-c99d7100296fc412d0320ee27658acfeR54

Can you give a simple explanation of what's going on in the former (so I can document this obviously expected behaviour) and revise the latter so it actually works..?

Cheers..! :-)

ntoll:
Forgot to mention all the rest works as expected! As always, great stuff! :-) :clap:

ntoll:
@dpgeorge OK... we're passing like ships in the night... I'll pull again and re-test now that I see things have been modified. :-D

ntoll:
@dpgeorge great stuff. It all works as expected. Care to explain the arbitrary cut off for `getrandbits`..?

dpgeorge:
> It all works as expected.

Great!

> Care to explain the arbitrary cut off for getrandbits..?

Because the underlying generator function returns at most 30 bits.  Yes, we could call it twice and or the missing, upper 2 bits with extra random data.  We could even support arbitrarily large number of bits by using bignum, eg:
```c
mp_obj_t ret = mp_obj_new_int(0);
for (as long as we need more bits) {
    ret = mp_binary_op(MP_BINARY_OP_LSHIFT, ret, 30);
    ret = mp_binary_op(MP_BINARY_OP_OR, ret, mp_obj_new_int(uBit.random()));
}
```
Is that worth it?

ntoll:
It's not worth it... we just need to make sure we document it very clearly. ;-)

ntoll:
Documented in d127c121770bc26cc3c441c7ac2cdd5116f66498

PR #168, opened 2016-01-24
Initial pass at doc updates. Still things to be written.

ntoll:
**DO NOT MERGE**

This is a work in progress. Perhaps @willingc or @deshipu may want to take a look too. I hope @dpgeorge likes the comic (created here: http://pycomic.github.io/)


docs/accelerometer.rst
```diff
@@ -3,7 +3,10 @@ Accelerometer
 
 .. py:module:: microbit.accelerometer
 
-This module gives you access to the on-board accelerometer.
+This module gives you access to the on-board accelerometer. As a convenience
```
deshipu: I wonder if we should be lying about this being a module. I initially wrote it like that, because I thought they were modules, but they really are not.

docs/accelerometer.rst
```diff
@@ -3,7 +3,10 @@ Accelerometer
...
+the accelerometer provides convenience functions for detecting gestures. The
```
ntoll: too much convenience... ;-)

docs/accelerometer.rst
ntoll: What would you describe it as? "This object gives you access..." etc..?

docs/button.rst
```diff
@@ -36,6 +36,19 @@ Classes
...
+        Returns ``True`` or ``False`` to indicate if the button was pressed
+        since the device started or the last time this method was called.
+
+    .. py:method:: get_presses()
+
+        Returns the running total of button presses.
+
+    .. py:method:: reset_presses()
+
+        Resets the running total of button presses to zero.
```
deshipu: Does this also reset ``was_pressed``?

ntoll: I believe it does (will confirm). This needs mentioning... well spotted!

docs/accelerometer.rst
deshipu: Not sure. I guess "object" is the closest to truth.

ntoll: I agree. We say somewhere else that such things are objects representing aspects of the device's hardware. It'd be confusing if we were to differ here.

ntoll:
Merged.

PR #169, opened 2016-01-26
Allow microbit.sleep function to take a floating point number.

dpgeorge:
Addresses issue #7.


ntoll:
Tested over a cheeky lunch-break code review. ;-)

Nice one.

Issue #11, opened 2015-09-17
Add temperature function?


dpgeorge:
The MCU has a temperature sensor and so does the compass.  We can easily expose either or both of these as a function like `microbit.temperature()`.  They need calibrating (they have good relative accuracy but bad absolute values) so we might also need `microbit.calibrate_temperature(degrees)`.

whaleygeek:
Note that this has just about surfaced in the TouchDevelop language now (I saw evidence of it in the docs online the other day). @finneyj does this mean it's in the runtime now?

remay:
Yes it's in the runtime

dpgeorge:
Temperature is in the DAL.  It's exposed as uBit.themometer.getTemperature().  It returns an integer for the degrees celsius.

But actually there is precision of 0.25 degrees behind the scenes.  We could make a function microbit.temperature() which returns a float.

ntoll:
+1

This would be especially timely since I've got a micro:bit running MicroPython with a friend who's spending some weeks at the British Antarctic Survey. If she updates to a version with a temperature she can do some interesting measurements. ;-)

whaleygeek:
>
> The accuracy of this temperature measurement in practice is very poor. It
> is the core temperature of one of the CPU's, and this is affected by
> thermal transfer characteristics of the silicon, and running more CPU
> intensive operations raises the core temperature.
>

According to the NRF datasheet Tacc (accuracy of the temperature sensor) is
+/- 4C.

Is reporting temperature to 0.25C increments on a sensor that measures core
temperature with an accuracy of +/-4C of any use?

dpgeorge:
> Is reporting temperature to 0.25C increments on a sensor that measures core temperature with an accuracy of +/-4C of any use?

No!  Ok, so let' just expose the DAL's temperature function as microbit.temperature().

ntoll:
+1 to `microbit.temperature()` if that's nice and easy (which is what I think you're suggesting).

ntoll:
Fixed by #170

PR #170, opened 2016-01-26
Add microbit.temperature() function, to return MCU core temperature.

dpgeorge:
Addresses issue #11.


ntoll:
Yay... tested above a radiator and in the fridge at home. The values go up and down as expected. ;-)

PR #171, opened 2016-01-26
Add SPI class.

dpgeorge:
Basic SPI class/singleton-instance with methods:
```python
spi.init(baudrate=1000000, bits=8, mode=0, *, sclk=pin13, mosi=pin15, miso=pin14)
buf = spi.read(nbytes, byte_out=0x00)
spi.write(buf)
spi.write_readinto(wr_buf, rd_buf) # wr_buf and rd_buf can be the same, or different
```
I haven't tested it properly with an actual SPI peripheral.  I originally implemented it to get neopixels working with SPI, but it couldn't be done due to timing.


ntoll:
Suggestions for an SPI peripheral to test it with most welcome! :-) I bet @deshipu has something.

deshipu:
Hmm, I have this radio module that I wanted to test, but not much documentation about it. Unfortunately, I still didn't figure out a way to connect to the small pins reliably -- I gave up the edge connector with Monifa.

ntoll:
Edge connector available here: https://www.kitronik.co.uk/catalogsearch/result/?q=microbit

If you can think of something SPI I can get to go with it then I can test it. Suggestions..?

deshipu:
I actually ordered something similar already, but it's going to take a while.

ntoll:
@deshipu I should get a micro:bit edge connector / breakout board thingamabob delivered on Thursday (just ordered it). I'll need something to plug it into though to test SPI.

deshipu:
Do you have an arduino board?

deshipu:
Let's move that to IRC.

whaleygeek:
You could try with one of these, and use my python code as a reference to
the commands to send to it...

https://energenie4u.co.uk/index.phpcatalogue/product/ENER314

https://github.com/whaleygeek/pyenergenie

You can buy bundles with mains sockets that you can control to turn your
lights on and off with...

http://cpc.farnell.com/energenie/miho002/energenie-mihome-adaptors-pack/dp/PL14993

dpgeorge:
Well, I do have a lot of SPI hardware (LCDs, SD card, SPI flash chips) so could get to testing it sometime.

dpgeorge:
Rebased.

dpgeorge:
I've updated and rebased this PR.  I've tested it.

To test:

1. Use this script: https://gist.github.com/dpgeorge/4b3fcac61cea3c70328a which implements an SPI slave on the pyboard (you'll need a pyboard to run it).  Download it to the pyboard and run it (eg copy it to main.py on the flash or an SD card).

2. Connect the pyboard's X-channel SPI (X6=sclk, X7=miso, X8=mosi) to 3 pins on the microbit (eg pin0, pin1, pin2, but it can be any pins).

3. Connect to the microbit via REPL.

4. Init the spi: `spi.init(125000, sclk=pin0, miso=pin1, mosi=pin2)` (be sure to use the pins relevent for your set-up).

5. Test read: `spi.read(2)`, should return `b'X\xd0'`.

6. Test write: `spi.write(b'12')`, should toggle the green LED on the pyboard.

7. Test write_readinto: `buf = bytearray(2); spi.write_readinto(b'34', buf); print(buf)`, should print `b'X6'` and blue LED should toggle.

Of course, you can play around with it and send other commands.  If the SPI goes out of sync, try writing single bytes (spi.write(b'1')) or reset the pyboard.

dpgeorge:
Note that there is no documentation...

ntoll:
@dpgeorge *THAT* was an interesting 30mins.

I got it set up but had the following errors when restarting the pyboard via the REPL

```
PYB: sync filesystems
PYB: soft reboot
node 0 cannot be compiled
  File "main.py", line 34
SyntaxError: internal compiler error
ViperTypeError: can't do binary op between types 0 and 2
ViperTypeError: type mismatch, local in_val has type 2 but source object has type 0
ViperTypeError: type mismatch, local in_val has type 2 but source object has type 1750
ERROR: stack size not back to zero; got -3
Micro Python v1.3.3 on 2014-10-02; PYBv1.0 with STM32F405RG
>>>
```

After a quick update of the firmware on the pyboard it booted correctly and I managed to get the following session in the REPL of the micro:bit.

```
>>> spi.init(125000, sclk=pin0, miso=pin1, mosi=pin2)
>>> spi.read(2)
b'\xb1\xa0'
>>> spi.write(b'12')
>>> buf = bytearray(2); spi.write_readinto(b'34', buf); print(buf)
bytearray(b'X6')
```

**NOTE THE DIFFERENCE:**  the result of the `spi.read(2)` command is not what was described. I have no idea if this is significant. All the other commands work as expected.

I realise there's no documentation. It might be a good exercise for me to write it tomorrow and you can check it. Then, assuming we're ok with the test results and docs we should merge.

Almost there... IIRC we just need scroll to react to ctrl-C and perhaps include the youthful exuberance of `import antigravity`. Would be good to get light sensors and persistent storage in there too for, perhaps, 1.1.

ntoll:
Actually, I just tried the `spi.read(2)` command again...

```
>>> spi.read(2)
b'X\xd0'
```

Each time it returns the expected result. Not sure why the first call resulted in the anomalous result. Thoughts..?

whaleygeek:
With the SPI, is it possible to set the mode (SPOL and SPHA) in the python
API?

It's easy, for example to drive the flash programming interface of an Atmel
via SPI (I wrote a pure python implementation that flashes code into a
AT328P for example), but there are often SPOL and SPHA requirements that
vary between different slave ends.

dpgeorge:
> Not sure why the first call resulted in the anomalous result. Thoughts..?

The first call is the correct result but left shifted by 1 bit, so the slave lost the first bit.  Don't know exactly why, but could be the pyboard's SPI slave code.

> With the SPI, is it possible to set the mode (SPOL and SPHA) in the python API?

@whaleygeek Yes: `spi.init(baudrate=1000000, mode=1)`, where the mode parameter is as per the Wikipedia section on CPHA/CPOL and can be 0,1,2,3.  This'll need to go in the docs.

dpgeorge:
Rebased.  Maybe we can merge this and then write docs as a separate PR?  Otherwise I can get to writing them later today maybe.

ntoll:
I'm mid-documenting them as we speak (I'm going out later with family so want to get this out so you have something to "correct"). ;-) Once your happy with the docs I can merge when I get back this afternoon.

ntoll:
@dpgeorge ugh... my git-fu is failing me. Documentation commit is here: 34d1694546e267549b757307f60e644e0ec7fc2c Can you get it into this PR (I thought my push would append it as a second commit, but it appears not despite the fact that GitHub references this PR from the branch I pushed it to). So sorry for the faffing around. :-(

dpgeorge:
@ntoll this has been rebased and should be good to merge now.

ntoll:
Yay... PR for docs incoming.

PR #176, opened 2016-01-29
Music and Neopixel documentation update

gbaman:
Adds new Neopixel module documentation (including an example) and updates the music module documentation.

### DIFF
--- a/docs/music.rst
+++ b/docs/music.rst
@@ -4,8 +4,10 @@ Music
 .. py:module:: music
 
 This is the ``music`` module. You can use it to play simple tunes, provided
-that you connect a speaker to your board.
+that you connect a speaker to your board.      
+Note that the ``music`` module is no longer included in the ``microbit`` module. To use it, it must be imported separately with::
 
+    import music
 
 Musical Notation
 ================
@@ -116,7 +118,7 @@ Built in Melodies
 For the purposes of education and entertainment, the module contains several
 example tunes that are expressed as Python lists. They can be used like this::
 
-    >>> from microbit import music
+    >>> import music
     >>> music.play(music.NYAN)
 
 All the tunes are either out of copyright, composed by Nicholas H.Tollervey and
@@ -146,3 +148,8 @@ They are:
     * ``JUMP_DOWN`` - for use in a game, indicating downward movement.
     * ``POWER_UP`` - a fanfare to indicate an achievement unlocked.
     * ``POWER_DOWN`` - a sad fanfare to indicate an achievement lost.
+
+Example
+--------   
+.. include:: ../examples/music.py
+    :code: python
--- a/docs/neopixel.rst
+++ b/docs/neopixel.rst
@@ -1,0 +1,69 @@
+Neopixel
+****
+
+.. py:module:: neopixel
+
+The ``neopixel`` module lets you use Neopixel (WS2812) individually addressable RGB LED strips with the Microbit.    
+Note to use the ``neopixel`` module, you need to import it separately with::
+
+    import neopixel
+
+.. warning::
+    
+    Do not use the 3v connector on the Microbit to power any more than 8 Neopixels at a time.
+    If you wish to use more than 8 Neopixels, you must use a separate 3v-5v power supply for the Neopixel power pin.
+    
+.. note::
+    
+    From our tests, the Microbit Neopixel module can drive up to around 256 Neopixels. 
+    Anything above that and you may experience weird bugs and issues.   
+
+Functions
+=========
+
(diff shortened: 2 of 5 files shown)


docs/music.rst
```diff
@@ -4,8 +4,10 @@ Music
 .. py:module:: music
 
 This is the ``music`` module. You can use it to play simple tunes, provided
-that you connect a speaker to your board.
+that you connect a speaker to your board.      
+Note that the ``music`` module is no longer included in the ``microbit`` module. To use it, it must be imported separately with::
```
dpgeorge: I don't think we need to refer to legacy behaviour.  So simply say "Note thate the music module is not included in the ..."

ntoll:
This duplicates #168 - I'll merge this work into #168.

docs/neopixel.rst
```diff
@@ -0,0 +1,69 @@
...
+.. py:module:: neopixel
+
+The ``neopixel`` module lets you use Neopixel (WS2812) individually addressable RGB LED strips with the Microbit.    
+Note to use the ``neopixel`` module, you need to import it separately with::
+
+    import neopixel
+
+.. warning::
+    
+    Do not use the 3v connector on the Microbit to power any more than 8 Neopixels at a time.
```
ntoll: I've successfully used the 3v connector on the micro:bit to drive 30 pixels. Out of interest, what's the worst that could happen..?

ntoll:
This work has been merged into the #168 PR. See 6bab76984d780649f37fc9783c9c98bd07658c6d.

docs/neopixel.rst
gbaman: Well, for Neopixels the max figure if you put them full white is 60ma/pixel. So for 8 pixels, that is 480ma. For 30 pixels that is 1.8 amps...
Of course, the more normal figure used is in general will draw 20ma/pixel, but even that at 30 pixels is 600ma!
The 3v3 pin is going through the regular, am assuming 600ma wouldn't end well...

We have to keep in mind with kids, chances are they will at some point turn them all on full white though by mistake, in turn blowing their microbit if using many.

remay: Current recommendation is max of 100mA from the edge connector
image (source: micro:bit student guide)

gbaman: Whoops! Then even 8 pixels breaks that one... I guess they are unlikely to have them all on bright at the same time. What could possibly go wrong..

remay: Knowing the BBC I suspect that this figure will be exceedingly conservative, but I have no knowledge of what this pad is driven from ...

whaleygeek: It depends how you power the micro:bit. If you power it from the USB, then
you are ultimately limited by the 'negotiated current at enumeration time'.
See below:

Davids-Air-2:~ davidw$ system_profiler SPUSBDataType > after

Davids-Air-2:~ davidw$ system_profiler SPUSBDataType > before

Davids-Air-2:~ davidw$ diff before after

20a21,38

>         MBED CMSIS-DAP:

>

>           Product ID: 0x0204

>           Vendor ID: 0x0d28

>           Version: 10.00

>           Serial Number: 9900023431864e45001600100000006300000000cc4d28bd

>           Speed: Up to 12 Mb/sec

>           Manufacturer: MBED

>           Location ID: 0x14100000 / 5

>           *Current Available (mA): 500*

*>           Current Required (mA): 500*

>           Capacity: 8.5 MB (8,454,144 bytes)

>           Removable Media: Yes

>           Detachable Drive: Yes

>           BSD Name: disk2

>           Partition Map Type: Unknown

>           S.M.A.R.T. status: Not Supported


You are then further limited by whatever the on-board power path is through
the freescale (as it regulates 5V down to 3V) - you could check the data
sheet for that part, as it's listed on some of the blogs what part number
this device is.


The story *might* be different when running from a battery - I doubt there
is even a regulator at that point, because otherwise you'd be regulating 3V
to, erm, 3V, which is pointless unless you ever expect a higher input
voltage. Legend has it though that a certain person ran it from 6V and got
away with it, which might suggest regulation is present, although I'm sure
that's out of spec. Check the freescale datasheet, you'll probably get the
best answer from that.


NeoPixels are designed to run from 5V, they *usually* work from 3V. So the
current characteristics might be different when below the rated voltage.


However, I would say that if the BBC say the maximum current from that pin
should be 100mA, then it is advisable to design circuits with that in mind,
and recommend an external power supply (another battery pack) for more than
100mA of NeoPixels.

dpgeorge: I've used 8 NeoPixels at full brightness at the same time without a problem, but that's powered via USB.

PR #177, opened 2016-01-29
Added the XKCD antigravity comic in ASCII form.

JoeGlancy:
I re-created the XKCD antigravity comic (http://xkcd.com/353/) in ASCII form, and added it to the antigravity module to simulate the classic Python antigravity.


gbaman:
Have tested it and all works as expected.

screen shot 2016-01-29 at 19 40 16

dpgeorge:
That is very nice.  But I'd say it adds 30*80=2400 bytes to the code that could be much better spent on other things...

ntoll:
We already have an `antigravity` module. Put an image on the display then `import antigravity` and watch in wonder as the pixels in the image float to the top. ;-)

ntoll:
Ahh... I see what you mean... you're adding it to the existing module. As @dpgeorge mentions, we have a severely memory constrained device. :-(

markshannon:
Not to mention copyright.

JoeGlancy:
Okey dokes, 2.4kb is a bit much. I'll see how far I can get it down to, but it's not going to be pretty.

ntoll:
@JoeGlancy so... one way to do it would be to just use the first frame and take some spacing liberties. For example:

```
"+- xkcd.com/353 --------------------------+\n"
+"|                                        |\n"
+"|                               \\0/      |\n"
+"|                              /  \\      |\n"
+"| You're flying!      MicroPython! /\\     |\n"
+"|     How?                               |\n"
+"|     /                                  |\n"
+"|  0                                     |\n"
+"| /|\\                                    |\n"
+"| / \\                                    |\n"
+"+----------------------------------------+\n"
```

I believe you're doing some sort of RLE compression too. The important thing to remember is that it's **Micro**Python on a **micro**:bit so it's OK to take **micro** liberties. ;-) For example, try ``import this`` (and contrast it with the result of doing so in regular Python).

JoeGlancy:
(too many commits, my apologies - I was forwarding the code to someone else :) I've modified it (to a rather unreadable extent) and managed to get it to 1.9-ish kb, code and everything else. If it is *absolutely* necessary, I can get it down a few more bytes, but there is still room in MicroPython at the current moment - of course this can be shortened to something like the above or even removed in the future if space is required.

ntoll:
@JoeGlancy how small can you get it with just that single frame?

JoeGlancy:
Give me a few minutes and I'll see. I've contacted XKCD about licensing, however from what I can tell from http://xkcd.com/license.html it seems OK as MicroPython is nonprofit?

ntoll:
@JoeGlancy it's an open source project, so isn't a business as such. Would be fun to get Randall's blessing. ;-)

JoeGlancy:
For sure :D

JoeGlancy:
How's that (no micro:bit or build environment so cannot test for size)?

dpgeorge:
It seems to be 536 bytes now.

JoeGlancy:
I assume that's acceptable then?

whaleygeek:
Just a thought...

There's a lot of whitespace in that - could you run length encode the
spaces, as there a lot of them, about 300 to be precise.

It would then give you more space for other little cartoons like this? The
longest run of spaces is 40, so you could encode a long run in two bytes,
and a tiny bit of C code would easily uncompress them. The dashes at top
and bottom would also run length encode (first byte with bit 7 set and
length in lower 7 bits, giving a max RLE of 127, second byte the character
to repeat). The C to uncompress this is pretty tiny, probably less than the
cartoon itself, and it would give you a way to encode other cartoons later
as other little easter eggs to discover.

"+- xkcd.com/353 --------------------------+\n"
+"|                                        |\n"
+"|                               \\0/      |\n"
+"|                              /  \\      |\n"
+"| You're flying!      MicroPython! /\\     |\n"
+"|     How?                               |\n"
+"|     /                                  |\n"
+"|  0                                     |\n"
+"| /|\\                                    |\n"
+"| / \\                                    |\n"
+"+----------------------------------------+\n"

dpgeorge:
@JoeGlancy would you be able to try @whaleygeek's idea of run-length-encoding?  For example:
```c
const char *data =
"|\x8f-|\n"
"|\x8f |\n"
"|   some text    |\n"
"|\x8f |\n"
"|\x8f-|\n";

void decode_and_print(const char *s) {
    while (*s) {
        if (*s & 0x80) {
            // print run of chars of length (*s & 0x80) + 1, char being s[1]
        } else {
            // print *s itself
        }
    }
}
```
(Of course that code needs embellishing!)

ntoll:
@JoeGlancy I believe if you embellish @dpgeorge's suggested implementation to implement RLE for the smaller single pane version of the comic, then we'll have quite a fun solution!

JoeGlancy:
@whaleygeek neat - I don't know what I was thinking having lots of function calls anyways. @dpgeorge I've stuck to using the first byte for storing the amount of times the character appears and the second for the character itself, but you would be able to get a maximum of 255 characters into 2 bytes if you use the first for the (7-bit) character and the second for the amount. @ntoll let me fix up this commit mess and I'll push

JoeGlancy:
This might need closing and a new pull request opened, as my current fork is a new one with a single commit (23 is excessive). Might have been a bad idea :\

dpgeorge:
> you would be able to get a maximum of 255 characters into 2 bytes if you use the first for the (7-bit) character and the second for the amount

@JoeGlancy yes I thought of that as well, but it would make it harder to write the RLE strings by hand because you'd need to know the ascii code for the character and then or the high bit.

PR #179, opened 2016-01-30
Allow ctrl-C to stop a display animation (either scroll or show).

### LINKED ISSUE #175: Should be able to ctrl-C to stop an ongoing display animation
Currently if you do:
```python
>>> display.scroll('microbit')
```
then you can't use ctrl-C to stop it.  This should be fixed!

dpgeorge:
This fixes #175.

The display retains the image at the point ctrl-C was pressed.  If wanted it's easy to make it clear the screen (but I think it's better to do less and not clear).


ntoll:
Great stuff. It works!

Issue #182, opened 2016-02-01
String.format() right justification not working for string arguments


realcundo:
When playing with Marja (using [`firmware.hex`](https://github.com/ntoll/microrepl/blob/master/firmware.hex)) I came across this:
```python
print("x{:>4}x".format(42))
print("x{:>4}x".format("42")
```
produced
```
x  42x
x42  x
```
Not a big deal but I was wondering whether this was intentional, consequence of squeezing as much Python into as little space as possible (justification is ignored?), or a genuine bug.

dpgeorge:
This is now fixed in the upstream code.  I'll see if I have time to sneak those fixes into microbit before v1.0....

ntoll:
Yes please :-)

dpgeorge:
Actually, this is already fixed.  You must have an older version of the firmware.

PR #183, opened 2016-02-01
Add SPI documentation.

ntoll:
The docs as previously seen with corrections.

I'll add an issue about checking *all* the docs for * kwargs - I think it probably easier to do it all in a single PR.

### DIFF
--- a/docs/spi.rst
+++ b/docs/spi.rst
@@ -1,0 +1,68 @@
+SPI
+***
+
+.. py:module:: microbit.spi
+
+The ``spi`` module lets you talk to a device connected to your board using
+a serial peripheral interface (SPI) bus. SPI uses a so-called master-slave
+architecture with a single master. You will need to specify the connections
+for three signals:
+
+* SCLK : Serial Clock (output from master).
+* MOSI : Master Output, Slave Input (output from master).
+* MISO : Master Input, Slave Output (output from slave).
+
+
+Functions
+=========
+
+.. method:: init(baudrate=1000000, bits=8, mode=0, sclk=pin13, mosi=pin15, miso=pin14)
+
+    Initialize SPI communication with the specified parameters on the
+    specified ``pins``. Note that for correct communication, the parameters
+    have to be the same on both communicating devices.
+
+    The ``baudrate`` defines the speed of communication.
+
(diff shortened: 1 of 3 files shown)


docs/spi.rst
```diff
@@ -0,0 +1,61 @@
...
+    | SPI Mode | Polarity | Phase |
+    +==========+==========+=======+
+    | 0        | 0        | 1     |
+    +----------+----------+-------+
+    | 1        | 0        | 0     |
+    +----------+----------+-------+
+    | 2        | 1        | 1     |
+    +----------+----------+-------+
+    | 3        | 1        | 0     |
+    +----------+----------+-------+
```
dpgeorge: The phase is actually opposite to what you have: mode=0 is CPOL=0, CPHA=0, etc.

ntoll: OK... this was taken from the Wikipedia article you referenced. Just to be certain, can you let me know precisely what the values should be. Happy to do the leg-work.

ntoll: Also, is it more correct to say ``CPOL`` and ``CPHA``..? I used "normal" English so numpties like me don't get blinded by TLAs. Having said that I think it important we perhaps use both in this case.

dpgeorge: It should be:
```
0 0 0
1 0 1
2 1 0
3 1 1
```
Ideally we'd actually describe what the numbers mean:  Polarity (aka CPOL) 0 means that the clock sits at logic level 0 when idle, and goes high when active; polarity 1 means the clock sits at logic level 1 when idle.  Phase (aka CPHA) 0 means that data is sampled on the leading edge of the clock, and 1 means on the trailing edge.

ntoll: OK... update incoming in within minutes.

ntoll:
Updated and squashed. Is it OK?

ntoll:
**HOLD OFF** RST is munging the table. :-(

ntoll:
OK... good to go... sorry. I made a minor change just after testing the rendering. That'll teach me.

dpgeorge:
Looks good!

dpgeorge:
Yay!

ntoll:
Yay!

ntoll:
What's left for 1.0..? I suspect just #182..?

PR #191, opened 2016-02-02
Clear display when animation is interrupted.

markshannon:
Clears the display when scrolling is interrupted.
Otherwise, pressing ctrl C can leave part of a character on the display.


dpgeorge:
This behaviour was intentional.

I like to do the minimum possible for a given feature, which in this case means leaving the display as-is.  Maybe they pressed ctrl-C because they wanted to stop the display to see one of the frames of animation?  Ctrl-C will only be pressed at the REPL (in usual cases) and likely they are going to then do something else with the display which will erase the chars.

PR #192, opened 2016-02-03
Update to equality statement

corinnewelsh:
Added sixth equality strand (disability) and re-phrased two of the others.

### DIFF
--- a/docs/devguide/contributing.rst
+++ b/docs/devguide/contributing.rst
@@ -3,9 +3,7 @@ Contributing
 
 Hey! Many thanks for wanting to improve MicroPython on the micro:bit.
 
-Contributions are welcome without prejudice from *anyone* irrespective of
-age, gender, religion, race or sexuality. Good quality code and engagement
-with respect, humour and intelligence wins every time.
+Contributions are welcome without prejudice from *anyone* irrespective of age, disability, gender, race, religion and belief, or sexual orientation. Good quality code and engagement with respect, humour and intelligence wins every time.
 
 * If you're from a background which isn't well-represented in most geeky groups, get involved - *we want to help you make a difference*.
 * If you're from a background which *is* well-represented in most geeky groups, get involved - *we want your help making a difference*.


dpgeorge:
I'm ok with this, but personally would prefer not to say anything about "religion and belief".

ntoll:
@corinnewelsh many thanks for this.

I *would* have a problem working with neo-Nazi white supremacists hoping to contribute code that somehow promotes their belief (however unlikely that could be). Belief is different from religion. You may say you're a Buddhist and that's fine and totally your personal business, but if you say you believe homosexuals should be neutered then that's most definitely not OK because you're expressing a belief about the world that is incompatible with the spirit of this document. Sorry for the push back, but does this make sense?

Also, "sexuality" is not synonymous with "sexual orientation" - the first is broader, the latter quite specific about what is being referenced, if you see what I mean. The former encompasses trans* whereas the latter is specific to sexual orientation (they're different things).

Finally, can we replace the negative sounding "disability" with the more positively framed "special need"..?

Finally (take 2 - sorry) please respect the page width!!!! :-P

How about...

```
Contributions are welcome without prejudice from *anyone* irrespective of
age, gender, religion, race, special need or sexuality. Good quality code and 
engagement with respect, humour and intelligence wins every time.
```

Thoughts and please feel free to push back. I believe it important we do this sort of thing properly in an open and collaborative way.

Charlotteis:
Hi there, @ntoll thought I might be interested in this PR, and I was!

```
Contributions are welcome without prejudice from *anyone* irrespective of
age, gender, religion, race, special need or sexuality. Good quality code and 
engagement with respect, humour and intelligence wins every time.
```

is a good start but there's a few more things that I would want added if I were to ever engage with the project.

Not sure if you've heard of the Contributor Convenant (http://contributor-covenant.org/) but it's becoming a standard in Open Source and covers a lot of good things:

Including the suggested characteristics you'd get something like:

```
Contributions are welcome without prejudice from *anyone* irrespective of age, body
size, disability, ethnicity, gender identity and expression, level of experience,
nationality, personal appearance, race, religion, or sexual identity and
orientation. Good quality code and engagement with respect, humour and intelligence wins every time.
```

Importantly it both covers 'gender identity and expression'; as 'gender' by itself only tends to think about 'woman/man' and not the whole range of wonderful gender identities in the world. And it differentiates between sexual identity and orientation too, which is important. Also adds in a few other good things like appearance, body size and even level of experience with the project :) I think Religion is a good thing to have. 'belief' is not something I've encountered before, and @ntoll makes some good points about that.

Have a great day! :sun_with_face:

ntoll:
Hahaha... So @Charlotteis is a buddy of mine who is very active in this space so I knew she'd have the depth of experience to suggest something awesome. Thanks @Charlotteis..! :-)

Given the wider and more inclusive net @Charlotteis statement casts and the fact that it's based upon an existing and respected known prior-work in this area, is everyone OK that we update this PR to reflect @Charlotteis's suggestion..?

dpgeorge:
I think "body size", "personal appearance" and "nationality" are unnecessary.  Any person of good character would not even think to discriminate against these.  If we have "gender identity and expression" then we don't also need "sexual identity and orientation".

tjguk:
I've gone back and forth on whether to say this or not but I'll say it
and then step away from the mic:

There is no need to list a set of attributes which might reflect the
prejudices of one person or another. Especially since, as per this
thread, there isn't going be total agreement on what should be in that list.

Either we accept contributions on their own merits or we don't.

If we do, then the simple statement:

  We accept contributions without prejudice from anyone

or, more simply still:

  We accept contributions from anyone

is as inclusive as it is possible to be. No-one can complain that their
particularity is excluded by that statement because everyone is included
by it.

ntoll:
@tjguk that's a very good point. The bottom line is that I'd love it if people didn't feel excluded and understood we're a good bunch of people to work with. If we can say that in the widest possible sense but in the simplest possible way then that's a definitely plus. Thoughts..?

dpgeorge:
Yes, I'd be +1 on @tjguk's suggestion.  We can keep the "without prejudice" to make it clear what we are getting at.

deshipu:
I agree that we should always accept patches no matter who submitted them, based only on the contents of the patch, even if they we personally don't particularly like the person in question or their beliefs. However, I understand why explicitly listing the common discrimination criteria is important for the people who match them -- it encourages those people and assures them that yes, they are welcome. You see, they usually are used to read "except you" in between the lines. Of course we can't list all of them, but those examples do help.

deshipu:
By the way, I would also add "experience" as one of those.

Charlotteis:
I am a non-binary individual. I have avoided countless projects who don't include 'gender identity and expression', particularly when I mention it to them they mention how 'gender' is 'enough'. It is not enough. 

Recently I have worked with both the Rust and Node.js communities to get these exact characteristics included in their Codes of Conducts. They had no problems with them. I really see no harm in being more verbose here. Many communities are adopting these characterstics, including Microsoft and Apple (with ChakraCore and Swift). I really do not see any harm in following their lead.

Also, if no good person would hate any person about their weight or nationality, why not just remove all the characteristics as no good person of character would be horrible? It doesn't make sense to me.

The point is, this is becoming the norm. And why shouldn't MicroPython take the same measures to do this? Particularly being an educational tool as part of microbit, in which young children will be forming their own identities and seeing themselves included and such a list can be a phenomenal feeling (speaking again from my experience as a child growing up and forming their own queer and non-binary identities).

> it encourages those people and assures them that yes, they are welcome. You see, they usually are used to read "except you" in between the lines. 

exactly that. Society has taught me over an over again to think "except me"

ntoll:
OK... I'm trying to keep it infinite, unambiguous, simple but also want to address the "except you"-ishness that @deshipu and @Charlotteis point to. How about?

"We accept contributions from anyone. If you're thinking, 'but not me', then we especially mean you."

???

deshipu:
Those things are really tricky, I would really avoid reinventing the wheel and go with the tried and tested way that was worked out together, with feedback from a large audience. I don't really have the qualifications to come up with novel solutions here.

dpgeorge:
I just don't like it that a community is assumed "not good" and "not inclusive" unless they explicitly state otherwise.

corinnewelsh:
I don't have an opinion on whether we include an equality statement or not.

"Disability" was the word accepted by the Disability Discrimination Act superseded by the Equality Act 2010.  Many disabled people oppose the term "special need" -- on the basis of having the same needs as anyone else (food, shelter, etc.)  This is along the social model of disability: disabled by society.

I think the inclusion of belief in the "religion and belief" is that it encompasses philosophical belief where people do not define that as religious (e.g. humanist). Both also reference "or none".  Again, it was with reference to the Equality Act.

deshipu:
@dpgeorge I don't like it either, but that's how it is, and that's how it is going to be read by people considering helping. I honestly hope the situation will improve gradually.

ntoll:
@corinnewelsh great stuff. Just shows how ignorant I am. Apologies for muddying the waters... :-(

corinnewelsh:
@ntoll no need for apologies. This stuff is difficult and it's good we're discussing it.

dpgeorge:
I'll close this PR because there was no consensus and the conversation finished long ago.

If anyone is interested in picking this up again, this can be reopened, or a new PR created (and see MicroPython's code-of-conduct for inspiration: https://github.com/micropython/micropython/blob/master/CODEOFCONDUCT.md).

PR #195, opened 2016-02-06
Tiny change to help_text to document paste mode

paulegan:
Paste mode is super useful once you go beyond a few lines and are making use of an editor on the side. However it's not so easy to discover how to enable it unless you know where to look in the micropython documentation. I think it deserves to be up front in the help.


dpgeorge:
Nice!

PR #205, opened 2016-02-11
Update DAL to v1.4.15, and swap capabilities of pin4 with pin5

dpgeorge:


ntoll:
I'll test when I get home this evening. :-)

JoeGlancy:
Can you also change the mount command in the Make file to add a mount path (`mnt`)? I've found it necessary to avoid simple issues or to avoid adding an entry for fstab

dpgeorge:
@JoeGlancy best solution is to make your own "makefile" file which overrides the "Makefile", eg, my "makefile" lives in the root directory and contains:
```
all2: all
	# print size of firmware after building
	@size build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython

include Makefile

# use sdc device
deploy: $(HEX_FINAL)
	$(ECHO) "Deploying $<"
	@mount /dev/sdc
	@sleep 1s
	@cp $< /mnt/
	@sleep 1s
	@umount /mnt
```

JoeGlancy:
I need a mount point for it though, but I can just use my own. Would it
make a difference if you added the mount point?

dpgeorge:
> Would it make a difference if you added the mount point?

No, because we specify `/mnt` in the `cp` and `umount` commands.  I think it would be better to make it configurable with variables, eg:
```
MOUNT_DEVICE ?= /dev/sdb
MOUNT_POINT ?= /mnt

deploy: $(HEX_FINAL)
	$(ECHO) "Deploying $<"
	@mount $(MOUNT_DEVICE)
	@sleep 1s
	@cp $< $(MOUNT_POINT)
	@sleep 1s
	@umount $(MOUNT_POINT)
```
Would that help?  Then you can even override it in "makefile" as such:
```
MOUNT_POINT = $(HOME)/mymountpoint
```

JoeGlancy:
I meant in the mount command, having it in cp and unmount assume it was
mounted at /mnt. Using variables sounds like a better idea though.

dpgeorge:
The problem with adding the mount point to the mount command is that then you need root privileges to run the command.

The other option is to specify the mount point instead of the device, ie:
```
    @mount $(MOUNT_POINT)
```
Would that work for you?

dpgeorge:
@ntoll this is ok to merge, if you checked it.  We can do the makefile stuff separately.

ntoll:
@dpgeorge sorry matey... had no time yesterday evening. Will check/merge over lunch. ;-)

ntoll:
Yay! I'll update uflash, mu and the TD editor. :-)

dpgeorge:
We should be pretty close now to v1.0.

JoeGlancy:
@dpgeorge I `deploy` it as superuser anyway, but I'll check soon.

PR #209, opened 2016-02-14
Modify fake data structure to keep it aligned with the DAL. 

### LINKED ISSUE #208: Image instantiation BROKEN :-(
To recreate, with the latest version of MicroPython from master:

```
from microbit import *

i = Image("01010:"
          "01010:"
          "01010:"
          "11111:"
          "01110:")
display.show(i)
```

Should display a very dim sailing ship. It currently displays random pixels at full brightness. If I change the brightness to `9`, it works perfectly. It appears, any value not `0` or `9` causes the problematic output.

Has the recent update to the DAL caused this problem..? (E.g. pin4 and pin5 swapping capabilities.)

Would it be possible to fix this as a matter of some urgency. The BBC are currently testing MicroPython and if they discover this is broken it'll create a bunch of bureaucratic delays I'd rather we not face.

@dpgeorge @markshannon any ideas..?

markshannon:
Fixes https://github.com/bbcmicrobit/micropython/issues/208

### DIFF
--- a/source/microbit/microbitdisplay.cpp
+++ b/source/microbit/microbitdisplay.cpp
@@ -244,6 +244,7 @@ struct FakeMicroBitDisplay : public MicroBitComponent
     uint8_t mode;
     uint8_t greyscaleBitMsk;
     uint8_t timingCount;
+    uint8_t errorTimeout;
     Timeout renderTimer;
 };
 


dpgeorge:
Thanks!  I checked and it fixed the bug.

Probably we should make a bunch of simple tests to test the microbit module, so that this has less chance of happening again.

ntoll:
You guys are too quick... I'll update Mu, urepl and the TD editor. Thanks! You're life savers! Gold stars all round. ;-)

PR #213, opened 2016-02-14
Fix error in compass.py script in examples.

ntoll:
What it says.

### DIFF
--- a/examples/compass.py
+++ b/examples/compass.py
@@ -18,4 +18,4 @@ compass.calibrate()
 while True:
     sleep(100)
     needle = ((15 - compass.heading()) // 30) % 12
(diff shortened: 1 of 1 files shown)


dpgeorge:
I don't remember it even being "Images" with a plural...

eaglestone:
The change from `Images` to `Image` does not seem to be coming through on the [docs site](https://microbit-micropython.readthedocs.org/en/latest/compass.html#example)?

dpgeorge:
You need to retrigger a build of the docs manually.  Now done.

eaglestone:
Thanks. Is that something I could have done?

dpgeorge:
> Is that something I could have done?

Yes, anyone can do it.  But you've got to know the right things to click on :)

PR #216, opened 2016-02-22
Doc updates

ntoll:
This branch introduces four changes:

* Add missing modules to the help text.
* Include an appropriate BLE place-holder in the docs.
* Add a "Next Steps" conclusion to the tutorials.
* Update the `this` module with a full list of the current contributors.

### DIFF
--- a/docs/lessons/next.rst
+++ b/docs/lessons/next.rst
@@ -1,0 +1,34 @@
+Next Steps
+----------
+
+These tutorials are only the first steps in using MicroPython with the
+BBC micro:bit. A musical analogy: you've got a basic understanding of
+a very simple instrument and confidently play "Three Blind Mice".
+
+This is an achievement to build upon.
+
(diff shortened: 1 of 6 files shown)


docs/lessons/next.rst
```diff
@@ -0,0 +1,34 @@
+Next Steps
+----------
+
+These tutorials are only the first steps in using MicroPython with the
+BBC micro:bit. A musical analogy: you've got a basic understanding of
+a very simple instrument and confidently play "Three Blind Mice".
+
+Please don't be disheartened!
```
sparkslabs: Why not change this to - a positive statement instead - like "This is great!"

Or "... Which means you can do more than most adults!"

Or similar.

ntoll: Good point. Let me try something as you suggest.

ntoll:
Fixed, squashed and re-based.

PR #217, opened 2016-02-24
Rename lessons->tutorials. Add table of contents to tutorials.

ntoll:
What the title says. This is to clarify the intent of this content. "Projects" and "Lesson" will be incoming later this week. ;-)

Would be good to get this merged ASAP so I can share with the BBC.


dpgeorge:
I think the table of contents should only have depth 1, it's too long otherwise.

Also, probably the contents table for the tutorials should go after the introductory text.

Issue #218, opened 2016-02-24
Weird issue with music wait=False


gbaman:
Was fiddling around with doing piano stuff, where the note doesn't stop till something else stops it (holding down the key on the piano then let go).
I had this code below.
```python
from microbit import *
import music
print("Starting")

sleep(1000)

music.play('c4:4', wait=False)

while True:
    print("Test")
    sleep(1000)
```
When I ran it on the Microbit using Mu, got a bit of a surprising result... Is only an issue when I set ```wait=False```, which is required to do my idea and use ```.stop()``` later on.
![screen shot 2016-02-23 at 00 40 28](https://cloud.githubusercontent.com/assets/833193/13302090/fabfb72c-db3f-11e5-9fb3-5c94be18a205.png)

dpgeorge:
Yep, that is weird!  Likely it's something to do with trying to convert an object to a string to play the note, but instead of a string there is a boolean.  The strange part is, where does the boolean come from?  I'll look into it.

dpgeorge:
Work around is to put the single note in a list:
```
music.play(['c4:4'], wait=False)
```

ntoll:
Fixed by #219

PR #219, opened 2016-02-25
Put music async object pointers in root pointer structure.

dpgeorge:
This makes sure that wait=False works correctly by keeping the notes around during their playing.

Addresses issue #218.


ntoll:
Tested. It works! :-)

PR #220, opened 2016-02-25
During a display animation update, catch any exceptions from iternext().

dpgeorge:
Because if an exception is raised in the interrupt which handles the
display update, then this exception branches outside the interrupt
handler and interrupts remain disabled forevermore.

Can test with the following (which locks up prior to this patch):

    def f(n)
        display.scroll(str(n))
        f(n + 1)
    f(0)


ntoll:
It works! Yay!

PR #221, opened 2016-02-25
New, much more efficient ticker. Reduces overhead of display rendering by about 90%.

markshannon:
This is the first PR in a series aimed at bringing synthesised sound capabilities to the micro:bit.
Before we can handle sampled sound, we need an efficient and accurate timer.
(The hardware offers a range of quality timers, it is the software that is the problem).

The current timer used for rendering the display and doing the general updates is inefficient and jittery.
For example, on master, 
```python
display.show(Image("11111:"*5))
music.play(...)
```
will cause visible flicker as the tune plays.
"Greyscale" rendering can also interfere with the UART operation making `pyboard` unreliable.

With the new timer, there is no visible flicker regardless of the load, and the UART seems to operate correctly even with a "greyscale" image showing.

As well as being much more efficient, adding callbacks to the new timer never disables interrupts. This ensures that interrupts arrive exactly on time, which is important for sampled sound.

Some of the reduced overhead is from improvements to display rendering, but the vast majority is from the new timer.

## Performance Measurements

Unloaded throughput (in arbitrary units of work): 11974 (estimated 99.95% CPU available).
Estimated 100% throughput: 12034

Timer overhead is tested under two conditions.

1. The display cleared. 
2. Displaying the image:
    `Image("01234:56789:01234:56789:01234")`

### Throughput

Setup | Master | New Timer
------|--------|----------
Clear | 10959  | 11864
Image | 8669   | 11691

### Timer Overhead

Setup | Master | New Timer
------|--------|----------
Clear | 8.9%   | 1.4%
Image | 28.0%   | 2.9%

### Test program
```python
from microbit import *

if 0:   
    display.show(Image("01234:56789:01234:56789:01234"))

def use_time():
    i = 3
    while i:
        i -= 1
        
t = running_time()
work_done = 0

while running_time() - t < 5000:
    use_time()
    work_done += 1
    
print (work_done)
```

### DIFF
--- a/inc/lib/ticker.h
+++ b/inc/lib/ticker.h
@@ -1,0 +1,30 @@
+#ifndef __MICROPY_INCLUDED_LIB_TICKER_H__
+#define __MICROPY_INCLUDED_LIB_TICKER_H__
+
+/*************************************
+ * 62.5kHz (16µs cycle time) ticker.
+ ************************************/
+
+#include "nrf.h"
+
+void ticker_init(void);
+void ticker_start(void);
+void ticker_stop(void);
+
+typedef void (*callback_ptr)(void);
+typedef int32_t (*ticker_callback_ptr)(void);
+
+int clear_ticker_callback(uint32_t index);
+int set_ticker_callback(uint32_t index, ticker_callback_ptr func, int32_t initial_delay_us);
+
+int set_low_priority_callback(callback_ptr callback, int id);
+
+#define CYCLES_PER_MICROSECONDS 16
+
+#define MICROSECONDS_PER_TICK 16
+#define CYCLES_PER_TICK (CYCLES_PER_MICROSECONDS*MICROSECONDS_PER_TICK)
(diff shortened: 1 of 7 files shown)


dpgeorge:
This is great!  And I think it's an important step to decouple more from the DAL, in anticipation of (hopefully) a more component oriented DAL.

I checked the timings using the test script given above and do indeed see big improvements (especiall with grayscale image showing).  My numbers are:

Setup | Master | New Timer
------|--------|----------
Clear | 10600  | 11530
Image | 8400   | 11350

Regarding increase in code size, I see that it adds just over 500 bytes of code (due mostly to the new ticker code) and also 20 bytes of data and 16 bytes of bss.  I guess those numbers are ok (if we remove the DAL we would regain much more than that).

Main problem: compass.calibrate() is now broken.  So we'd need a way to fix this before merging.

markshannon:
`compass.calibrate()` is now working again.

PR #223, opened 2016-02-29
Refer to nested suites as "nesting"

ncoghlan:
The term "scoping" has a specific meaning in Python (related to variable
scopes), so it's preferable to use another term like "nesting" to
refer to nested suites.

### DIFF
--- a/docs/tutorials/buttons.rst
+++ b/docs/tutorials/buttons.rst
@@ -102,7 +102,7 @@ It's obvious that the main tasks are broken down into sub-tasks that are
 ``Bacon`` and ``Tomatoes`` are obviously related to ``Shopping``. By indenting
 things we make it easy to see, at a glance, how the tasks relate to each other.
 
-This is called *scoping*. We use scoping to define blocks of code like this::
+This is called *nesting*. We use nesting to define blocks of code like this::
 
     from microbit import *
 


ncoghlan:
There were a couple of other references to scopes which I've adjusted in this PR: https://github.com/bbcmicrobit/micropython/pull/224

However, one of those assumes the reader will understand the phrase "loop body", so it may make sense to introduce that here.

dpgeorge:
@ntoll, this one's for you :)

ntoll:
Hi @ncoghlan.  I agree we should change scoping to something because of the reason you state. However, I already introduce the term "nesting" in the context of nesting function calls on the same page (see the section that talks about Matrioshka dolls). I'm anxious that we don't overload the teachers who are reading this with potentially confusing re-definitions.

How about replacing the line you've changed with..?

```
Python works in exactly the same way: indentation indicates blocks of related code (confusingly, this is also called "nesting")::
```

Thoughts..? I'm not precious about any of the things I've written so all feedback, good, bad, indifferent is most welcome! :-)

ncoghlan:
https://github.com/bbcmicrobit/micropython/commit/94eb07342cbcebbc9fd43c7a2f4f9e93ea1f7592 now changes the later references to "scope" to instead refer to the body of the loop.

The official terms for distinguishing between "statements with a nested suite of code" and "statements without a nested suite of code" are "compound statements" and "simple statements", so some combination of "suite" and "compound statement" would hold people in good stead all the way up to and including actually reading the grammar file in CPython's source code. However, that may be above the reading level you're aiming for.

ntoll:
Hi @ncoghlan can you rebase with master and I'll merge. Thanks! :-)

ncoghlan:
Done (the commit still claims to be 23 days old in the GitHub UI, but has actually been rebased)

dpgeorge:
Thanks!

PR #224, opened 2016-02-29
Avoid the word "scope"

ncoghlan:
"Scope" has a specific meaning in Python (in the sense of "variable
scope") so it's best to avoid it when discussing other concepts.

### DIFF
--- a/docs/tutorials/movement.rst
+++ b/docs/tutorials/movement.rst
@@ -30,7 +30,7 @@ If you hold the device flat it should display ``-``; however, rotate it left or
 right and it'll show ``L`` and ``R`` respectively.
 
 We want the device to constantly react to change, so we use an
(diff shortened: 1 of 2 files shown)


ncoghlan:
The change to the "Movement" tutorial here refers to a "loop body", but the corresponding change to the previous introduction of scoping doesn't currently introduce that term: https://github.com/bbcmicrobit/micropython/pull/223

Perhaps it should? It's the right term for the concept that was previously being referred to as a scope.

dpgeorge:
I'm happy with this.  @ntoll what about you?

ntoll:
This is a great catch @ncoghlan. I'll take this branch, introduce "loop body" in an appropriate way, rebase and merge.

ntoll:
Rebased and merged.

PR #225, opened 2016-03-01
PWM based on new ticker and GPIO. 

markshannon:
New implementation of (low-frequency) pulse width modulation.
The motivation for this is to free up the GPIOTE and PPI components for use by the (upcoming) sound module. 

Compared to master, this implementation has advantages (free choice of pins) and disadvantages (lower speed and increased overhead).

When used for driving servos or motors, the free choice of pins is a distinct advantage.
When used for generating sound, the lack of resolution and low-frequency are a problem.

However, the upcoming sound module will make the high-frequency limitations irrelevant.
Until the sound module is available, the music module will continue to use the Nordic/mbed PWM, so there will be no change in the behaviour of the music module.

### Feature comparison

     Feature  | Master | New Timer
-------------|--------|----------
Min period | 20µs | 1000µs
Max period | 131ms  | 1000ms
Resolution | 4µs | 16µs
Usable pins | 6  | all
Max simultaneous pins | 3 | all
CPU cost (3 pins/20ms) | ~0.6% | ~1.2% |


source/lib/pwm.c
```diff
@@ -0,0 +1,183 @@
...
+        if (event->timestamp >= timestamp) {
+            break;
+        }
+        previous = event;
+        event = event->next;
+    }
+    if (event->timestamp == timestamp) {
+        event->pins |= (1UL<<pin);
+    } else {
+        event = gc_alloc(sizeof(pwm_event), false);
```
dpgeorge: Do you need to allocate on the heap?  Can it be statically allocated?

dpgeorge:
When the new sound module comes along, will we still have the music module with the same API, and same functionaity?

source/lib/pwm.c
markshannon: It could, but then we would be using RAM even if PWM is unused.
This way we don't need to pre-determine how many pins can be supported.

markshannon:
There will be a change in that two pins will be required, rather than one pin and earth. There seems to be no other way to get sufficient fidelity for speech otherwise.
Otherwise, the music API will be exactly the same.

I also plan to add the ability to specify an "instrument", i.e. waveform generator, but using a keyword argument, so it will be backwards compatible.

dpgeorge:
From what I understand the way this new PWM works is as follows: there's a circularly linked list with one element for each event (usually each new pin) and a final element to turn them all of and restart to cycle.

Whenever the duty cycle changes, an old event is deleted and a new one created and inserted into the chain.

For applications that do a lot of changing of the duty cycle (eg in a loop) it's inefficient to keep allocating memory each duty change.  At the least it could remember the last free event structure and reuse it instead of reallocating via the GC.

The set variable in pwm_event is also not necessary: they only time this is not "true" is for all_off, so you could just have all_off being a special case.  But, actually, optimising this is not going to do anything because the GC always allocates in multiples of 4 words (ie 16 bytes is minimum allocation size).

> There will be a change in that two pins will be required, rather than one pin and earth.

You mean that the music module itself will need to use 2 pins?  I'd say that's a no-go, since it changes not only the API but also the way you physically connect things.  This is not something we can change (easily).  I know that you can get more volume using 2 pins out of phase.  Maybe we can have an option to use 2 pins, but the default is 1 pin plus ground?

As for the bigger picture, I don't think we should merge this as part of master.  We should make a new branch (eg v2-dev) and do development there.  Then once all the new features land, things can be tidied up, optimised, documented, etc, and then it can be merged as v2.0 (or something like that).

ntoll:
+1 on another branch. This would make it a *lot* easier to work with the BBC on this when it comes time to release it. Any update to the web-based editor in TouchDevelop would need to go through a "process" that should be transparent to you, but I think it important that you know that there is one (essentially, a bunch of testing and checks around documentation and so on).

markshannon:
I've fixed up the change of duty cycle code to reuse the same event object.

A development branch sounds like a good idea. I've created https://github.com/bbcmicrobit/micropython/tree/audio-dev and created a new PR against that branch: https://github.com/bbcmicrobit/micropython/pull/227

I don't really see why changing the default pair of pins from `0` and `GND` to `0` and `1` is an issue.
I'll put the full explanation of why we need two pins in the relevant PR, but it does help a lot better being able to drive current in two directions.

PR #232, opened 2016-03-11
Fix missing comma(s) in list of 8 ball responses

moreati:
The penultimate item in the 8-ball program is missing a trailing comma. I'm assuming this is an omission, and not an intentional big for the learners to fix.

I also took the liberty of adding a comma after the final item. If this is a bad feature/practise to show the learners please let me know and I'll amend this PR.


ntoll:
It's a bug.

ntoll:
Merged via rebase.

PR #238, opened 2016-03-13
Audio module. Basic ability to play iterables of audio-frames.

markshannon:
Prerequisite for `synth` and (hopefully) `speech` modules.


docs/audio.rst
```diff
@@ -0,0 +1,71 @@
...
+
+.. note::
+    You don't need to understand this section to use the ``audio`` module.
+    It is just here in case you wanted to know how it works.
+
+The ``audio`` module consumes samples at 7812.5 kHz, and uses linear interpolation to
+output a PWM signal at 32.5 kHz, which gives tolerable sound quality.
+
+The function ``play`` fully copies all data from each ``AudioFrame`` before it
+calls ``next()`` for the nexy frame, so a sound source can use the same ``AudioFrame``
```
dpgeorge: nexy->next

inc/microbit/modaudio.h
```diff
@@ -0,0 +1,29 @@
...
+#include "py/runtime.h"
+
+void audio_start(void);
+void audio_stop(void);
+mp_obj_t audio_init(void);
+void audio_play_source(mp_obj_t src, bool wait);
+void audio_set_pins(mp_obj_t pin0_obj, mp_obj_t pin1_obj);
+
+#define LOG_AUDIO_CHUNK_SIZE 5
+#define AUDIO_CHUNK_SIZE 32
```
dpgeorge: #define AUDIO_CHUNK_SIZE (1 << (LOG_AUDIO_CHUNK_SIZE))

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,430 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "string.h"
+#include "MicroBit.h"
```
dpgeorge: Do you use anything from MicroBit.h?  Would be nice not to have it as a dependency.  Might even be able to make this file a .c file instead of .cpp.

inc/microbit/modaudio.h
```diff
@@ -0,0 +1,29 @@
...
+#define AUDIO_BUFFER_SIZE (AUDIO_CHUNK_SIZE*2)
+#define AUDIO_CALLBACK_ID 0
+
+typedef struct _microbit_audio_frame_obj_t {
+    mp_obj_base_t base;
+    int8_t data[AUDIO_CHUNK_SIZE];
+} microbit_audio_frame_obj_t;
```
dpgeorge: This is not great from the point of view of the memory manager, whose allocation quanta is 4 words = 16 bytes.  Would be best to have AUDIO_CHUNK_SIZE=28, but probably that's hard to do (28 is not a power of 2).

examples/waveforms.py
```diff
@@ -0,0 +1,40 @@
...
+def show_wave(name, frame):
+    display.scroll(name + " wave", wait=False,delay=100)
+    for i in range(32):
+        print(frame[i], end=", ")
+    print()
+    audio.play(repeated_frame(frame, 1500))
+
+frame = audio.AudioFrame()
+
+for i in range(32):
```
dpgeorge: Can you implement `len(frame)` to avoid need for magic number 32?

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,430 @@
...
+    mp_obj_t pins = args[2].u_obj;
+    if (pins == mp_const_none) {
+        audio_set_pins((mp_obj_t)&microbit_p0_obj, (mp_obj_t)&microbit_p1_obj);
+    } else {
+        mp_obj_t len = mp_obj_len(pins);
+        if (mp_obj_get_int(len) != 2) {
+            nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError, "wrong number of pins"));
+        }
+        mp_obj_t pin0 = mp_obj_subscr(pins, MP_OBJ_NEW_SMALL_INT(0), MP_OBJ_SENTINEL);
+        mp_obj_t pin1 = mp_obj_subscr(pins, MP_OBJ_NEW_SMALL_INT(1), MP_OBJ_SENTINEL);
```
dpgeorge: There is an API function just for this pattern: `mp_obj_get_array_fixed_n(pins, 2, &items);`

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,430 @@
...
+STATIC mp_obj_t play(mp_uint_t n_args, const mp_obj_t *pos_args, mp_map_t *kw_args) {
+    static const mp_arg_t allowed_args[] = {
+        { MP_QSTR_source, MP_ARG_REQUIRED | MP_ARG_OBJ, {.u_obj = MP_OBJ_NULL} },
+        { MP_QSTR_wait,  MP_ARG_BOOL, {.u_bool = true} },
+        { MP_QSTR_pins,   MP_ARG_OBJ, {.u_obj = mp_const_none } },
+    };
+    // parse args
+    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
+    mp_arg_parse_all(n_args, pos_args, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
+    mp_obj_t src = args[0].u_obj;
```
dpgeorge: Does it ever validate that src is an AudioFrame?  Does it even matter?

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,430 @@
...
+        mp_int_t value = mp_obj_get_int(value_in);
+        if (value < -128 || value >= 128) {
+            nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError, "value out of range"));
+        }
+        self->data[index] = value;
+        return mp_const_none;
+    }
+}
+
+const mp_obj_type_t microbit_audio_frame_type = {
```
dpgeorge: AudioFrame looks exactly like a bytearray (actually array.array('b', ...)).  Why not just use array.array?

dpgeorge:
I tried the example script.  Only the square and sawtooth sounds are audible for me (and they are very clean!).  The sine and triangle are not.

source/microbit/modaudio.cpp
markshannon: I need to call `microbit_obj_get_pin`, which returns a `MicroBitPin` to get the pin name.

inc/microbit/modaudio.h
markshannon: You only need to allocate one of these for each audio source, as they should be reused, so the wastage only occurs the once.

examples/waveforms.py
markshannon: That would be nicer. I'll do that.

source/microbit/modaudio.cpp
markshannon: thanks

source/microbit/modaudio.cpp
markshannon: Since we are passing in an iterable, we can only check when we get the result of `next()` which happens in `audio_data_fetcher`

source/microbit/modaudio.cpp
markshannon: The combination of timer speed, ticker speed and output frequencies means that the buffer ought to be a fixed size. `array.array` is a variable size object.

Using a custom array is efficient as we don't have to check whether the input is a buffer and whether it is the right size.

I think it also makes the API simpler.

markshannon:
There is more energy in a sawtooth and square wave than in the other two. Also, the base frequency is only 250Hz, so it might be that a small unamplified speaker has much better output of the harmonics than the base frequency (or it could just be a bug).

What sort of speaker are you using?

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,430 @@
...
+STATIC MP_DEFINE_CONST_FUN_OBJ_0(microbit_audio_reset_obj, audio_reset);
+
+
+mp_obj_t audio_init(void) {
+    static bool initialised = false;
+    if (!initialised) {
+        audio_source_iter = NULL;
+        initialised = true;
+        //Allocate buffer
+        audio_buffer_ptr = (int8_t *)malloc(AUDIO_BUFFER_SIZE);
```
dpgeorge: Please use m_new(int8_t, AUDIO_BUFFER_SIZE).

dpgeorge:
> What sort of speaker are you using?

An 8 ohm, 0.8W, 20Hz-20kHz speaker, with no amplification.

I had a play converting byt hand the saw tooth (which works) into a triangle (which doesn't work) and it seems that you only get something audible when there is a large step in the waveform (which the square and sawtooth both have).

How are you testing it?

markshannon:
I'm using a mylar speaker I bought for about £2.50 from Maplins.I don't know the spec.
I also tested it with my normal PC speakers, which are amplified and have decent base.

There are various reasons why a square wave seems louder than a sine wave:
http://electronics.stackexchange.com/questions/179906/which-is-the-loudest-in-terms-of-sound-intensity-sine-square-or-sawtooth-wave

Small speakers don't handle lower frequencies well. With an amplified speaker with decent base, the difference between sine and square waves seems much less.

I've added a final output which gradually changes from triangle to square wave. The perceived volume changes smoothly, as one would expect.

dpgeorge:
Ok, I'll try again.  It's strange though that it doesn't work for me on this speaker, because I did have basic pre-recorded samples playing through this same speaker....

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,439 @@
...
+}
+
+static int32_t audio_ticker(void);
+
+volatile int32_t audio_buffer_read_index;
+
+#define AUDIO_BUFFER_MASK (AUDIO_BUFFER_SIZE-1)
+
+#define audio_buffer_ptr MP_STATE_PORT(audio_buffer)
+#define audio_source_iter MP_STATE_PORT(audio_source)
```
dpgeorge: Need to move these macros to before the audio_stop() function.

dpgeorge:
I tried it with an amplified speaker.  The sine is still barely audible, but the triangle is correct (although still quite compared with square/saw-tooth).  The smooth triangle-to-square waveform also works.

I'd suggest merging after #240 to take advantage of that new pin function.

source/microbit/modaudio.cpp
ntoll: Yes please... I just got bitten by this.

ntoll:
Hi @markshannon, it's great to see this development taking place. 

I ran the test script: sine is very quiet but detectable and triangle is a tad louder. Square and saw are nice and loud and I like the very effective crescendo (!) for the ascending wave. All tested on a cheap plastic speaker from Maplin. :-)

I realise there's not really much that I can helpfully contribute here.... but I wanted to have a play. ;-)

Any reason why the sine is so quiet?

markshannon:
I think it is a combination of our ear's frequency response, the speaker's frequency response and that the sine wave contains the least power. 
I will do some more experiments with a 'scope to check that the output is as I expect it to be.

markshannon:
This is the unamplified speaker, I was using:
http://www.farnell.com/datasheets/1662278.pdf
It has a rated minimum frequency of 350Hz, and the sine wave in the example is a 250Hz, so it isn't surprising that it is quiet.

Producing various "DC" signals:
```python
dc = audio.AudioFrame()
for i in range(len(dc)):
    dc[i] = K
def out():
      while True:
           yield dc
```
the output on the scope, looks to be as I would expect: 31.25kHz rectangular wave with duty cycle proportional to `K`.
With no load, 300mV when on.
With the speaker load, an instantaneous peak of 300mV dropping very quickly to about 40mV (8Ω * 5mA).

PR #240, opened 2016-03-18
Add microbit_obj_get_pin_name function to reduce need for MicroBit.h.

dpgeorge:
Added microbit_obj_get_pin_name, which returns the PinName value (from mbed API), and reduces the need for MicroBit.h.  There are no functional changes.

@markshannon this should mean you can make modaudio a pure .c file.

### DIFF
--- a/source/microbit/microbitspi.cpp
+++ b/source/microbit/microbitspi.cpp
@@ -24,13 +24,12 @@
  * THE SOFTWARE.
  */
 
-#include "spi_api.h"
-#include "MicroBit.h"
-
 extern "C" {
 
+#include "spi_api.h"
 #include "py/runtime.h"
 #include "modmicrobit.h"
+#include "microbitpin.h"
 #include "microbitobj.h"
 
 typedef struct _microbit_spi_obj_t {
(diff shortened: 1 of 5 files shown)


source/microbit/microbitspi.cpp
```diff
@@ -57,17 +55,17 @@ STATIC mp_obj_t microbit_spi_init(mp_uint_t n_args, const mp_obj_t *pos_args, mp
         MP_ARRAY_SIZE(allowed_args), allowed_args, (mp_arg_val_t*)&args);
 
     // get pins
-    PinName p_sclk = MICROBIT_PIN_P13;
-    PinName p_mosi = MICROBIT_PIN_P15;
-    PinName p_miso = MICROBIT_PIN_P14;
+    PinName p_sclk = P0_23; // microbit edge pin P13
```
markshannon: We ought to centralise the pin mapping somewhere, if we aren't going to use the DAL headers.
`inc/microbit/microbitpin.h`, perhaps?

markshannon:
One quibble, otherwise LGTM.

dpgeorge:
I added inc/microbit/microbitpin.h as suggested, and re-pushed.

My yotta is broken in the moment so I can't test the build...

markshannon:
LGTM, builds cleanly on my machine and seems to work OK. Any particular tests you would suggest?

dpgeorge:
> Any particular tests you would suggest?

Well... either UART, SPI or neopixel :)  They all need external hardware.  Neopixel would be the easiest to test.  I can try it tomorrow.

dpgeorge:
Builds fine for me now.  I checked with a NeoPixel and it works.  Also rebased it.

markshannon:
OK, merging.

PR #242, opened 2016-03-21
Enable built-in "set" and "frozenset" types.

dpgeorge:
When I first put together the MicroPython microbit build I disabled everything to get it to fit, then enabled stuff as deemed necessary.  I never added "set" or "frozenset", but I guess they should be included because they are basic built-in types.

Adds 3368 bytes to code size.

Any objections?


ntoll:
No objection from me. I'll wait until this evening to merge in case there is  legitimate objection raised for including a builtin. ;)

ntoll:
OK... I presume there are no objections. :-)

PR #244, opened 2016-03-23
Add internal USB uart pins; modify uart.init method

dpgeorge:
There are 3 commits here:

- microbit.pin_usb_tx and microbit.pin_usb_rx pins are added, to expose the internal USB-serial pins.  This is useful so you can configure the USB UART peripheral (eg change the baudrate of the REPL, or restore the REPL after using the UART on external pins).

- uart.init() is modified to add the 'tx' and 'rx' arguments, to specify the tx/rx pins.  Originally there was the 'pins' argument for this, which is a tuple of (tx, rx) pins, but it was discussed (somewhere, long ago) that using 'tx' and 'rx' parameters is clearer and self documenting.  It also follows the scheme used by spi.init (which has sclk, mosi and miso args for the pins).

- Docs are updated to reflect the changes.

I'm not precious about the names pin_usb_tx/pin_usb_rx, so if anyone has better ideas then please say!


dpgeorge:
It uses 36 bytes of RAM, for the new pin objects (overhead from the DAL's way of creating pins).

deshipu:
I wonder if it wouldn't make more sense to make the ``tx`` and ``rx`` parameters optional, and use the USB pins when those are not specified or ``None``? I mean, we are not going to need them anywhere else.

dpgeorge:
> I wonder if it wouldn't make more sense to make the tx and rx parameters optional, and use the USB pins when those are not specified or None?

That's a good idea, keeps it much simpler.  I'll change the PR later.

dpgeorge:
This PR has now been updated to be a lot simpler, as per @deshipu's suggestion.

docs/uart.rst
```diff
@@ -10,17 +10,19 @@ a serial interface.
 Functions
 =========
 
-.. method:: init(baudrate=9600, bits=8, parity=None, stop=1, pins=None)
+.. method:: init(baudrate=9600, bits=8, parity=None, stop=1, \*, tx=None, rx=None)
```
ntoll: what's the `\*` for..? If a typo please fix. ;-)

ntoll:
@deshipu have you tested this?

docs/uart.rst
deshipu: That's to mark the tx and rx arguments as keyword-only -- that is, they can't be specified as positional arguments.

ntoll: @deshipu thanks for the heads up... I'd never seen that before. You learn something new every day... ;-)

deshipu: It was added to Python 3 at some relatively late point.

deshipu:
I have not, I shall do it now.

deshipu:
Works perfectly.

ntoll:
Great stuff!

dpgeorge:
Thanks!

Issue #246, opened 2016-03-26
Create "filesystem" (local storage) module based on DAL implementation


ntoll:
What the title says.

markshannon:
The persistent module referenced above was designed to store audio samples, so doesn't have a file-like API.
I've just had a look at the DAL PR: https://github.com/lancaster-university/microbit-dal/pull/110 and it seems to be rather inflexible in terms of file size and number, and rather inefficient as well. It may also cause problems with the hardware by repeatedly writing to the same areas of the flash.

As to the API, I would suggest:
```python
open(name, mode='r') #returns a file descriptor
remove(name) # Removes a file
fd.read(len=None) # Read len bytes (or the whole file if len is None)
fd.write(buf) # Writes the bytes in buf to the file.
fd.close() # Writes any necessary metadata to the file-system.
```
I have a prototype implementation built on top of https://github.com/bbcmicrobit/micropython/pull/249 that organises 20kb of flash into 80 256-byte chunks, with 254 usable bytes in each, Files can be formed of any integral number of chunks. The design aims to keep erasure and writing to a minimum and avoid any hotspots which might lead to hardware failure.

It might be nice to automatically execute any file called `main.py` on reset.

larryhastings:
I suspect you want it more granular than that--perhaps 160 128-byte chunks, 124 usable bytes in each?  There's a certain amount of overhead for packages (empty __init__.py files).  Also, if this is used to store precompiled bytecode files those will often be quite small.

Would it be possible to drop the "chunk" approach completely and simply pack the data as tightly as possible?  One-byte length, filename, two-byte length, file data?

markshannon:
The problem with packing the data is fragmentation.

Given the per-chunk overhead is only 2 bytes, then reducing the chunk size to 128 would make sense. There is an upper limit of 253 chunks (1 byte index - 3 special values) per "file system", so 64 byte chunks would be too small.
Another thing to consider is that the flash must be erased in 1k pages, although I doubt it make much difference whether the chunks are 128 or 256 bytes.

larryhastings:
ISTM there are two use cases for a "filesystem" on the micro:bit:

1. Storing Python source files, data tables, and other read-only data.
2. Storing dynamically-generated data (high scores, accelerometer telemetry).

I had only the first use case in mind, I gather you were thinking of both.

[...]

PR #249, opened 2016-03-27
Add persistent module. Fixes issues 246

markshannon:
Adds a new "persistent" module with two persistent 8k data blocks to which data can be saved, one of which is the start up script.

Includes new `--save` option to `pyboard` to save script without have to flash µPy firmware.

@ntoll Might be useful in `mu` as it could reduce the download by a factor of 50 or so.

Example use:
```python
import persistent
persistent.save_script("from microbit import display\ndisplay.scroll('Hello persistent world!')")
```
Then press the reset button...

Or load the slalom game:
```bash
./tools/pyboard.py --save /dev/ttyACM0 examples/simple_slalom.py
```

See https://github.com/bbcmicrobit/micropython/issues/246


markshannon:
Closing in favour of upcoming "file system" support.

PR #250, opened 2016-03-30
Prevent map resize failure from destroying map.

stephenkyle-ARM:
Previously if you tried to add an element to a map that couldn't be expanded further, this would leave the original map unable to be used (i.e., len(map) would return 0, and no elements were accessible). This defers changing the original map until a point where the resizing is known to have succeeded.

### DIFF
--- a/source/py/map.c
+++ b/source/py/map.c
@@ -118,11 +118,14 @@ void mp_map_clear(mp_map_t *map) {
 
 STATIC void mp_map_rehash(mp_map_t *map) {
     mp_uint_t old_alloc = map->alloc;
+    mp_uint_t new_alloc = get_doubling_prime_greater_or_equal_to(map->alloc + 1);
     mp_map_elem_t *old_table = map->table;
-    map->alloc = get_doubling_prime_greater_or_equal_to(map->alloc + 1);
+    mp_map_elem_t *new_table = m_new0(mp_map_elem_t, new_alloc);
+    // If we reach this point, table resizing succeeeded, now we can edit the old map.
+    map->alloc = new_alloc;
     map->used = 0;
     map->all_keys_are_qstrs = 1;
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks!  It would be good if this was merged upstream, https://github.com/micropython/micropython .  Were you able to test it?

stephenkyle-ARM:
Yep, it was tested on the microbit, see below. I'll submit a PR upstream as well.

`Before
======

MicroPython v1.6-11-gfa67b52 on 2016-02-01; micro:bit with nRF51822
>>> a = dict()
>>> for i in range(0, 347):
...     a[i] = i + 1
...     
...     
... 
>>> len(a)
347
>>> a[346]
347
>>> a[347] = 348
MemoryError: memory allocation failed, allocating 5176 bytes
>>> len(a)
0
>>> a[346]
KeyError: <value>

After
=====

MicroPython v1.6-11-gfa67b52 on 2016-02-01; micro:bit with nRF51822
>>> a = dict()
>>> for i in range(0, 347):
...     a[i] = i + 1
...     
...     
... 
>>> len(a)
347
>>> a[346]
347
>>> a[347] = 348
MemoryError: memory allocation failed, allocating 5176 bytes
>>> len(a)
347
>>> a[346]
347`

dpgeorge:
Great, thanks for getting it done upstream!

Leave this open until the latest version of upstream makes it into this repo.

ntoll:
@dpgeorge when will the next version of upstream make it into this repo..? ;-)

dpgeorge:
@ntoll It depends on when we want it :)  First we should make a concrete plan for getting v1.0 tagged.

ntoll:
OK... let's grab this from upstream!

dpgeorge:
Fixed by 1e1ea9d5e45ca12bf6d736c05b66539d2e90b3a2.

PR #251, opened 2016-04-06
Add file sytem and file module.

markshannon:
Includes the following features:

1. Allows files to written and read.
2. Files can be uploaded using `tools/upload.py`
3. Automatically runs the file `__main__.py` after a hard reset.
4. Modules can be imported from the file system.
5. Files will persist over reset and loss of power. However, flashing the device will wipe all files (this is an inherent limitation of the device).
6. ~20000 bytes of usable space and up to 160 files (this is configurable at compile time).

Example use:
```python
>>> f = open("test.py", "w")
>>> f.write("from microbit import display\n")
>>> f.write("display.scroll('Test module')")
>>> f.close()
>>> import test 
#Imports the `test` module from the newly created `test.py`
# The text "Test module" will be scrolled across the display
>>> test   
<module 'test'>
>>> import file
>>> file.list()
['test.py']
>>> file.len('test.py')
58
```

I have also taken the novel approach of including some tests :wink: 

Fixes https://github.com/bbcmicrobit/micropython/issues/246


dpgeorge:
Very nice!

A few things:

- Why `__main__.py`?  Upstream micropython uses `boot.py` and `main.py`.
- Why `file.list()` and not `os.listdir()`?  I would strongly suggest implementing the filesystem functions as per the os module, eg os.listdir, os.rename, os.remove and os.stat (instead of file.len).  Then it's proper Python :)
- There is a tool called rshell https://github.com/dhylands/upy-shell/ which allows you to access the filesystem on a "pyboard" via the REPL.  Maybe we should take this and improve it so it works also with the microbit?
- Maybe we should "hack" it so that the appended script (from the .hex file at the fixed location) is also part of the filesystem and can be opened and read (but not written)?  Eg using the name `hexmain.py`.

source/microbit/mprun.c
```diff
@@ -135,22 +150,17 @@ void gc_collect(void) {
 }
 
 mp_lexer_t *mp_lexer_new_from_file(const char *filename) {
-    (void)filename;
-    return NULL;
+    file_descriptor_obj *fd = microbit_open(filename, strlen(filename), false, false);
+    return microbit_file_lexer(qstr_from_str(filename), fd);
```
dpgeorge: What happens if fd is NULL due to no file?

markshannon: It crashes :cry:  I'll fix that.

markshannon:
Why `__main__.py`, because of `__name__ = "__main__"`. I'm not attached to it. Would you prefer `main.py`?
I didn't want an `os` module as that suggests a whole bunch of other stuff, but we might as well be as standard as possible.

Modifying rshell and exposing the appended script makes sense as well. Presumably not blocking for this PR, though?

source/microbit/modfiles.c
```diff
@@ -0,0 +1,596 @@
...
+        advance(self, to_write, true);
+        data += to_write;
+        len -= to_write;
+    }
+    return mp_const_none;
+}
+MP_DEFINE_CONST_FUN_OBJ_2(microbit_file_write_obj, microbit_file_write);
+
+static mp_obj_t microbit_file_writable(mp_obj_t self) {
+    return ((file_descriptor_obj *)self)->write ? mp_const_true : mp_const_false;
```
dpgeorge: mp_obj_new_bool

dpgeorge:
You can reduce the code size and get more file methods (eg readlines) if you reuse the stream wrappers.  Basically, all you need to implement is read, write and ioctl and then you have a full file object.  See https://github.com/micropython/micropython/blob/master/extmod/vfs_fat_file.c for hints how to do it.

dpgeorge:
> Why __main__.py, because of __name__ = "__main__". I'm not attached to it. Would you prefer main.py

Yes please, main.py would be consistent with other MP implementations.

> I didn't want an os module as that suggests a whole bunch of other stuff, but we might as well be as standard as possible.

Yes it would be good.  Don't need all os methods, just the ones you already have.  See here for hints:  https://github.com/micropython/micropython/blob/master/stmhal/moduos.c

> Modifying rshell and exposing the appended script makes sense as well. Presumably not blocking for this PR, though?

Sure.  We can do that later.

markshannon:
Issues addressed and branch rebased.
I've not implemented `os.stat(fd)` as having it return a tuple of one element seemed a bit silly. Instead we have `os.size(fd)` which returns the size.

I've also modified the test to more closely conform to the interface of https://github.com/bbcmicrobit/micropython/pull/254

dpgeorge:
Can you please modularise the source files a little bit?  I would suggest separating the non-uPy code from the uPy specific stuff.  Eg:

- uPy independent file IO stuff moved from modfiles.c to filesystem.c
- uPy file object stuff (textio, bytesio types and their methods) moved from modfiles.c to fs_file.c
- uPy file lexer stuff moved from modfiles.c to fs_lexer.c
- modfiles.c renamed to modos.c and uses funcs from filesystem.c and fl_file.c to implement the module
- modfile.h renamed to filesystem.h (and what about persistent.h?)

Note that modos could include other os functions like urandom, that have nothing to do with file io.

inc/microbit/mpconfigport.h
```diff
@@ -80,10 +80,10 @@ void mp_hal_stdout_tx_strn_cooked(const char *str, mp_uint_t len);
 
 // extra built in names to add to the global namespace
 extern const struct _mp_obj_fun_builtin_t mp_builtin_help_obj;
-extern const struct _mp_obj_fun_builtin_t mp_builtin_open_obj;
+extern const struct _mp_obj_fun_builtin_t microbit_open_obj;
```
dpgeorge: I don't see why this needs to be renamed.  It's still the builtin open function (and doesn't need to be microbit specific, it could eventually support opening of other kinds of files).

inc/microbit/modfile.h
```diff
@@ -0,0 +1,95 @@
...
+    union {
+        char data[DATA_PER_CHUNK];
+        file_header header;
+    };
+    uint8_t next_chunk;
+} file_chunk;
+
+#define FILE_NOT_FOUND ((uint8_t)-1)
+
+#define CHUNKS_IN_FILE_SYSTEM 160
```
dpgeorge: Probably need a comment that this can't be changed without changing the initialiser in persistent.c

source/microbit/persistent.c
```diff
@@ -0,0 +1,176 @@
...
+    int8_t *page = (void *)(((uint32_t)dest)&(-page_size));
+    int8_t *tmp_storage = NULL;
+    while (addr < end_data) {
+        int8_t *next_page = page + page_size;
+        uint32_t data_in_page = min(end_data-addr, next_page-dest);
+        if (can_write(dest, addr, data_in_page)) {
+            persistent_write_unchecked(dest, addr, data_in_page);
+        } else {
+            if (tmp_storage == NULL) {
+                tmp_storage = gc_alloc(page_size, false);
```
dpgeorge: This is kind of bad.  It means you need 1k of free heap (that's the nRF51's page size, right?) to do file writing.  What if the allocation fails?  Also, should use m_new(uint8_t, page_size).

test/test_files.py
```diff
@@ -0,0 +1,110 @@
+
```
dpgeorge: It should be tests, not test.

dpgeorge:
I think we should tag "v1.0" before merging this.

markshannon:
Why exclude this from `v1.0`?

dpgeorge:
> Why exclude this from v1.0?

Because it's a big change, because we need to draw the line somewhere, and because v1.0 has been a long time coming.  Current code has also been well tested so far and would be nice to make v1.0 the code that existed when the microbit was launched a few weeks ago.

markshannon:
I've changed the design of the filesystem to avoid using heap. It now uses n+1 pages instead of n. When there are no free pages, it moves the whole filesystem up or down one page which effectively erases the chunks of memory that aren't being used.
I've added more tests to stress this and check for correct behaviour for some corner cases.

I've also reorganised the code into:
modos.c which contains the OS module.
filesystem.c which contains the filesystem logic, and
fileobj.c which contains the object level wrappers.

Rebased.

source/microbit/filesystem.c
```diff
@@ -0,0 +1,436 @@
...
+static inline void *first_page(void) {
+    return &file_system_data[0];
+}
+
+static inline void *last_page(void) {
+    return &(((file_chunk *)&file_system_data)[CHUNKS_IN_FILE_SYSTEM]);
+}
+
+void microbit_filesystem_init(void) {
+    file_chunk *base = &(((file_chunk *)&file_system_data)[0]);
```
dpgeorge: This line gives me a strict alias warning.  If you change it to `file_chunk *base = (file_chunk *)&file_system_data;` then it doesn't give the warning and is semantically the same.

source/microbit/modmicrobit.cpp
```diff
@@ -73,6 +73,7 @@ STATIC mp_obj_t microbit_temperature(void) {
 }
 MP_DEFINE_CONST_FUN_OBJ_0(microbit_temperature_obj, microbit_temperature);
 
+
```
dpgeorge: Not needed.

source/microbit/modos.c
```diff
@@ -0,0 +1,46 @@
...
+
+#include "py/obj.h"
+#include "filesystem.h"
+
+STATIC MP_DEFINE_CONST_FUN_OBJ_1(microbit_remove_obj, microbit_remove);
+STATIC MP_DEFINE_CONST_FUN_OBJ_0(microbit_file_list_obj, microbit_file_list);
+STATIC MP_DEFINE_CONST_FUN_OBJ_1(microbit_file_size_obj, microbit_file_size);
+
+static const mp_map_elem_t _globals_table[] = {
+    { MP_OBJ_NEW_QSTR(MP_QSTR_remove), (mp_obj_t)&microbit_remove_obj },
```
dpgeorge: Need an entry for `__name__`.

source/microbit/mprun.c
```diff
@@ -105,9 +118,11 @@ void mp_run(void) {
     readline_init0();
     microbit_init();
 
-    if (APPENDED_SCRIPT->header[0] == 'M' && APPENDED_SCRIPT->header[1] == 'P') {
+    if (hard_reboot && APPENDED_SCRIPT->header[0] == 'M' && APPENDED_SCRIPT->header[1] == 'P') {
```
dpgeorge: This changes the behaviour of soft reboot: I got used to it re-executing the appended script and found it useful (sometimes).  Do we want/need to change this behaviour?

dpgeorge:
Ok, everything seems to work and it is very cool!  BUT, I get some "lag" issues: when executing the tests, the final "happy" image is not displayed until about 10 seconds after executing the line of code that does the display.  It prints "PASS" but then takes ages to show the image.  Also, if I then try to connect via USB serial, I don't get a prompt for a while.  It's hard to tell what the problem is...

source/microbit/mprun.c
markshannon: I think we do need this change. Otherwise how do you upload a file? 
If every time you ctrl-C the program restarts, there is no way to get into the repl.

markshannon:
@dpgeorge I've addressed your comments.
I think the delay between the print and the display updating was because I wasn't always stopping the timer for every write, which meant that the timer would get way ahead of the CPU. The pause was the CPU catching up (I think).
Anyway, the timer is now always stopped across writes and the problem has gone away.

ntoll:
@markshannon are we going to get os.rename and os.remove methods? Also, I'd like to explore how we can move forward on embedding scripts into the hex (as we currently do) in such a way that if the user does an os.listdir they see a "main.py" or "hexmain.py" file (containing the script we've embedded). I prefer the simpler "main.py" name, btw. I also like the idea of flashing *just* the Python file a la the utility @dpgeorge mentions, rather than the whole runtime. Perhaps something for uflash (upon which Mu sits)..? 

This is really great work. Whilst playing around with this branch my wife noticed I had a large silly grin on my face. :-)

I wonder if we could include the `json` module as a way to serialise simple Python data structures. I know there's already one for MicroPython.

Finally, I assume it'd be relatively easy to create a utility to extract the files in some way via USB/serial commands.

Really great stuff. Makes me itch to write lessons using this work... ;-)

markshannon:
@ntoll `os.remove` is already in this PR. 
`os.rename` would be a relatively expensive operation. I'm not saying we can't have it, but I'd rather leave it for another PR. Please file an issue if you think it is sufficiently important.

You can use `tools/upload.py` to directly upload files to the µBit, without having to flash it. I don't know how easy it would be to integrate it with mu.

source/microbit/mprun.c
dpgeorge: What we can do here is only execute the appended script, or main.py, if in "friendly REPL" mode.  If in "raw REPL" mode then don't execute the start up script.  This is how the pyboard works.  When you use the pyboard.py tool to upload/execute a script it will go into raw REPL mode so that the initial main.py is not executed.  I think that's how we should solve this problem.

dpgeorge:
@markshannon I've tested it again and still get some strange behaviour on occasion: after uploading using the tools/upload.py file I can't get a serial prompt.  I need to load the serial prompt, exit, then load it again and then it works.  It's hard to say where this problem comes from...

Did you look at using SPIFFS https://github.com/pellepl/spiffs ?  It's pretty minimal and works well with flash ROM.  There are no directories but that's ok for our use.

The benefit of using something like SPIFFS is that it is very well tested.  It's hard to test filesystem code.  A good way is to test it on a PC with a lot of demanding tests (using the RAM to store the data so it's fast).

markshannon:
I don't understand what you mean by "load the serial prompt". When you say "can't get a serial prompt", do you mean in picocom?
As you may have noticed `tools/upload.py` is a hacked up version of pyboard, so I suspect you understand it better than I.

SPIFFS looks far too RAM hungry and uses quite a lot of space for metadata. 
Since our "disk" is so small we can get away with linear scans to find files which keeps things simple and space efficient.

dpgeorge:
> I don't understand what you mean by "load the serial prompt". When you say "can't get a serial prompt", do you mean in picocom?

I mean access /dev/ttyACM0 using picocom.  If I run tools/upload.py, then picocom, then the device does not respond to characters.  I need to quit picocom and restart it to get a response (but this is not all the time, only sometimes... :( )

markshannon:
Odd, I've never seen that.
I often get problems connecting to the microbit with picocom, but that applies to any build (including master).

What file were you uploading? I'll see if I reproduce the problem.

markshannon:
@dpgeorge
You are only seeing problems after using `tools/upload.py`, not when using any of the `os` methods in the repl or a script. Is that correct?

dpgeorge:
@markshannon regarding the behaviour of script execution on reboot: What we can do here is only execute the appended script, or main.py, if in "friendly REPL" mode. If in "raw REPL" mode then don't execute the start up script. This is how the pyboard works. When you use the pyboard.py tool to upload/execute a script it will go into raw REPL mode so that the initial main.py is not executed. I think that's how we should solve this problem.  See https://github.com/micropython/micropython/blob/master/stmhal/main.c#L590

dpgeorge:
> You are only seeing problems after using tools/upload.py, not when using any of the os methods in the repl or a script. Is that correct?

At the moment, yes, tools/upload.py is the only time I see an issue.  I will try to find something reproducible.

dpgeorge:
I can't reproduce any problems with file writing/reading.  All the tests from the upstream MP repository that are expected to pass are still passing.

dpgeorge:
I've tested this again and it works fine so I say we can merge.  Just one last thing: the patch includes 20k of 0xff to create the filesystem.  This translates to roughly 60k increase in size of the .hex file.  It should be possible to avoid this, since the flash is anyway fully erased when uploading the file.  We can just use the region of flash at the end, but before the 8k chunk for the appended script.

source/microbit/mprun.c
markshannon: But how do you get to the raw REPL? Any attempt to leave the running script with ctrl-C will merely restart the script, so it impossible to enter the raw REPL.

markshannon:
I've modified the initialisation. 
It selects a region of ROM that starts after the program image and does not overlap any embedded script, so there is no need to incorporate a block of -1s into the hex file.

source/microbit/mprun.c
dpgeorge: > But how do you get to the raw REPL?

Type ctrl-A at the REPL (either raw or friendly) (note: double ctrl-A if using picocom, since ctrl-A is picocom's escape code).

> Any attempt to leave the running script with ctrl-C will merely restart the script, so it impossible to enter the raw REPL.

Ctrl-C should drop you to the REPL, not restart the script.

markshannon:
Is there anything blocking this?

markshannon:
And rebased...

dpgeorge:
> Is there anything blocking this?

We want to work out the issue with the script being executed on boot (or not).  Do you agree that the way other MP ports work is the correct thing for our case here?

markshannon:
I don't see how it is possible to get to the raw repl, if the script restarts every time there is an interrupt.
What am I missing?

dpgeorge:
> I don't see how it is possible to get to the raw repl, if the script restarts every time there is an interrupt.
> What am I missing?

If you press ctrl-C then it drops to the REPL, it doesn't restart.  It only restarts if you enter ctrl-D.  But if you first do ctrl-A then it goes to raw REPL and ctrl-D stays in raw REPL mode and doesn't execute any scripts.

Should I provide a PR to implement this behaviour, one that we can merge before the FS?

markshannon:
Yes please.

dpgeorge:
> Yes please.

See #265.

johnhunt:
Great work. I very recently just got my hands on a MB and I was pretty surprised that this functionality wasn't available already. Personally, I think it's essential! Storage is part of computing and I think young programmers can learn a lot from this.. keep up the good work 👍

dpgeorge:
@markshannon I checked that my patch in #265 works correctly with your tools/upload.py script (ie that even if there is a main script that runs forever, you can still upload a file using raw REPL).

damianburrin:
Hi Chap - has this been implemented into the live version @markshannon ?  Would be really useful to be able to log data directly to the Micro:bit rather than streaming it to an open logger..

Thanks
Damian

dpgeorge:
@damianburrin: no, it's not yet merged.

markshannon:
Squashed, rebased and waiting to be merged...

dpgeorge:
/source/microbit/main.cpp:40:9: error: 'hard_reboot' was not declared in this scope

dpgeorge:
Merged, yay!

I guess we need to tag the commit before this as 1.0, and now change the version string to 1.1?

dpgeorge:
Also, documentation and tutorial updates will be needed.

ntoll:
I'll make a start... ;-)

damianburrin:
Awesome.  Will try and develop a simple tutorial this weekend if that helps
Damian
On 17 May 2016 23:31, Damien George <notifications@github.com> wrote:Also, documentation and tutorial updates will be needed.

damianburrin:
Ace.
Well done chaps
On 17 May 2016 23:30, Damien George <notifications@github.com> wrote:Merged #251.

damianburrin:
I'm struggling a little bit with this @markshannon  i get an import error with the os library and attribute error for f.write()

Should this be available via the microbit.co.uk webversion yet or will it take a while to update?

I'll continue to monitor but i'm really excited about this and it will link in well with some serial coms and data looging project i've developped for school

ntoll:
@damianburrin you'll need to compile MicroPython for the micro:bit yourself. I'm not going to add it to Mu or uflash until we have comprehensive documentation and a tutorial (which I'll finish over this weekend). Getting such changes into the microbit.co.uk version of the editor is a pain due to BBC induced bureaucracy. Adding a file system is a large change that the BBC will want to test (potentially taking several weeks, depending upon their manpower). I'll reference the documentation in this ticket, so you can give feedback. :-)

damianburrin:
Ah ok thankyou. That's a shame - I was under the impression from a comment today
at some MicroBit training I delivered that the BBCs involvment was vertually
over now and most the development teams disbanded.  I'll hang on and look out
for your docs and try to compile but to be fair everything i'm doing id for use
in school so it needs to run on the online version so will have to wait and see.


Still great to see it being developed though.


Thank you

ntoll:
The BBC involvement is certainly not over. They're in control of and still run the microbit.co.uk website, and until some "foundation" or charitable trust is set up to continue the micro:bit project, they'll be (reluctantly, I guess) in the driving seat.

I can certainly assure you that the development team for the Python aspects of this project is going from strength to strength with new community members (such as yourself) contributing code, resources and documentation.

Thank YOU!

:-)

damianburrin:
That's great to hear, @ntoll  this is an awesome project.  I delivered some teacher training today via a CAS hub and it was buzzing,

I had staff with a huge range in skills developing in Python (and C++) and they were very keen to move away from block and see how they can take those next step.  I've been demoing wireless radio data links between Micro:bits an I see that could be coming to Python to soon which is another great tool. :-)

dpgeorge:
> I've been demoing wireless radio data links between Micro:bits an I see that could be coming to Python to soon which is another great tool. :-)

@damianburrin can you confirm that you are using C++ and the simple radio protocol from the DAL and not BLE for this?  I don't think any high-level languge yet supports simple radio (but Python will soon enough!).

damianburrin:
>  can you confirm that you are using C++ and the simple radio protocol from the DAL and not BLE for this?
@dpgeorge - Yes via the embed site.  This is my code if you're interested http://www.microbitsandbobs.co.uk/downloads/Shootout.cpp I'm sure it's not ace and is just based on the demo code thy show.

I've written a rough set of functions to drive a standard LCD screen as well but in Python (http://www.microbitsandbobs.co.uk/downloads/lcd3.00.py ) and working on a head tracker with some of my sixth form student as a side project in school.  I'm really enjoying teaching with the Micro:bit to my Y7 (and older) at the moment.  http://www.microbitsandbobs.co.uk/downloads/lcd3.00.py 

You guys are the best :-)

dpgeorge:
Thanks @damianburrin for the info!  It means we should try and make the Python radio compatible (over the air) with the DAL protocol.

Issue #222, opened 2016-02-26
We need a test suite


dpgeorge:
It goes without saying that we need a bunch of tests that make sure the microbit API remains stable and doesn't regress, especially with things like #221.

MicroPython upstream has a big test suit for Python features, so we don't need to duplicate that (or you can just run that test suite on the microbit, it mostly works).  We need tests for things like display rendering, accelerometer, gestures and music.

deshipu:
I wonder how you even test a hardware project such as this? Wire it up to a second microcontroller and verify that the outputs are what you expect?

JoeGlancy:
Is there an M0 emulator which might exist? a A universal test suite that
everyone could use might be a but complicated if it required an external
microcontroller, unless it was something commonplace like an Arduino which
a lot of people have.

dpgeorge:
There are various things you can do to test hardware: at the least you just want to verify that the functions can be called as specified by the API docs.  Then you can test some things like accelerometer and buttons by prompting the user.  You can also connect pins to other pins and test GPIO that way.  Connecting to another microcontroller is the only way to fully test everything, but is probably overkill.

JoeGlancy:
You could have a `selftest` module even, which does

dpgeorge:
A simple test suite (just one script at the moment) has been added.

PR #254, opened 2016-04-08
Add simple hardware test script and README.

ntoll:
Starts to address #222. Unfortunately, this first stab is not automatic - i.e. it needs a user of the device to interact with it to confirm things are working.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,9 @@
+Tests
+=====
+
+This directory contains script[s] that can be used to confirm various features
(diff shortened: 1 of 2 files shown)


whaleygeek:
If you have access to a Mega:Bit, we have a way to press buttons
(electrically) and monitor the LEDs (optically) automatically now. You
could build a more automated test harness using that as a basis. Servos
could be used to mostly test compass and accelerometer.

markshannon:
What is a "Mega:Bit"?

PR #255, opened 2016-04-11
In mphalport.{h,cpp},  Use "unsigned int" instead of "uint".

dpgeorge:
A recent change meant that uint was no longer being defined in some cases, and the build is now broken (at least on my system).  This change removes the dependency on the uint definition.


dpgeorge:
@ntoll @markshannon would you mind please merging this?

markshannon:
LGTM.

Issue #257, opened 2016-04-17
Button get_presses/reset_presses needs reworking


dpgeorge:
The idea of the `button.get_presses()` and `button.reset_presses()` methods is that you can get the number of queued presses and react on them (eg incr a counter).  But actually it's still possible to miss a press with these methods, namely:
```
i = button.get_presses()
# button is pressed here
button.reset_presses()
# i doesn't take into account the most recent press
```
To fix this requires a `button.get_and_reset_presses()` method.

I think it's bad if we let/encourage incorrect code like the above.  Instead we should provide proper atomic functions like `get_and_reset_presses` so people get used to the complexities of doing such things.

Gestures are also affected by the same thing.

markshannon:
The  `was_pressed` pressed method is the boolean version of `get_presses` and it resets itself. Perhaps `get_presses()` should do likewise and we can drop `reset_presses()`. Running totals can easily be obtained with `cnt += button.get_presses()`.
Will need the docs to updated, of course.

dpgeorge:
> Perhaps get_presses() should do likewise and we can drop reset_presses(). Running totals can easily be obtained with cnt += button.get_presses().

Yes, this seems like the most elegant way to do it.  Dropping the `reset_presses()` method makes the API simpler to use.  And I don't really see any need to have it a separate method.  You'd probably always call reset after get.

ntoll:
I like @markshannon's suggestion. Simple wins every time. ;-)

ntoll:
Fixed by #258

whaleygeek:
Thanks for breaking all my example programs :(

Real users are using this deployed code (even though you are not at 1.0).

ntoll:
@whaleygeek show me the code and I'll fix it.

markshannon:
@whaleygeek Sorry about that. Version 1.0 is coming very soon, so no more API changes thereafter.

Would you like to contribute some of your example programs? We could always do with more examples.

dpgeorge:
@whaleygeek unfortunately, that is the nature of progress :)

Where do you pull your source/binaries from?  Do you compile directly from this repo?

If needed we can add back a dummy, undocumented reset_presses() function.  That will un-break things.

markshannon:
I don't think adding an undocumented `reset_presses()` is a good idea.
Undocumented doesn't mean unsupported; it just means that we have a hole in the documentation.

dpgeorge:
[...]

PR #258, opened 2016-04-17
Rework button and accel events to remove reset method

dpgeorge:
Should address issue #257.

Please think a bit if this is indeed the right thing to do.


ntoll:
Tested and works for me. Nice an simple. :-)

docs/microbit_micropython_api.rst
```diff
@@ -46,10 +46,8 @@ These are both objects and have the following methods::
     # returns True or False to indicate if the button was pressed since the device
     # started or the last time this method was called.
     button.was_pressed()
-    # returns the running total of button presses.
+    # returns the running total of button presses, and resets this counter to zero
```
markshannon: "returns the total number of button presses since the device started or the last time this method was called." would follow the same template as for `was_pressed()` and is more explicit.

markshannon:
I should have noticed this was merged, before commenting.

PR #259, opened 2016-04-17
Master

Neon22:
changes to fire simulation example as per feedback.
Not sure if my git foo is going to work.


Neon22:
oh that doesn't look right :( only trying to get single file change

PR #260, opened 2016-04-18
Update to latest MicroPython core, v1.7-9.

dpgeorge:
Main changes are:

- nested str formatting with {}
- complete bignum support for negative args to and/or/xor
- fix for failed map reallocation
- improved float formatting, and fixes for buffer overflows
- inline asm funcs can take 4 args, and specify a return type


ntoll:
Great stuff. I believe we're at 1.0. :-)

PR #262, opened 2016-05-01
flame simulation

Neon22:
text adj


Neon22:
I just dont; get it - I unchecked everything except my changes.
My git-foo is so weak :(

Neon22:
this was from a completely barand new checkout - deleted everything and still I can't get it to work.
I think I need to take up vagrancy instead of this computing lark.. Its time to go...

dpgeorge:
Simplest thing would be to just: Copy your script out of the repo.  Delete the repo.  Clone it from scratch.  Copy back your script.  Add and commit it in a single commit.  Push.  Open PR.

Neon22:
that's exactly what I did. :( trying again
thanks

Neon22:
for the record - removing it doesn't delete any of the files in the local github folder :(
After manual delete - works as expected.

PR #263, opened 2016-05-01
flame sim example

Neon22:
demo example that simulates small flame in display - with masking for
shape


dpgeorge:
Great!  Thanks for persisting with it!

PR #265, opened 2016-05-03
Only execute the initial script if in friendly-REPL mode.

dpgeorge:
This allows you to enter raw REPL mode (ctrl-A at the REPL) and prevent
the microbit from executing any code when it does a soft reboot.

### DIFF
--- a/source/microbit/mprun.c
+++ b/source/microbit/mprun.c
@@ -105,12 +105,16 @@ void mp_run(void) {
     readline_init0();
     microbit_init();
 
-    if (APPENDED_SCRIPT->header[0] == 'M' && APPENDED_SCRIPT->header[1] == 'P') {
-        // run appended script
-        do_strn(APPENDED_SCRIPT->str, APPENDED_SCRIPT->len);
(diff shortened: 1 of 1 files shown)


markshannon:
I haven't tested this, but it looks to me as if it has the same problem as before.
Suppose we start with a script running in "friendly" repl mode. Crtl-C is hit and a soft reboot occurs, but remains in "friendly" repl mode. The embedded script is then restarted and we're back to where we started.

dpgeorge:
> Suppose we start with a script running in "friendly" repl mode. Crtl-C is hit and a soft reboot occurs, but remains in "friendly" repl mode.

Ctrl-C stops the script but does not do a soft reboot.  You then do ctrl-A to enter raw repl, then any soft reboot keeps it in raw repl without executing the script.

markshannon:
Sorry for the delay. I've mapped out the state machine and this change does do what I want :+1: 
Does this need a rebase before merging?

dpgeorge:
> I've mapped out the state machine and this change does do what I want

Great!

> Does this need a rebase before merging?

Nope, it's ready to merge.

markshannon:
Ok, merging.

Issue #264, opened 2016-05-03
Allow to configure range (and rate?) of accelerometer


dpgeorge:
The accelerometer defaults to a 2g max range, but can do up to 8g.  The DAL provides a function to get/set the range (and sample rate) of the accelerometer and we should probably expose this in Python.  Eg:

```
accelerometer.set_range(8)
accelerometer.get_range()
accelerometer.set_rate(50) # 50Hz sample rate
accelerometer.get_rate()
```

catchpolej:
I am right in thinking that a reading of 1024 using default settings is equal to 'g'?

dpgeorge:
> I am right in thinking that a reading of 1024 using default settings is equal to 'g'?

The return value is is milli-g's, so 1000 is equal to 'g'.

microbit-mark:
This has been requested again recently in micro:bit support ticket: 37999

It looks like this PR is still open https://github.com/bbcmicrobit/micropython/pull/266 
is this something that we can implement? It seems a useful feature and would be comparable to our MakeCode editor.

PR #266, opened 2016-05-04
Add acceleromter range and rate acccessors (in progress). 

carlosperate:
Didn't really get the chance last night, but I found some time this morning to play with this. I've encountered something a bit odd with my current implementation, so I'd though I'd create the PR for quick review (it is rather simple) in case I'm not setting something correctly.

When I set the range and rate it seems like it is heavily rounding up the final calculation.  So to set a sampling frequency of 50hz, I need to set the value 45 or 46, anything higher sets it to 100hz. Similar thing happens with the range. 

I'll see if I can replicate tonight the same behaviour using a simple Cpp program with just the DAL.


docs/accelerometer.rst
```diff
@@ -59,6 +59,22 @@ Functions
...
+    Set the accelerometer samping rate, in Hz.
+
+.. py:function:: get_range()
+
+    Get the accelerometer sensitivity range, in g's.
+
+.. py:function:: set_range()
+
+    Set the accelerometer sensitivity range, in Hz.
+
```
carlosperate: Should probably add more info about what changing the range actually does.

dpgeorge:
> When I set the range and rate it seems like it is heavily rounding up the final calculation.

I think the accelerometer can only do specific rates/periods: 1.25ms, 2.5ms, 5ms, 10ms, 20ms, 80ms, 160ms, 640ms.

Also it's restricted to 2, 4 or 8 g's.

carlosperate:
Yes, I meant when trying to set an exact accepted value, so let's say 20ms/50Hz, in that case it seems to actually set it to 10ms/100Hz, I have to do `microbit.accelerometer.set_rate(45)` for it to actually resolve to 50Hz. Same thing with the g sensitivity.

source/microbit/microbitaccelerometer.cpp
```diff
@@ -75,6 +75,32 @@ void microbit_accelerometer_init(void) {
...
+
+mp_obj_t microbit_accelerometer_get_range(mp_obj_t self_in) {
+    microbit_accelerometer_obj_t *self = (microbit_accelerometer_obj_t*)self_in;
+    return mp_obj_new_int(self->accelerometer->getRange());
+}
+MP_DEFINE_CONST_FUN_OBJ_1(microbit_accelerometer_get_range_obj, microbit_accelerometer_get_range);
+
+mp_obj_t microbit_accelerometer_set_range(mp_obj_t self_in, mp_int_t g) {
+    microbit_accelerometer_obj_t *self = (microbit_accelerometer_obj_t*)self_in;
+    return mp_obj_new_int(self->accelerometer->setRange(g));
```
carlosperate: Need to change this to  `return mp_const_none;`

carlosperate:
The using the DAL on a plain Cpp application sworks as expected on version 2. I tried also using version 1.4.15 and I've had some issues getting the program to create a `microbit-combined.hex` file. Had to convert my `main()` application to the old `app_main()`, and while it compiled fined, it only produced the `microbit.hex` file (and after loading that one nothing really happens). Is there any old documentation somewhere where I could check what exactly I might need to do?

carlosperate:
Okay, so I was finally able to spend a bit more time on this. So I went back and fixed the issue I had building the C++ program using the 1.4.15 version of the DAL, and as expected, that works well.

I looked at the data being sent to the DAL accelerometer instance and noticed that it was basically the user input `>> 1 | 1`, so I started looking around and noticed that's how the mp objects for integers are formed. Looks like the function arguments in the line before were passed as `mp_obj_t` instead of `mp_int_t`
```
mp_obj_t microbit_accelerometer_set_rate(mp_obj_t self_in, mp_int_t hz)
```

Is that always the case? I saw a few cases around using `mp_int_t` and I just assumed the data would be formated as integers. Is there any information online that describes some or any of this stuff? Googling I only found a rather small entry in the micropython wiki about creating new modules.


Anyway, it is working now, however because we are doing integer arithmetic with the frequency conversion to a period value in miliseconds, which is then converted in the DAL to microseconds, we do lose some accuracy. For the rate values configured in the accelerometer, we will read them as: 

```c
const MMA8653SampleRateConfig MMA8653SampleRate[MMA8653_SAMPLE_RATES] = {
    {1250,      0x00},       // Read as 1000Hz, supposed to be 800Hz
    {2500,      0x08},       // Read as 500Hz, supposed to be 400Hz
    {5000,      0x10},       // Read as 200Hz, correct
    {10000,     0x18},       // Read as 100Hz, correct
    {20000,     0x20},       // Read as 50Hz, correct
    {80000,     0x28},       // Read as 12Hz, supposed to be 12.5Hz
    {160000,    0x30},       // Read as 6Hz, supposed to be 6.25Hz
    {640000,    0x38}        // Read as 1Hz, supposed to be 1.5625Hz
};
```

Apart from that, if the user tries to set the rate to 0Hz, it will be configure to the 1250 us / 800Hz value.

carlosperate:
Apart from the extra info needed in the documentation, is the implementation correct based on what you had in mind with https://github.com/bbcmicrobit/micropython/issues/264?

markshannon:
OOI why is the API in terms of rate, rather than period?
Given that only a fixed set of rate/periods are available would it make sense to make a set of symbolic constants available?

carlosperate:
> OOI why is the API in terms of rate, rather than period?

I was implementing the suggested API from #264 , but period in ms or us would be easier to pass through.

> Given that only a fixed set of rate/periods are available would it make sense to make a set of symbolic constants available?

I'm not a big fan of having constants for these, but maybe we should return the configured value? So (using still rate) it would look like this:

```
> microbit.accelerometer.set_rate(10)
12
```

carlosperate:
@dpgeorge any feedback on this?

dpgeorge:
@carlosperate it's very clean.

But now that we have the discussion in #296 about setting/getting the I2C frequency, I realise that configuring the accelerometer is similar, and maybe it should use an init() method to set the rate and range?

carlosperate:
Well, the accelerometer object gets constructed by the DAL, so it would be a bit odd to have an init method if we are able to successfully execute any of the other methods without calling it first, no?

I guess this is already the case for UART, SPI, and now I2C, but those are internal peripherals which, generally speaking, are a bit more advanced and in most cases have to be "set up" before use. From my perspective the display, buttons, accelerometer, and compass are more like "pre-packaged" components, for which we are facilitating their use, but are ready out of the box. Having a init function suggests that they need to go through a process of initialisation, that we are completely bypassing in most situations, and I would personally find that slightly confusing.

On the other hand having that init would continue a general convention, which could be good. So I don't mind that much either way.

Which way should we go?

docs/accelerometer.rst
dpgeorge: Yes, include the possible values to set.

docs/accelerometer.rst
```diff
@@ -59,6 +59,22 @@ Functions
...
+.. py:function:: set_rate()
+
+    Set the accelerometer samping rate, in Hz.
+
+.. py:function:: get_range()
+
+    Get the accelerometer sensitivity range, in g's.
+
+.. py:function:: set_range()
```
dpgeorge, line 74: needs to be `set_range(value)`

docs/accelerometer.rst
```diff
@@ -59,6 +59,22 @@ Functions
     Return a tuple of the gesture history. The most recent is listed last.
     Also clears the gesture history before returning.
 
+.. py:function:: get_rate()
+
+    Get the accelerometer samping rate, in Hz.
+
+.. py:function:: set_rate()
```
dpgeorge, line 66: I think we should remove the get/set rate methods, there is no such functionality in MakeCode

carlosperate:
Rebased and updated, in the end I also removed `get_range` to save memory.

PR #272, opened 2016-05-15
Add os.uname to 0.9 branch to form 0.9 release

markshannon:


dpgeorge:
There is thing specific to uPy called "attribute tuple" which is an efficient version of a named tuple.  In particular an attrtuple can be created in ROM so that no RAM is needed.  An attrtuple is used for implementations of os.uname() in other ports of uPy.  See https://github.com/micropython/micropython/blob/master/stmhal/moduos.c#L57-L80

markshannon:
Thanks, that's the thing I wanted.
Are there any C API docs for uPy? I couldn't find any.

dpgeorge:
> Are there any C API docs for uPy?

Unfortunately not.

I can whip up a PR for the os.uname patch if you'd prefer.

markshannon:
I'll need to rebase the filesystem stuff on top of it, so it is easier if I do it.

markshannon:
I've used the `MP_DEFINE_ATTRTUPLE` as you suggested.

dpgeorge:
Looks good!  And works for me :)

One issue: os.uname().sysname and os.uname().nodename are "micro:bit", but sys.platform is "microbit".  It might be good to make these all the same.  But I don't think we can change sys.platform because people are probably already using it / relying on it.

markshannon:
Done

PR #276, opened 2016-05-17
In mphalport.{h,cpp},  Use "unsigned int" instead of "uint".

dpgeorge:
A recent change meant that uint was no longer being defined in some
cases.  This change removes the dependency on the uint definition.

[This patch was already applied to master but needs to be back ported to 0.9.]


markshannon:
OOI when isn't `uint` defined?
The change itself looks fine.

dpgeorge:
> OOI when isn't uint defined?

It's not a standard definition (although it is very common).  It's defined in inc/py/misc.h, but that's not always included.

PR #279, opened 2016-05-22
Filesystem docs

ntoll:
This PR is currently somewhere to hang a discussion. Currently only the "API" documentation is written. A tutorial is on its way.

Feedback most welcome.


docs/filesystem.rst
```diff
@@ -0,0 +1,227 @@
...
+    ['hello.py']
+    >>> os.remove('hello.py')
+    >>> os.listdir()
+    []
+
+.. py:function:: open(filename, mode='r')
+
+    Returns a file object representing the file named in the argument
+    ``filename``. The mode defaults to ``'r'`` which means open for reading in
+    text mode. The other common mode is ``'w'`` for writing (truncating the
```
carlosperate: truncating? I'm not sure I follow.

docs/filesystem.rst
```diff
@@ -0,0 +1,227 @@
+Local Persistent File System
+****************************
+
+It is useful to store data in a persistent manner so that it remains intact
+between restarts of the device. On traditional computers this is often achieved
+by a file system containing named files that contain raw data, and named
+directories that contain files. Python supports the various operations needed
```
carlosperate: the word contain is used 3 times in a row, maybe we can rephrase it a little?

docs/filesystem.rst
```diff
@@ -0,0 +1,227 @@
...
+
+    Removes (deletes) the file named in the argument ``filename``. If the file
+    does not exist an ``OSError`` exception will occur.
+
+.. py:function:: size(filename)
+
+    Returns the size, in bytes, of the file named in the argument ``filename``.
+    If the file does not exist an ``OSError`` exception will occur.
+
+.. py:function:: uname()
```
carlosperate: Perhaps there should be a different document for this function? Doesn't really fit with the filesystem information.

docs/filesystem.rst
```diff
@@ -0,0 +1,227 @@
...
+        already closed. Once the file is closed, any operation on the file
+        (e.g. reading or writing) will raise an exception.
+
+    .. py:method:: name()
+
+        Returns the name of the file the object represents. This will be the
+        same as the ``filename`` argument passed into the call to the ``open``
+        function that instantiated the object.
+
+    .. py:method:: read(size)
```
carlosperate: Would this crash is the size is larger than available RAM?

docs/filesystem.rst
```diff
@@ -0,0 +1,227 @@
...
+        if there are less than ``size`` characters remaining to be read from
+        the file.
+
+        If 0 characters are returned, and ``size`` was not 0, this indicates
+        end of file.
+
+    .. py:method:: readall()
+
+        Read all the characters from the file until EOF (end of file) is
+        reached and return them as a single string.
```
carlosperate: Would this start from the beginning of the file, or from the current position?

docs/filesystem.rst
ntoll: Good catch... update incoming.

docs/filesystem.rst
ntoll: This is directly copied from the CPython Python docs. I guess they're trying to say it overwrites existing data and truncates the file to just the length of the latest version before it gets closed. Does this make sense? I think we can make this clearer, so will re-write.

docs/filesystem.rst
ntoll: I'm not sure what it'd specifically do, but I can't imagine it being very good. ;-)

As far as I can tell, there's no upper limit on the size of an individual file. Perhaps there should be? @markshannon, do you have an opinion on this?

docs/filesystem.rst
ntoll: Current position. I'll add a clarification.

docs/filesystem.rst
ntoll: Well, given that these functions sit under the ``.. py:module:: os`` declaration in RST it's logically the right place. Totally get how this might be seen as an odd fit, but I'm not sure where else it would go.

Suggestions?

docs/filesystem.rst
carlosperate: That was more or less my assumption when I first read it, but because it doesn't clearly state that the file is overwritten I wasn't quite sure if that was part of the "truncation". Specially taking in consideration that there is no append operation I think that needs to be made a bit more clear.

docs/filesystem.rst
carlosperate: Of course it makes sense to have the file related functions with the file system documentation, but if I wasn't familiar with the docs and was looking for info on the `uname` function it'd be quite odd to find it here. Since separating the `os` functions is not a good solution either, I would suggest to create a new RST just for the `os` module and add a link to that here?

ntoll: I did think about writing it like this, but decided on putting things into a single file since all the file system related things would be together. Upon reflection I think I prefer your suggestion. I'll update and let's see how it reads.

ntoll:
OK... having played around with things a little I think I agree with @carlosperate and it's good to separate the `os` module out. Tutorial coming soon.

docs/filesystem.rst
```diff
@@ -0,0 +1,179 @@
...
+    Re-flashing the device will DESTROY YOUR DATA.
+
+    Since the file system is stored in the micro:bit's flash memory and
+    flashing the device rewrites all the available flash memory then all your
+    data will be lost if you flash your device.
+
+    However, if you switch your device off the data will remain intact until
+    you either delete it (see below) or re-flash the device.
+
+MicroPython provides a flat file system; i.e. there is no notion of a directory
```
dpgeorge: Can we reword this so it doesn't sound like all uPy implementations provide just a flat FS?  Eg something like "MicroPython on the micro:bit"?

docs/filesystem.rst
```diff
@@ -0,0 +1,179 @@
...
+hierarchy, the file system is just a list of named files. Reading and writing
+a file is achieved via the standard Python ``open`` function and the resulting
+file-like object (representing the file) of types ``TextIO`` or ``BytesIO``.
+Operations for working with files on the file system (for example, listing or
+deleting files) are contained within `the os module <os.html>`__.
```
dpgeorge: I think you should link to the os docs using a tag, not a html file name.

docs/filesystem.rst
```diff
@@ -0,0 +1,179 @@
...
+
+If a file ends in the ``.py`` file extension then it can be imported.
```
dpgeorge: Maybe: "... can be imported using 'import foo'."?

docs/filesystem.rst
dpgeorge: It should just raise a MemoryError, which is the right thing to do.  I don't think we should put arbitrary limits on anything.

docs/filesystem.rst
```diff
@@ -0,0 +1,179 @@
...
+        Read and return at most ``size`` characters from the file as a single
+        string. As a convenience, if ``size`` is unspecified or -1,
+        ``readall()`` is called. Fewer than ``size`` characters may be returned
+        if there are less than ``size`` characters remaining to be read from
+        the file.
+
+        If 0 characters are returned, and ``size`` was not 0, this indicates
+        end of file.
+
+    .. py:method:: readall()
```
dpgeorge: We should remove this function.  It's not needed, just use file.read() instead.

docs/filesystem.rst
```diff
@@ -0,0 +1,179 @@
...
+        Return ``True`` if the file supports writing. If ``False``, ``write()``
+        will raise ``OSError``.
+
+    .. py:method:: write(s)
+
+        Write the string ``s`` to the file and return the number of characters
+        written.
+
+
+.. py:class:: BytesIO
```
dpgeorge: Intead of writing basically the same thing for BytesIO as TextIO, why not combine the sections and just say at the start they work with either str or bytes objects?

docs/os.rst
```diff
@@ -0,0 +1,53 @@
...
+    * ``sysname`` - operating system name
+    * ``nodename`` - name of machine on network (implementation-defined)
+    * ``release`` - operating system release
+    * ``version`` - operating system version
+    * ``machine`` - hardware identifier
+
+
+.. note::
+
+    There is no underlying operating system in MicroPython. As a result the
```
dpgeorge: I don't think we should say this.  os.uname() does contain useful info, eg what version of uPy you are using.

docs/filesystem.rst
ntoll: I'll add something to this effect.

docs/filesystem.rst
ntoll: That's what I did originally, but they actually have different methods (`BytesIO` has `readinto` whereas `TextIO` does not). In the end it was confusing and clunky so I just went to simple and clean and separated the two classes out.

docs/os.rst
ntoll: Quite. I'll rewrite. I was trying to indicate that sysname and nodename are a bit redundant.

docs/filesystem.rst
ntoll: Having this method makes it consistent with CPython. But having one less method makes the API cleaner. Happy to document however it's implemented. Thoughts @markshannon..?

ntoll:
@dpgeorge thanks for the feedback - I've made as many of the changes as I can. Tutorial coming soon.

docs/filesystem.rst
dpgeorge: CPython does not actually have readall() for TextIO or BytesIO (try it!).

docs/filesystem.rst
dpgeorge: I see.  I didn't notice that TextIo and BytesIO are using different method tables.  They should use the same tables (Ie have exactly the same methods, using the same underlying code, to save code size; that's how it is in upstream uPy).  Then the docs are the same.

docs/filesystem.rst
ntoll: My god, you're right! OK... +1 on dropping the method (I guess in a different branch).

docs/filesystem.rst
ntoll: OK... this suggests to me that the filesystem related code introduced in c1039cd2419281a8e6bdfd27bc4ac3c64a3713f3 perhaps needs refactoring?

dpgeorge: See #280.

ntoll:
Folks, I've finally managed to finish the tutorial. All feedback most welcome.

docs/filesystem.rst
```diff
@@ -0,0 +1,130 @@
...
+    []
+
+.. py:function:: open(filename, mode='r')
+
+    Returns a file object representing the file named in the argument
+    ``filename``. The mode defaults to ``'r'`` which means open for reading in
+    text mode. The other common mode is ``'w'`` for writing (overwriting the
+    content of the file if it already exists). Two other modes are available
+    to be used in conjunction with the ones describes above: ``'t'`` means
+    text mode (for reading and writing strings) and ``'b'`` means binary mode
```
dpgeorge: I think you need to provide an example to show that t/b are *combined* with r/w, eg simply say: "so you can use, for example, 'rb' to read binary data from a file".

docs/tutorials/storage.rst
```diff
@@ -0,0 +1,231 @@
...
+    in an open state) you will be writing OVER the content of the file if it
+    already exists.
+
+    If you want to append data to a file you should first read it, store the
+    content somewhere, close it, append your data to the content and then open
+    it to write again with the revised content.
+
+    While this is the case in MicroPython, "normal" Python can open
+    files to write in "append" mode. That we can't do this on the micro:bit is
+    a result of the simple implementation of the file system.
```
dpgeorge: @markshannon how hard is it to support append mode?

markshannon: It is doable, but a little tricky and somewhat slow.
The file length will need to be modified in place (which will involve copying and erasing pages).

dpgeorge:
Looks good!

docs/tutorials/storage.rst
dpgeorge: Ok.  I thought it might be as simple as re-opening the file (and refilling caches, etc) like it was never closed from the last time it was written to.

ntoll:
OK... @dpgeorge added the example as you requested, rebased and ready to merge.

dpgeorge:
Hmm, @ntoll it seems yrou rebasing did not work.... try again?

ntoll:
@dpgeorge apologies, it helps to pull master before rebasing... ;-) (blame it on the jet lag).

dpgeorge:
Done.

PR #280, opened 2016-05-23
File simplification; and raise ValueError when file is closed

dpgeorge:
There are 2 patches: to make TextIO and BytesIO use the same method table, and to raise a ValueError when reading/writing a closed file.


ntoll:
Great stuff. Love @markshannon's slithy tests (that all pass). I'll update #279 as we discussed in terms of unifying the class docs.

PR #282, opened 2016-05-25
Fixed needle direction

pddring:
I'm not sure if I'm right here so please correct me if I'm wrong:

compass.heading() returns a number from 0 - 360 where 0 is N. I'm assuming 90 is E, 180 is S etc...

Without the change, the needle points NWW for 90 rather than E: the needle spins counter-clockwise rather than clockwise as the heading increases from 0 to 360.

See https://create.withcode.uk/python/55

### DIFF
--- a/examples/compass.py
+++ b/examples/compass.py
@@ -17,5 +17,5 @@ compass.calibrate()
 # Try to keep the needle pointed in (roughly) the correct direction
 while True:
     sleep(100)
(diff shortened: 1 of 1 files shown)


dpgeorge:
I'm pretty sure the example is correct, it works correctly for me (points roughly N all the time).

Probably the docs could be improved to make it clearer what the return value is.

Anyway, I'll close this because the example is correct.

PR #289, opened 2016-06-02
Display on/off

### LINKED ISSUE #274: Add display.disable() and display.enable()

komeP:
At PyCon 2016, we are converging on using on()/off()/is_active() to set and query the display state. Activate/deactivate are relatively complex words, and using a single function seems a lot like a property.

When the display is off, state is preserved and animation is paused. When the display is turned back on, the state is restored and animation is resumed from the point where it was paused. This seems a reasonable behavior, and is perhaps simpler than discarding the state.

If the display state is modified when off, the effect of the modifications will be visible when the display is turned back on. This seems a reasonable behavior, and is simpler to implement than raising exceptions. Note, the display will flash the maximum brightness LEDs from its static state briefly before resuming the animation loop.

Also added documentation of the PWM loop as I understand it.

Resolves #274


source/microbit/help.c
```diff
@@ -105,6 +105,9 @@ STATIC const mp_doc_t help_table_instances[] = {
     {&microbit_display_clear_obj, "Use clear() to clear micro:bit's display.\n"},
     {&microbit_display_get_pixel_obj, "Use get_pixel(x, y) to return the display's brightness at LED pixel (x,y).\nBrightness can be from 0 (LED is off) to 9 (maximum LED brightness).\n"},
     {&microbit_display_set_pixel_obj, "Use set_pixel(x, y, b) to set the display at LED pixel (x,y) to brightness 'b'\nwhich can be set between 0 (off) to 9 (full brightness).\n"},
+    {&microbit_display_on_obj, "Use on() to turn on the display.\n"},
+    {&microbit_display_off_obj, "Use off() to turn off the display.\n"},
+    {&microbit_display_is_active_obj, "Use is_active() to query if the micro:but's display is on (True) or off (False).\n"},
```
carlosperate: micro:bit instead of micro:but :)

markshannon:
LGTM.
Could you squash and rebase this please.

markshannon:
LGTM

PR #292, opened 2016-06-02
New examples, bugfixes, docs

c-mart:
First PR to this project: example programs that 11-year-old developers will hopefully comprehend, and some doc tweaks, also 2-player space shooter game using radio module

### DIFF
--- a/docs/tutorials/hello.rst
+++ b/docs/tutorials/hello.rst
@@ -64,3 +64,9 @@ change the scroll method's argument.
     in a way that MicroPython can't understand. Check you're not missing any
     special characters like ``"`` or ``:``. It's like putting. a full stop in
     the middle of a sentence. It's hard to understand exactly what you mean.
+
+    Your microbit may stop responding: you cannot flash new code to it or
+    enter commands into the REPL. If this happens, try power cycling it. That
+    is, unplug the USB cable (and battery cable if it's connected), then plug
+    the cable back in again. You may also need to quit and re-start your code
+    editor application.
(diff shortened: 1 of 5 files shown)


docs/tutorials/hello.rst
```diff
@@ -64,3 +64,9 @@ change the scroll method's argument.
     in a way that MicroPython can't understand. Check you're not missing any
     special characters like ``"`` or ``:``. It's like putting. a full stop in
     the middle of a sentence. It's hard to understand exactly what you mean.
+
+    Your microbit may stop responding: you cannot flash new code to it or
```
dpgeorge: Did you experience such behaviour?  It should be that you can always flash new code because the chip that does the flashing is independent from the nRF chip that runs the code.

c-mart: I did yesterday, @ntoll saw it. The microbit stopped accepting new flashes after I held down the reset button for a few seconds when my program wasn't behaving as expected.

Perhaps we should add something to the tutorial about what exactly the reset button does and when you should push it.

carlosperate: > after I held down the reset button for a few seconds 
Did you just hold down the reset button, or did it while it was power cycling? https://www.mbed.com/en/development/hardware/prototyping-production/daplink/daplink-on-kl26z/

c-mart: I just held the reset button and did not power cycle, and this caused the lock-up.

dpgeorge:
Since the radio module is still in alpha (and not even merged to master) we can't yet merge the space-shooter example (even if it is very cool!).  The other things are fine.  So please separate out the space-shooter code to a different PR for later.

c-mart:
Sure, done. I'll try to make future PRs more granular.

dpgeorge:
@c-mart almost there: can you please rebase against the current master, and then squish some of the commits so the space-shooter example is completely wiped from the history?  Thanks!

c-mart:
Done.

dpgeorge:
Thank you!

PR #293, opened 2016-06-02
Use display.show instead of display.animate

tcervo:
The analog_watch example was using the outdated display.animate, so no longer worked. Change to display.show so that it works again.


dpgeorge:
Thanks!

PR #297, opened 2016-06-03
Add a 20-sided die roller to the examples

tcervo:
A simple D20 die roller that displays the number rolled, with special displays (and optional sound) if either a 1 or a 20 is rolled.


dpgeorge:
Thanks!  Merged in 085aeb1fdefe4b80eca84f6e44a0fd5282ff42af (with file renamed to `examples/die20.py`).

PR #300, opened 2016-06-04
Created at PyCon2016, I donate this to public domain!

VivianePons:


examples/tetris.py
```diff
@@ -0,0 +1,84 @@
+from microbit import *
+import random
+
+class simpleTetris():
```
markshannon: Capital `S` please.

examples/tetris.py
```diff
@@ -0,0 +1,84 @@
...
+    
+    def __init__(self):
+        self.screen = [[0 for i in range(5)] for j in range(5)]
```
markshannon: You could make this an `Image` directly, which would simplify the rendering.

VivianePons:
I followed the advice and changed the SimpleTetris to use Image directly. The reason I hadn't done so is because it made SimpleTetris independent of the API and so easier to test for me. Anyway, I changed it because I guess it's better not to copy the data structure on a such a small device.

BUT, I have a question. Do you have a way to do some testing on the program without pushing to the microbit? Because I got tired at reading the microbit scrolling error message even on such a small program...

dpgeorge:
> Do you have a way to do some testing on the program without pushing to the microbit?

Try using tools/pyboard.py to quickly upload and run your script:
```
$ ./tools/pyboard.py /dev/ttyUSB0 examples/tetris.py
```
(replace ttyUSB with your device)

pddring:
Just found this tetris game and love it - nicely done.

In terms of testing without pushing to the microbit, I've been working on a microbit simulator to make it easier for students / teachers to debug code in a browser. Here it is running your tetris game: https://create.withcode.uk/python/Nb

dpgeorge:
This is great, but I'm not sure it can be merged as-is due to the Tetris copyright.  Probably renaming "Tetris" in the code would be enough to overcome that problem.

dpgeorge:
I've renamed it to FallingBlocks and merged it in 77c2e2ad8d8340f83a809166997109421d3c7127.

Thanks for the contribution, it's a really fun little game!

PR #303, opened 2016-06-04
cleaning up the compass / direction docs

octaflop:
Updated the compass and direction docs so calibration steps are a bit more clear. Also updated links to explain floor and modulo operations so students can use external links to research these operations themselves.


dpgeorge:
Thanks!

Issue #288, opened 2016-05-30
Setting Pull-Down on io pins !?


cefn:
There seems to be a software capability to set pull-up, pull-down or pull-none on digital pins...
http://lancaster-university.github.io/microbit-docs/ubit/io/#setpull
...as also explored in this thread...
https://mail.python.org/pipermail/microbit/2016-May/001100.html

The micropython version I have access to through microbit.co.uk seems to default to pull-down whenever I make a digital_read() call which confused and confounded me for some time. 

I was searching explicitly for some indication there was pull-down capability used, and the default choice doesn't seem to be documented anywhere. The absence of a pull-mode API signature anywhere in python made me assume no pull behaviour was employed, and made me check and re-check the board I soldered together as I couldn't fathom where the short was :)

Is default pull-down as expected? Is it in your control or set by the DAL? Certainly seems to violate the principle of least surprise for me, and apparently the authors of these platforms...
- https://learn.sparkfun.com/tutorials/raspberry-gpio/python-rpigpio-api
- https://www.arduino.cc/en/Reference/PinMode

dpgeorge:
AFAIK we are just following how the other editors work (TouchDevelop,
etc).  It's the DAL that provides this behaviour of default pull-down.  I
don't know why it defaults to pull down but that's the behaviour that
people now expect (rightly or wrongly).

I think we should provide a function to change the pull mode.  And I guess
we need to document that it's pull down by default.

markshannon:
@cefn What would you consider to be the "obvious" default for the pull resistor?

cefn:
Sadly I agree with @dpgeorge that the micro:bit platform should be consistent across languages, so if all other languages have pull-down as default, then that forces your hand.

However, given a time machine back to the meeting where that was decided, I would say pull-none is the obvious engineering and educational STEM default, as chosen by both Arduino and Raspberry PI, with pull being a concept which is optionally introduced, not forced into the frame and having to work around. That's because...
* A voltage divider is a core concept which is deeply confused by having an invisible resistor forced into your schematic
* Consistency with Analog is better achieved by pull-none.
* The buttons have pull-up by default, confusing things even further, and it is easy to need to wire in both given the screen leaves so few IO pins.
* Given no pull API has made it into any of the downstream languages, Pull-none can be compensated for in a circuit by adding a resistor, but pull-down cannot
[...]

PR #304, opened 2016-06-05
add set_pull method to pin class

johndela1:
This change adds a set_pull function to control pull resistors.

see: https://github.com/bbcmicrobit/micropython/issues/288


dpgeorge:
DId you test that this works?  I think the DAL will override the setting once you call read_digital.

johndela1:
This is the behavior:
>>> pin1.set_pull(pin1.PULL_UP)  
>>> pin1.read_digital()        
1
>>> pin1.read_digital()
1
>>> pin1.set_pull(pin1.PULL_DOWN)
>>> pin1.read_digital()          
0
>>> pin1.read_digital()
0

now if you call read_analog then call read_digital it will set it to PULL_DOWN

with this code:
```

 // Move into a Digital input state if necessary.                                           
    if (!(status & IO_STATUS_DIGITAL_IN)){
        disconnect();
        pin = new DigitalIn(name,PullDown);
        status |= IO_STATUS_DIGITAL_IN;
    }
```

markshannon:
We now have a `set_pull` method on pins. 
@johndela1 thanks for your efforts

Issue #245, opened 2016-03-26
Create radio module based on DAL implementation


ntoll:
What the title says. This should be a separate module so the radio feature is not automatically available to kids (they have to explicitly import it).

damianburrin:
What us the concern @ntoll ?

ntoll:
The BBC have a concern surrounding child protection. I believe this shouldn't stop us implementing this feature since I believe it spurious.h

damianburrin:
>The BBC have a concern surrounding child protection
Oh ok.  Seems a bit over the top to me. You'd need to have a program instalked to recieve any form of message and then from fairly close. 
I hope to be able to use this feature in python rather than C.  Ive got a neat idea of a classroom 'Exit Pass' 
On 22 May 2016 19:45, Nicholas Tollervey <notifications@github.com> wrote:The BBC have a concern surrounding child protection. I believe this shouldn't stop us implementing this feature since I believe it spurious.h

ntoll:
Exactly... OTT.

It's coming.

markshannon:
We have a radio module now.

PR #305, opened 2016-06-10
Very rough first draft of radio module related documentation.

ntoll:
**READY TO PROOF READ / MERGE**

A "straw man" for the new `radio` module. Please annotate, comment and correct. Some of the details may need checking since this was written on a train so some of the specifications about the capabilities of the radio were from memory. 

I'm not precious about any of the text I've written, so please be merciless. ;-)

Once we're happy with the API we can massage @dpgeorge's code found in #245.


docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+.. py:function:: config(length=32, queue=3, channel=7, power=0, address=b'uBit\x00', data_rate=radio.MODE_1MHZ)
+
+    Configures the various settings relating to the radio. The specified
+    default values are sensible.
+
+    The ``length`` defines the size, in bytes, of a message sent via the radio.
+    It can be up to 256 (????? CHECK THIS from memory) bytes long.
+
+    The ``queue`` specifies the number of messages that can be stored on the
+    incoming message queue.
```
carlosperate: Is this a circular queue? We should probably indicate what happens when receiving more messages than there are spaces left.

docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+
+    The ``channel`` can be an integer value between 0 and 100 (inclusive) that
+    defines an arbitrary "channel" to which the radio is tuned. Messages will
+    be sent via this channel and only messages received via this channel will
+    be put onto the incoming message queue. Each step is 1MHz wide, based at
+    2400MHz.
+
+    The ``power`` is an integer value between 0 and 7 to indicate the strength
```
carlosperate: Looks like we have 8 power values, so since the previous range indicated "100" was inclusive maybe we should do the same thing here with "7".

docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+    Configures the various settings relating to the radio. The specified
+    default values are sensible.
+
+    The ``length`` defines the size, in bytes, of a message sent via the radio.
+    It can be up to 256 (????? CHECK THIS from memory) bytes long.
+
+    The ``queue`` specifies the number of messages that can be stored on the
+    incoming message queue.
+
+    The ``channel`` can be an integer value between 0 and 100 (inclusive) that
```
carlosperate: I might be wrong about the following comment, since English is not my first language, but `between 0 and 100 (inclusive)` to me sounds like it includes 100 but not 0, is it more clear to say "both inclusive"? or maybe `from 0 to 100 (inclusive)`.

docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+    2400MHz.
+
+    The ``power`` is an integer value between 0 and 7 to indicate the strength
+    of signal used when broadcasting a message. The higher the value the
+    stronger the signal, but the more power is consumed by the device.
+
+    The ``address`` is an arbitrary name, expressed as bytes, that's used to
+    filter incoming packets at the hardware level, keeping only those that
+    match the address you set. The default used by other micro:bit related
+    platforms is ``b'uBit\x00'``.
```
carlosperate: It looks like the base and prefix are combined here, maybe there should be more info about the restrictions around this value. At the very least indicate that it should be 40 bits / 5 bytes.

docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+
+    The ``data_rate`` indicates the speed at which transmission (???) takes
+    place. Can be one of: MODE_1MHZ, MODE_???? and MODE_???.
```
carlosperate: Looks like we have `RADIO_MODE_MODE_Nrf_1Mbit`, `RADIO_MODE_MODE_Nrf_2Mbit`, and `RADIO_MODE_MODE_Nrf_250Kbit`.

docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+
+    If ``config`` is not called then the defaults described above are assumed.
+    Since these are named arguments with sane defaults, re-setting is as simple
+    as calling ``config()``.
```
carlosperate: There should be a discussion about this in #306, but whatever the result (even if there is no change) resulting from that, we should probably indicate if something like `config(power=3)` would cause all the other settings to revert to their default values.

docs/radio.rst
```diff
@@ -0,0 +1,67 @@
...
+=========
+
+.. py:function:: send(message)
+
+    Sends a message, expressed as a string.
+
+.. py:function:: receive()
+
+    Receive the next incoming message on the message queue. Returns ``None`` if
+    there are no pending messages.
```
dpgeorge: We should specify the type of the returned value, ie bytes.  It should be bytes because that's what goes through the air.

ntoll: @dpgeorge I agree that we should be explicit - an oversight on my part. 

Folks, I'm going to play devil's advocate here: if I were an 11yo the concept of a string of characters is understandable but the abstraction of such strings encoded as bytes is perhaps a level of abstraction too far *at this stage* (it's definitely something they should know about). Also, as a programmer I might want to both send and receive bytes.

I'm trying to find a way to allow both the naive 11yo sending/receiving of a message (as a student, I can `send("a string")` so it's probably reasonable to expect to receive messages as strings) and the correct / more powerful sending and receiving of bytes.

I wonder if we could wrap byte related functions with string related ones as a convenience for kids? For example:

```
$ radio.send("Hello, World")
$ radio.receive()
'Hello, back!'
$ radio.send(b'bytes')
$ radio.recv_bytes()
b'ack'
```

Note the *obvious* (to a naive user) names for sending and receiving strings. The `send` method is overloaded to automatically decode the string to bytes if a string is passed in. The `recv_bytes` method returns bytes and, since this is for "advanced" users, we signal it as such by truncating to "recv" and mentioning "bytes". under the hood, the `receive` method calls `recv_bytes` and attempts to encode the result as a string (I guess raising a `UnicodeEncodeError` if it fails).

Thoughts..?

dpgeorge: Note that to communicate with devices using DAL-based radio you need to send a few special bytes at the start of the message, like protocol number and version.  These are specific to how the DAL works and have nothing to do with any other nRF device or on-air protocol, so we shouldn't bake them into our implementation.  Rather we should let users just send whatever bytes they like, and they can send the special prefix if they want to talk to other DAL-based devices.  Any data coming from a DAL device will be bytes, not a string.

Also note that I2C, SPI and UART classes all return bytes objects.  But I guess that radio will find more use with younger users than I2C/SPI/UART and so would benefit from a friendlier API.

markshannon:
See my comment about the API w.r.t bytes/strings https://github.com/bbcmicrobit/micropython/pull/283#issuecomment-227326721

dpgeorge:
> See my comment about the API w.r.t bytes/strings #283 (comment)
> ...
> I send a string, you recv a string. I send bytes, you receive bytes. Python is a dynamically typed language after all.
> ...

At the minimum we should provide a "raw" API whereby it's just bytes sent/received over the air, with no headers or anything added.  Then people can use that to implement things however they like.  On top of that we can, if needed, add a "friendly" API which does some pickling/unpickling of the data (eg using repr/eval) so that you can send typed data.

whaleygeek:
The API offered up by m.pxt.io is fine for all purposes I have seen and
implemented to date - I presume most of this is just built on top of the
DAL and the underlying nordic gazelle layer.

The key feature is the groupid byte, that 'feels' like tuning to a channel,
and makes it easy to implement independently working solutions in the same
room. RSSI measurements, setting the power level, transmitting a byte, a
string, or a keyed value, adding in an optional send timestamp and an
optional micro:bit serial number. Everything is broadcast (or simulated
multicast if you take the group byte into account automatically as PXT
does). Leave everything else up to the user to innovate themselves - this
seems fine to me.

Security for example - I don't see that we need any. Let the inventors of
the future realise the problems of non secure communications by
experiencing it first hand, and coming up with their own solutions to that
learning opportunity. We don't want to pre-join too many of the dots for
them.

markshannon:
@dpgeorge What does adding a "raw" protocol gain? If you want to send bytes, just send them. 
Anyone can still write their own protocols and encoding on top of that.

The radio protocol already add channels, headers, data-whitening, lengths and check sums. So there really isn't such thing as raw data.

dpgeorge:
> What does adding a "raw" protocol gain?

So you can communicate simply with non-Python code.  The way I implemented it it already does "raw" bytes.  Adding typed data is an extra layer on top of that.

If we eventually allow to control all the radio parameters then raw bytes are the only thing that makes sense to send.

markshannon:
It doesn't have to be an extra layer. Bytes is merely one of many possible types that can be transmitted.
We can add the extra types alongside the bytes protocol, rather than on top of it. 
There are 8 logical addresses. We can use one for bytes (as we do now) and the other 7 can be used for other types.

ntoll:
Thanks for the feedback so far. I've made some revisions in light of the conversation above.

@whaleygeek - I agree with what you're saying... leave it to the user to discover, explore and play. The most important thing we can do is make it easy and facilitate such playful exploration via a simple and obvious API.

I think it important that the API fulfil two functions (IMHO in order of importance):

* It makes it as easy/simple/fun as possible for an 11yo kid to do wireless communication.
* It is powerful/flexible/configurable enough for advanced users to do amazing things.

@dpgeorge @markshannon to fulfil the second requirement we should expose all the various configuration options (as I suggest with the `config` method) and allow access to the "raw bytes" level of sending a message. Given these two capabilities it should be possible for anyone to make full use of the device's radio capabilities.

Regarding the first requirement: it should be as simple as possible with the potential for playful exploration built in. From my point of view that simply means sending and receiving strings. Such messages would suffice for most basic use-cases. If an incoming message cannot be converted to a string it should be ignored (although it should be possible to override this behaviour) . It also means sending a message is three lines of code:

```
import radio

radio.on()
radio.send("Hello, World!")
```

Each line is easy to understand and tells an obvious part of the "story of using the radio" in a learner's mind.

Why not any other data types? 

Because realising why we have different types and that they can be represented in different ways is something for kids to learn about, and this module is a good way to introduce some aspects of this area to them. It'll mean they encounter, first hand, the difference between `"2"` and `2`.

Just using strings for the high level methods keeps the API very very simple. 

Not doing anything else would also provide an excellent opportunity for a lesson in data serialisation (e.g. via `repr`) and a cool warning for why `eval` has problems (hint: `radio.send("from microbit import panic; panic()")`). 

Finally, limiting messages to strings is both conceptually easy for a non-technical 11yo kid to understand (it's like writing a letter that's sent via the radio) yet limited enough that they'll soon realise why it's interesting to drop to the "raw bytes" layer to have more control over the messages being sent.

Does this make sense?

docs/radio.rst
```diff
@@ -0,0 +1,79 @@
...
+    ``fail`` argument is ``False`` (the default) such messages and associated
+    exceptions are silently ignored.
+
+.. py:function:: config(length=32, queue=3, channel=7, power=0, address=b'uBit\x00', data_rate=radio.RADIO_MODE_MODE_Nrf_1Mbit)
+
+    Configures the various settings relating to the radio. The specified
+    default values are sensible.
+
+    The ``length`` defines the size, in bytes, of a message sent via the radio.
+    It can be up to 256 (????? CHECK THIS from memory) bytes long.
```
ntoll: Can someone confirm what this upper limit is? I remember reading it somewhere but can't, for the life of me, find where that's referenced.

deshipu:
> If an incoming message cannot be converted to a string it should be ignored (although it should be possible to override this behaviour)

Errors should never pass silently.
Unless explicitly silenced.

examples/radio.py
```diff
@@ -0,0 +1,4 @@
+# To contain code that demonstrates the radio module.
+import radio
+
+radio.send("Hello, World!")
```
dpgeorge: radio.on() first?

docs/radio.rst
dpgeorge: I think it's 251 (254 - 3 bytes for S0, LENGTH and S1 preamble).

docs/radio.rst
```diff
@@ -0,0 +1,79 @@
...
+    Receive the next incoming message on the message queue. Returns ``None`` if
+    there are no pending messages. Messages are returned as bytes.
+
+.. py:function:: receive(fail=False)
+
+    Works in exactly the same way as ``recv_bytes`` but messages are expressed
+    as strings. If the optional ``fail`` argument is ``True``, a ``ValueError``
+    is raised if the incoming message cannot be converted to a string. If the
+    ``fail`` argument is ``False`` (the default) such messages and associated
+    exceptions are silently ignored.
```
dpgeorge: I think it's simpler to not have the fail argument at all and just always raise a ValueError.  A "try:except:" block can be used if the user wants to protect against such failures.

dpgeorge: Maybe we should define/say that `receive()` is equivalent to `str(recv_bytes(), 'utf8')`?

docs/radio.rst
```diff
@@ -0,0 +1,79 @@
...
+networks.
+
+Functions
+=========
+
+.. py:function:: send(message)
+
+    Sends a message, expressed as a string or bytes.
+
+.. py:function:: recv_bytes()
```
dpgeorge: Maybe `receive_bytes()`?

docs/radio.rst
```diff
@@ -0,0 +1,79 @@
...
+
+The ``radio`` module allows devices to work together via simple wireless
+networks.
+
+Functions
+=========
+
+.. py:function:: send(message)
+
+    Sends a message, expressed as a string or bytes.
```
dpgeorge: Maybe define/say that `send(s)` is equivalent to `send(bytes(s, 'utf8'))`?

docs/radio.rst
```diff
@@ -0,0 +1,79 @@
...
+
+    The ``channel`` can be an integer value from 0 to 100 (inclusive) that
+    defines an arbitrary "channel" to which the radio is tuned. Messages will
+    be sent via this channel and only messages received via this channel will
+    be put onto the incoming message queue. Each step is 1MHz wide, based at
+    2400MHz.
+
+    The ``power`` is an integer value from 0 to 7 (inclusive) to indicate the
+    strength of signal used when broadcasting a message. The higher the value
+    the stronger the signal, but the more power is consumed by the device.
```
dpgeorge: It's expressed as dBm in the hardware (-30 dBm to +4 dBm).  Shall we stick to physical units of dBm (and choose the supported value that's closest to the requested one) or use "simple" numbers from 0 to 7?  Note that if we ever support RSSI (received signal strength indicator) then that value will be in dBm.

ntoll:
@deshipu you're right. I wasn't thinking things through (my head was in the "just make it work for the kids no matter what" space). Upon re-reading it's painfully obvious that errors MUST be reported and, to address @dpgeorge's point, it's a good opportunity to introduce the notion of `try:except:finally:` blocks.

I'll update the docs in light of the above and other recent documents.

docs/radio.rst
ntoll: The important information in an 11yo's mind is that there are 8 potential settings for strength - the higher the number, the higher the strength. If we do support RSSI then I don't see a problem with the return value measured in dBm. However, I believe it important that we mention in the documentation how the simple numbering system translates to dBm.

ntoll:
OK... it feels like we're getting there. I'm certainly happy with the way the API is shaping up - it's certainly feels simple enough to get an 11yo sending wireless messages while still providing enough configuration and access to "raw bytes" to facilitate more complicated forms of communication.

Thoughts..?

markshannon:
I really don't think that we need to abandon the "everything is an object" paradigm of Python that we use throughout the rest of the API, just for the radio module. 
Everything in computing is composed of many layers from objects all the way down to transistors (and lower than that). We hide that.
Why say that the "bytes" layer in the radio module is special and must be exposed?

I'm sure that it is good for kids to understand that everything is composed of bytes. 
But, I don't want the radio module to be unusable unless they have that understanding.
IMO, the audio module is probably a better place for that, as it must necessarily handle samples.

For interoperability with DAL based implementations is important that we can send and receive bytes, but strings and ints are more useful objects.

Much of the rest of the API is polymorphic, it keeps the API small and memorable.
The radio module should be the same.

docs/radio.rst
```diff
@@ -0,0 +1,81 @@
...
+
+The ``radio`` module allows devices to work together via simple wireless
+networks.
+
+Functions
+=========
+
+.. py:function:: send(message)
+
+    Sends a message, expressed as a string or bytes. Sending a string is
```
markshannon: I don't like "expressed as a string or bytes", I would prefer "which can be a string or bytes object". The message (in the context of the API) is an object.

dpgeorge:
> I really don't think that we need to abandon the "everything is an object" paradigm of Python that we use throughout the rest of the API, just for the radio module.

Most (all?) the other communication entities in the API use bytes objects exclusively: uart, spi and i2c.  It should be the same for radio.

You might want to add a layer to the uart/spi/i2c interfaces to send and receive arbitrary objects (eg i2c.send(42)) but first and foremost you should support transfer of raw bytes because that's how it works on the wire, and allows you to communicate with arbitrary devices.

docs/radio.rst
dpgeorge: > The message (in the context of the API) is an object.

The "message" should be a string of bytes, not an object.  The ability to send a string (unicode) is a convenience thing.

docs/radio.rst
dpgeorge: With dBm a larger number does correspond to a greater transmit strength.  I think setting the power is an advanced option and won't be used much by the kids, so the value should be in "native" dBm, saving the need to add further documentation about the translation from 0-7 to dBm values.

markshannon:
At no point have I suggested we don't support sending and receiving bytes. I just don't see why we need to make it unnecessarily difficult to send other objects. 
My proposed interface has a strict super set of the capabilities of the one currently proposed, yet is more concise.
It is also (IMO) easier to understand and explain.
On 25 Jun 2016 12:51 p.m., Damien George <notifications@github.com> wrote:
I really don't think that we need to abandon the "everything is an object" paradigm of Python that we use throughout the rest of the API, just for the radio module.


Most (all?) the other communication entities in the API use bytes objects exclusively: uart, spi and i2c.  It should be the same for radio.

You might want to add a layer to the uart/spi/i2c interfaces to send and receive arbitrary objects (eg i2c.send(42)) but first and foremost you should support transfer of raw bytes because that's how it works on the wire, and allows you to communicate with arbitrary devices.

dpgeorge:
> My proposed interface has a strict super set of the capabilities of the one currently proposed, yet is more concise.

I disagree with your proposal.  I think that bytes should be special and at a lower layer, and typed data/objects at a higher level (and typed bytes objects can also exist at this higher level), building on the low-level bytes send/recv functionality.  In particular you can choose to receive anything as bytes, even if it was sent as typed data (then you'd see the encoding scheme).

The logical address should be exposed to the user so they can write protocols that are compatible with C/C++ code that uses logical addresses in an arbitrary way.

So I would propose something like:
```py
# raw bytes protocol
radio.send_bytes(bytes_obj)
bytes_obj = radio.receive_bytes()
# higher-level object protocol
radio.send(obj) # equivalent to radio.send_bytes(bytes(repr(obj), 'utf8'))
obj = radio.receive() # equivalent to obj = eval(str(radio.receive_bytes(), 'utf8'))
```

ntoll:
In the most recent push I renamed ``RADIO_MODE_MODE_Nrf_1Mbit`` -> ``RADIO_MODE_Nrf_1Mbit`` (too many MODEs).

Also, @dpgeorge I think kids may use the power settings. Evidence: a teacher was recently telling me how interesting it would be to measure the range of the devices given a remote-controlled robot project they had in mind.

markshannon:
I'm happy with the two level protocol, with "raw" and "object" levels.
I'd like the "object" level to be the obviously "preferred" API in any documentation and to be able handle bytes, numbers, booleans, and (unicode) strings.

If we are to have two protocols, then I would like the low level protocol not to do any allocation.
`receive_bytes` should take a writeable buffer and return an integer showing the length of the message (returning -1 for nothing received).
`def receive(buf: buffer)->int`.

Not doing any allocation at the lower level would make the higher level protocol more efficient (although we could do that by exposing an allocation-free method in C).
It would also make it possible to receive data at a higher rate, e.g. for audio.

markshannon:
Actually, now I think about it, `receive` would read its data directly from the radio buffer. So adding the buffer parameter to `receive_bytes` wouldn't make any difference to the performance of `receive`.

I still think an allocation-free API would be better for `receive_bytes`.

deshipu:
> If we are to have two protocols, then I would like the low level protocol not to do any allocation.
> receive_bytes should take a writeable buffer and return an integer showing the length of the message (returning -1 for nothing received).
> def receive(buf: buffer)->int.

What should happen when the buffer is not large enough to contain the received message?

markshannon:
@deshipu It should raise. You already have to select a maximum message size in `config`, so it adds no new difficulties.

ntoll:
So, I'm happy with the current version of the API as it stands in this PR. 

With my teacher's hat on I like the fact that the simplest possible way of conceptualising a message (you send/receive text) would make sense to an 11yo. 

I also like that the send/receive bytes is low level enough that you could do all sorts of interesting things (such as communicating with non-micro:bit devices and/or other arbitrary communication). 

Finally, I like how the limitations of the simple text based send/receive based API provide enough functionality that kids can get started, but are limited enough to provide learning opportunities: for example, it should be relatively easy to write custom functions to send/receive arbitrary objects at the Python level, for educational purposes).

I'd like a radio module in the code base and documented with a tutorial by EuroPython for reasons that will become clear *at* EuroPython. ;-)

dpgeorge:
> If we are to have two protocols, then I would like the low level protocol not to do any allocation.

+1

> I'd like the "object" level to be the obviously "preferred" API in any documentation and to be able handle bytes, numbers, booleans, and (unicode) strings.

Yes, agree.  For now we can just handle (unicode) strings (and maybe bytes?) and the implementation should be something like: first byte on the air is the object type, remaining bytes are the data.  That allows for extensibility.

One remaining thing: we should make things compatible with DAL/pxt at some level.  Pxt has methods to "send number", "send string", "receive number" and "receive string" (ie you've got to know the type you're expecting).  I don't know exactly how their protocol works, and how much it's subject to change, but it would be nice if our `radio.send('abc')` was compatible with their "radio receive string", and their "radio send string" was compatible with our `radio.receive()`.

ntoll:
@dpgeorge https://github.com/search?q=org%3AMicrosoft+pxt

dpgeorge:
Ok, so to match the DAL/pxt we need to prefix all messages with the bytes `b'\x01\x00\x01'`, and also strip this off any incoming data.  But we can only be compatible with their "send string" and "receive string" functions because they don't include anything in the header to distinguish between string and number types (the user must know what they are expecting).  This is a pxt issue and not a DAL issue.  We are compatible with the DAL "datagram" protocol by using the `b'\x01\x00\x01'` header.

So the proposal is:

- `send_bytes(b)` just sends raw bytes, no headers or anything.

- `receive_bytes(b)` receives raw bytes into the given buffer.

- `send(s)` where s is a str: same as `send_bytes(b'\x01\x00\x01' + bytes(s, 'utf8'))`.

- `send(i)` where i is an int: same as `send_bytes(b'\x02' + bytes(str(i), 'utf8'))` (optional to implement at this point)

- `receive()` will check the first byte for the type, if it's 1 then strip 2 more bytes of the header and then parse the rest as a UTF8 string.  If the first byte is 2 then it's an integer, etc.

The above will need testing against a pxt device to check that it actually works.

ntoll:
+1 LGTM... I'll update the docs.

ntoll:
Actually, upon reflection, can we just make this string only for the moment.

dpgeorge:
There is still some ambiguity in the docs: if I call on(), then config(...), then off(), then on() again, does the second "on" reset the setting to the defaults, or use the last settings?

Stepping back a bit, I think it's a bit convoluted/confusing to have config() with no args *reset* all the state, and config(...) with a few args reset everything except the given args.  I'd propose the following:

- all radio settings are module-wide global variables and upon a fresh import of the module are set to their defaults
- on() turns the radio hardware on, using the current settings, and does not change any settings
- off() turns the radio hardware off, and does not change any settings
- reset() resets all settings to their defaults
- config(**kwargs) sets zero or more settings as given in the keyword arguments, but leaves the rest unchanged

ntoll:
@dpgeorge given the discussion today on the mailing list I was going to propose more-or-less the same. :-)

dpgeorge:
> given the discussion today on the mailing list

Ah, yes, I see it now!

Any tweaks to my proposal?  In #306 I suggested config('param') is used for get_config.

ntoll:
How about `get_config('param')` to be explicit and so the `config` function only does one thing?

ntoll:
BTW... my work from this morning is on ntoll/radio. Unfinished and with some documentation re-arrangement too so things a laid out more logically. YMMV but you can see where I was going with it.

dpgeorge:
> How about get_config('param') to be explicit and so the config function only does one thing?

That makes it clear and explicit, but then "config" should be "set_config" (as per the mailing list discussion).  Probably "reset" can stay as "reset".  Note that the decision here is ultimately the decision for #306...

deshipu:
I don't know if that helps, but most Python style guides recommend that function and method names should be verbs or verb phrases, such as "get_configuration", "configure" or "reset", while attributes and properties should be nouns or noun phrases, such as "configuration" (they also advise against using shortened words or acronyms). However, Micropython seems to be developing this pattern where a method is used in place of a property (because property is inefficient), and where calling it with a parameter sets the property, while calling it without a parameter gets it.

There is also that recommendation that if an operation is costly (in time or otherwise) or has side effects, it should be a method, not a property. So the question is, does setting those properties have significant side effects or time penalties, or does it just set some internal registers, and the actual work is done when calling other methods?

ntoll:
A while back, and after much discussion at both PyCon UK and on the mailing list, it was decided to use *just* method for the microbit related work. BBC based evidence suggested kids found these easier to understand and work with. Personally, I was on the "we should use properties, when they make sense" side of the fence but respect the decision of the group. It's evidence based, brings a certain consistency to things, and we've not had any feedback to the contrary of our decision except from programmers occasionally asking why we don't use properties. :smile:

The micro:bit is but a first step. Those that take to it will dive into "wider" Python when given the chance. :-)

ntoll:
This documentation is ready to be proof-read and (hopefully) merged ASAP. It includes two additions to the docs:

* Complete API documentation.
* A teacher friendly tutorial that can be adapted for classroom use.

Happy to make modifications.

docs/index.rst
```diff
@@ -70,7 +69,11 @@ Projects related to MicroPython on the BBC micro:bit include:
    os.rst
    random.rst
    neopixel.rst
-   ble.rst
+   pin.rst
+   radio.rst
+   random.rst
+   spi.rst
+   uart.rst
```
dpgeorge: Why are random, spi and uart now included twice?

docs/radio.rst
```diff
@@ -0,0 +1,129 @@
...
+    The ``power`` is an integer value from 0 to 7 (inclusive) to indicate the
+    strength of signal used when broadcasting a message. The higher the value
+    the stronger the signal, but the more power is consumed by the device. The
+    numbering translates to positions in the following list of dBm (decibel
+    milliwatt) values: -30, -20, -16, -12, -8, -4, 0, 4.
+
+    The ``address`` is an arbitrary name, expressed as a 32-bit address, that's
+    used to filter incoming packets at the hardware level, keeping only those
+    that match the address you set. The default used by other micro:bit related
+    platforms is ``b'uBit\x00'``. It must be exactly 5 bytes in length.
```
dpgeorge: Change `b'uBit'` to 0x75626974.  And remove sentence "It must be exactly 5 bytes in length.".

docs/radio.rst
```diff
@@ -0,0 +1,129 @@
...
+
+    The ``data_rate`` indicates the speed at which data throughput takes
+    place. Can be one of the following contants defined in the ``radio``
+    module : ``RATE_250KBIT``, ``RATE_1MBIT`` (the default)  or ``RATE_2MBIT``.
+
+    If ``config`` is not called then the defaults described above are assumed.
+    Since these are named arguments with sensible defaults, re-setting is as
+    simple as calling ``config()``. Calling with a subset of the named
+    arguments (for example, ``config(power=3)`` will cause all the other
+    settings to default their values.
```
dpgeorge: The behaviour has changed from this.  Call reset() to reset the settings.

docs/radio.rst
```diff
@@ -0,0 +1,129 @@
...
+.. py:function:: on()
+
+    Turns the radio on. This needs to be explicitly called since the radio
+    draws power and takes up memory that you may otherwise need.
+
+.. py:function:: off()
+
+    Turns off the radio, thus saving power and memory.
+
+.. py:function:: config(length=32, queue=3, channel=7, power=0, address=0x75626974, group=0, data_rate=radio.RATE_1MBIT)
```
dpgeorge: Technically the type signature is more like `config(**kwargs)` (since that's how you'd implement this function in Python), where kwargs can be on of the listed parameters.  I'd change it to that, provide a short description what the notation **kwargs means, and then write the default value in the description of each param below.

docs/radio.rst
```diff
@@ -0,0 +1,129 @@
...
+
+.. py:function:: receive_bytes()
+
+    Receive the next incoming message on the message queue. Returns ``None`` if
+    there are no pending messages. Messages are returned as bytes.
+
+.. py:function:: send(message)
+
+    Sends a message string. This is the equivalent of
+    ``send_bytes(bytes(message, 'utf8'))``.
```
dpgeorge: It actually prepends the bytes `b'\x01\x00\x01'` to the message, so that it's compatible with the DAL format.

docs/radio.rst
```diff
@@ -0,0 +1,129 @@
...
+
+.. py:function:: receive()
+
+    Works in exactly the same way as ``receive_bytes`` but returns
+    strings. Equivalent to ``str(receive_bytes(), 'utf8')``. Will raise a
+    ``ValueError`` exception if conversion to string fails.
```
dpgeorge: It checks that the first 3 bytes are 0x01, 0x00, 0x01, then strips them before converting to a string.

docs/tutorials/radio.rst
```diff
@@ -0,0 +1,163 @@
+Radio
+-----
+
+Interaction at a distance feels like magic.
+
+Magic might be useful if you're an elf, wizard or unicorn, but such things only
+exist in stories.
+
+However, there's something much better than magic: physics!
```
dpgeorge: Yay!

docs/tutorials/radio.rst
```diff
@@ -0,0 +1,163 @@
...
+
+Wireless interaction is all about physics: radio waves (a type of
+electromagnetic radiation) have some sort of property (such as their amplitude,
```
dpgeorge: I'd say: "(a type of electromagnetic radiation, similar to visible light)"

docs/tutorials/radio.rst
```diff
@@ -0,0 +1,163 @@
...
+Bytes
++++++
+
+A byte is a unit of information that (usually) consists of eight bits. A bit is
+the smallest possible unit of information since it can only be in two states:
+on or off.
+
+Bytes work like a sort of abacus: each position in the byte is like a
+column in an abacus - they represent an associated number. In an abacus these
+are usually thousands, hundreds, tens and units (as I would call them in the
```
dpgeorge: I'd say: "(as one would call them in the UK)" since I don't think you use the first person anywhere else.

docs/tutorials/radio.rst
```diff
@@ -0,0 +1,163 @@
...
+
+The problem with radio is that you can't transmit directly to one person.
+Anyone with an appropriate aerial can receive the messages you transmit. As a
+result it's important to be able to differentiate who should be receiving
+broadcasts.
+
+The way the radio built into the micro:bit solves this problem is quite simple:
+
+* It's possible to tune the radio to different channels (there are 100 to choose from). This works in exactly the same way as kids' walkie-talkie radios: everyone tunes into the same channel and everyone hears what everyone else broadcasts via that channel.
+
```
dpgeorge: There are actually 101 channels, 0-100 inclusive.

examples/radio.py
```diff
@@ -0,0 +1,30 @@
+# A micro:bit Firefly.
+# By Nicholas H.Tollervey. Released to the public domain.
+import radio
+from microbit import display, Image, button_a, sleep
+import random
+
+# Create the "flash" animation frames.
+flash = [Image((str(i) * 5 + ':') * 5) for i in range(9, -1, -1)]
```
dpgeorge: Alternatively: `[microbit.Image().invert()*(i/9) for i in range(10)]`

docs/radio.rst
markshannon: Could you just say "returns whatever was sent"? That future proofs it for when we add new types.

examples/radio.py
ntoll: I think you win the most obscure one-liner of the year award with that one... :-) I hadn't thought of using invert. However, the result of the list comprehension is in reverse order so the convoluted `range` call should remain (or we do [reversed([...])] which seems even more obscure) ;-)

ntoll:
OK... corrections and spelling mistakes sorted (it actually said "the best way to receive a massage" in the tutorial). :-)

Merge it!

PR #309, opened 2016-06-14
Audio: Various enhancements to simplify playing of samples

markshannon:
This a major overhaul of the audio module, to:
* Convert samples to unsigned 8bit. 
* Make the AudioFrame object support the buffer protocol.
* Auto-detect whether speaker is connected to one or two pins.

In combination, this will allow playing samples from files and allow the use of two pins while having an API compatible with that of the music module.


dpgeorge:
I can't test this for a while, but it looks ok, so if it works we should just merge :)

markshannon:
Given that this is going into audio-dev rather than master, I've no objection to you merging.
I expect that the code will be re-reviewed and tested before audio-dev is merged into master.

dpgeorge:
Merged.  I had issues compiling with latest gcc (v6.1.1) in that it didn't fit in RAM, was about 4 bytes too large.

PR #314, opened 2016-06-23
Audio: Allow various effects like reverb.

markshannon:
This PR:
Adds basic arithmetic (addition and multiplication) to `AudioFrame`, which allows some simple effects.

I've also noticed some erratic behaviour which I suspect is related to GC. I think that allowing GC in interrupts is causing trouble. Consequently, I've forbidden any allocation in the data-fetcher. 
It seems to have fixed the problem, but does lead to less than ideal error messages.
(We probably ought to do something similar for the display).


dpgeorge:
It works!  So I merged it.

But examples/waveforms.py is now broken, due to samples now needing to be 0-255, and also because of disabling of allocation during interrupts.

PR #315, opened 2016-06-26
Display: Prohibit allocation in interrupts.

markshannon:
This prevents non-deterministic and impossible-to-debug errors caused by asynchronous GC.

Unless we can make the GC fully re-entrant from interrupts, which is not feasible, then we must prohibit GC in interrupts. 
This PR does that. (https://github.com/bbcmicrobit/micropython/pull/314 does the same for audio)

### DIFF
--- a/source/microbit/microbitdisplay.cpp
+++ b/source/microbit/microbitdisplay.cpp
@@ -30,6 +30,7 @@
 
 extern "C" {
 #include "py/runtime.h"
+#include "py/gc.h"
 #include "modmicrobit.h"
 #include "microbitimage.h"
 #include "microbitdisplay.h"
@@ -130,7 +131,6 @@ static volatile bool wakeup_event = false;
 static mp_uint_t async_delay = 1000;
 static mp_uint_t async_tick = 0;
 static bool async_clear = false;
-static bool async_error = true;
 
 STATIC void async_stop(void) {
     async_iterator = NULL;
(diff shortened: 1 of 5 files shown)


dpgeorge:
This would be a breaking change for the microbit API behaviour.

Other uPy implementations don't allow allocation in interrupts because the GC is not re-entrant.  Actually, there is pending support for multithreading (https://github.com/micropython/micropython/pull/2165) which includes a thread-safe GC that could be used to allow allocation during an IRQ, but the issue is that IRQ's would be disabled all the while during a garbage collection.

One could also use "soft interrupts" to get around the problem (https://github.com/micropython/micropython/pull/2081) but likely that's not real-time enough for audio (might be for display stuff though).

Note that uPy has built-in support for raising exceptions even with the heap locked (ie when gc_lock is active).  It uses an emergency exception buffer; see config variable MICROPY_ENABLE_EMERGENCY_EXCEPTION_BUF.  That allows all error messages to propagate outside of an interrupt handler if the heap is unusable.

So we need a plan to move forward, considering the following points:

- is it ok to make a breaking change whereby allocation cannot occur in the iterables used to make animations on the display?

- can we switch to soft interrupts for the display update so that allocation can occur?

I think the display iterable is the only place where allocation can occur within an interrupt at the moment (excl audio which is in development).

markshannon:
I think it is OK to make a "breaking" change in this case, as allowing GC in interrupts was already broken.
Allocating in an interrupt works most of the time, but when it doesn't it is next to impossible to debug.
Soft interrupts might work, but there is still the problem of GC pauses causing glitches that are *very* hard to diagnose. 

Banning allocation might seem quite restrictive, but this only impacts sophisticated users who are using generators or custom iterators. If you pass a list or tuple of images, then this PR makes no difference.

dpgeorge:
> I think it is OK to make a "breaking" change in this case, as allowing GC in interrupts was already broken.

Yes I would agree with that.

source/microbit/microbitdisplay.cpp
```diff
@@ -300,14 +301,24 @@ static void microbit_display_update(void) {
...
             } else {
+                gc_unlock();
                 if (!mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t*)nlr.ret_val)->type),
                     MP_OBJ_FROM_PTR(&mp_type_StopIteration))) {
                     // an exception other than StopIteration, so set it for the VM to raise later
-                    MP_STATE_VM(mp_pending_exception) = MP_OBJ_FROM_PTR(nlr.ret_val);
+                    //If memory error, add appropriate message.
+                    mp_obj_exception_t *ex = (mp_obj_exception_t *)nlr.ret_val;
+                    if (mp_obj_get_type(nlr.ret_val) == &mp_type_MemoryError) {
+                        ex->args = (mp_obj_tuple_t *)mp_obj_new_tuple(1, NULL);
```
dpgeorge: This (and the line below) is allocating memory within an interrupt handler, something we are trying to avoid!  So I'd say just printf the message (actually, use mp_printf(&mp_plat_print, ...)) and let the MemoryError exception through as-is.

source/microbit/microbitdisplay.cpp
```diff
@@ -300,14 +301,24 @@ static void microbit_display_update(void) {
...
+                        ex->args->items[0] = mp_obj_new_str("Allocation in interrupt handler",
+                                                            strlen("Allocation in interrupt handler"), false);
+                    }
+                     MP_STATE_VM(mp_pending_exception) = MP_OBJ_FROM_PTR(nlr.ret_val);
```
dpgeorge: Bad indentation.

markshannon:
@dpgeorge I've addressed your comments and rebased.

markshannon:
I've also modified the handling of iterators to allow allocation during the first iteration, much like the audio module.
So, this is OK:
```python
def g():
     img = Image(5,5)
     for j in range(5):
          img.set_pixel(j,j,j)
          yield img
display.show(g())
```

dpgeorge:
Merged!

PR #317, opened 2016-07-10
Audio: Allow allocation in first iteration of audio source.

markshannon:
This simplifies writing generators. 
This is now legal:
```python
def audio_gen():
    f = audio.AudioFrame()
    while True:
          yield f
```
It also shortens the start-up pause, by starting playing from the first frame, not an empty frame.
I've updated `examples/waveform.py` although it still needs the modification to `display.scroll` in https://github.com/bbcmicrobit/micropython/pull/315 to work.

### DIFF
--- a/examples/waveforms.py
+++ b/examples/waveforms.py
@@ -20,7 +20,7 @@ def show_wave(name, frame, duration=1500):
 frame = audio.AudioFrame()
 
 for i in range(len(frame)):
-    frame[i] = int(math.sin(math.pi*i/16)*124+0.5)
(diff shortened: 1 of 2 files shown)


dpgeorge:
This looks good, and it compiles, but unfortunatelly I no longer have any speakers so can't test it.  Still, I'd be happy to merge it.

markshannon:
If you have crocodile clips, then (cheap) headphones work as well as speakers.
Take care with the volume though, holding an earphone near to your ear, rather than in or on it, is best.

dpgeorge:
I found a speaker, so tested it and merged.

PR #319, opened 2016-07-15
Merge Audio in master.

markshannon:
This includes one final commit, which ensures that *all* state is reset properly between plays.
A version of "sam" based on this branch works without glitches.


markshannon:
Note that for `audio` and `display.scroll` to work correctly together faultlessly, #315 will need to be merged as well.

ntoll:
Just to be certain, you mean SAM based on commit e36f5782b625b3b2d38dc5735f981066855a7a49 in the sam-speech branch..?

markshannon:
I was trying to be indirect to avoid letting the cat out of the bag, but:
8f2b368e9d3f5626c6f846d03d96ed67eb61fd07 which is sam-speech rebased onto this branch.

ntoll:
Hah! I like it ~ cloak and dagger although I don't think *that* many people are watching what we're doing. I think if we just don't announce anything until Monday (keynote). The new version of Mu and uFlash won't be pushed until Monday morning, so unless you have a compiler, you won't be able to get it until then. ;-)

docs/audio.rst
```diff
@@ -0,0 +1,71 @@
...
+    ``pins`` specifies which pins the speaker is connected to.
+
+Classes
+=======
+
+.. py:class::
+    AudioFrame
+
+    An ``AudioFrame`` object is a list of 32 samples each of which is a signed byte
+    (whole number between -128 and 127).
```
dpgeorge: Should be 0 and 255.

docs/audio.rst
```diff
@@ -0,0 +1,71 @@
...
+
+A sound source is an iterable (sequence, like list or tuple, or a generator) of
+frames, each of 32 samples.
+The ``audio`` modules plays samples at the rate of 7812.5 samples per second,
+which means that it can reproduce frequencies up to 3.9kHz.
+
+Functions
+=========
+
+.. py:function:: play(source, wait=True, pins=(pin0, pin1))
```
dpgeorge: pins should be pin and return_pins.

examples/play_file.py
```diff
@@ -0,0 +1,14 @@
...
+    ln = -1
+    while ln:
+        ln = file.readinto(frame)
+        yield frame
+
+def play_file(name, pin=None, return_pin=None):
+    #Do allocation here, as we can't do it in an interrupt.
+    frame = audio.AudioFrame()
+    with open(name) as file:
+        audio.play(audio_generator(file, frame), pin=pin, return_pin=return_pin)
```
dpgeorge: I guess you can now rewrite this example to create the `frame` and open the `file` at the start of `audio_generator`?  Not important though.

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,648 @@
...
+        if (!mp_obj_is_subclass_fast(MP_OBJ_FROM_PTR(((mp_obj_base_t*)nlr.ret_val)->type),
+            MP_OBJ_FROM_PTR(&mp_type_StopIteration))) {
+            // an exception other than StopIteration, so set it for the VM to raise later
+            //If memory error, add appropriate message.
+            mp_obj_exception_t *ex = (mp_obj_exception_t *)nlr.ret_val;
+            if (mp_obj_get_type(nlr.ret_val) == &mp_type_MemoryError) {
+                ex->args = (mp_obj_tuple_t *)mp_obj_new_tuple(1, NULL);
+                ex->args->items[0] = mp_obj_new_str("Allocation in interrupt handler",
+                                                    strlen("Allocation in interrupt handler"), false);
+            }
```
dpgeorge: Similarly to display, we should just print an error here then let the VM raise the MemoryError as usual.

source/microbit/modaudio.cpp
```diff
@@ -0,0 +1,648 @@
...
+    PinName p = microbit_obj_get_pin_name(pin);
+    nrf_gpio_cfg_input(p, NRF_GPIO_PIN_NOPULL);
+    // Allow 1µs to settle.
+    nrf_delay_us(1);
+    return nrf_gpio_pin_read(p);
+}
+
+static const mp_obj_t big_pins[3] = { (mp_obj_t)&microbit_p0_obj, (mp_obj_t)&microbit_p1_obj, (mp_obj_t)&microbit_p2_obj };
+
+static void audio_auto_set_pins(void) {
```
dpgeorge: I don't think this functions should search all pins.  What if you have a device connected to pin1 which shouldn't have a voltage source applied (eg you connected pin1 to pin2 with a wire and have pin2 in output mode, pin1 in input mode)?  Or you have something connected to pin1/pin2 which is set up and running (eg servo on pin1, or a i2c/uart/spi on pin1/pin2)?

I think it should work as follows: if both pin and return_pin are None then it uses the default, which is to have pin=pin0, and then return_pin either None, pin1 or pin2.  This function can then safely output on pin0 (since the user should at least have 1 of the wires of the speaker connected to pin0) and input on the other pins to see if they change.  Although there is still the issue that the other pins (pin1/pin2) need to already be in input mode for it to work.  You can't change their mode because they may be doing something else (eg servo PWM).

dpgeorge:
@markshannon did you check that servo control works as before with your new PWM implementation?

dpgeorge:
It seems that the music module still uses the existing setAnalogPeriodUs function to control the sound.  Will that work correctly?

ntoll:
From an "aural" point of view... it still plays music that sounds as it was. Not sure that helps though.

dpgeorge:
From a backwards-compatible point of view, the only (?) things that need to be checked before merging are: servo control working (since it uses new PWM code) and music module working.  Otherwise the audio stuff is new and so there's no existing code using it, and the issues I raised above can be fixed at a later date.

ntoll:
OK... so I can confirm that the music module works (I was using it yesterday evening), at least in terms of producing sound.

ntoll:
I'm sitting in a tutorial session at EuroPython, not sure I'm in a position to check servo control.

ntoll:
OK... so the speech-docs branch is updated and ready to merge.

deshipu:
I brought some servos we can test on, as well as a pocket osciloscope.


On Sun, 17 Jul 2016 01:06:12 -0700
Nicholas Tollervey <notifications@github.com> wrote:

> I'm sitting in a tutorial session at EuroPython, not sure I'm in a
> position to check servo control.
> 
> ---
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/bbcmicrobit/micropython/pull/319#issuecomment-233170559


-- 
Radomir Dopieralski

ntoll:
@deshipu where are you? I'm (hiding) at the beginner's day. Care to come over so we can test?

ntoll:
@deshipu I'm trying to get things so they can merge ASAP... once done I can update uflash, Mu and other things so they'll be available to the attendees after my keynote.

deshipu:
I will come in about 20 minutes, where do I find you?

ntoll:
@deshipu I'm at the beginner's day, hiding in a corner by the lifts on the 2nd floor.

dpgeorge:
@ntoll can you please check the compilation of this PR on your machine.  We need to reduce the size of the heap (again) to make it work but I need to know how much by.  Do the following:

- edit line 102 of source/microbit/mprun.c so it's: `static uint32_t heap[9708 / 4];`
- rebuild
- run `size build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython` and tell me the output.

ntoll:
@dpgeorge 

```
$ size build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython
   text	   data	    bss	    dec	    hex	filename
 225228	    292	  11928	 237448	  39f88	build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython
```

ntoll:
@dpgeorge wrt servo control:

I've checked with @deshipu who's stood next to me right now. It works, we're seeing the right things via the oscilloscope and music plays too. We have noticed that if you use music first and then attempt to use write_analog afterwards it doesn't work, perhaps something isn't being reset or released by the music module. In any case I'm not sure you'd want to use music and servos at the same time.

dpgeorge:
> We have noticed that if you use music first and then attempt to use write_analog afterwards it doesn't work,

Music uses the old (DAL) way of setting PWM.  write_analog uses the new (direct hardware registers) way.  So probably there is some conflict there.  We can resolve that later.

ntoll:
+1

dpgeorge:
Merged.  What is next, speech docs?  Can they be rebased?

ntoll:
Mark's sam-speech branch. Final cleanup for speech-docs happening now, they'll be ready in 2 mins.

ntoll:
OK... speech-docs are good to be merged once Mark's speech work is in.

dpgeorge:
I can't locally see the commit that Mark references above... what branch is it on?

ntoll:
https://github.com/markshannon/micropython/tree/sam-speech

ntoll:
I rebased and compiled from latest master, just about to check it actually works.

ntoll:
speech-docs rebased to (current) master. The demos work with what's in Mark's branch.

dpgeorge:
Ok, I guess I'll merge Mark's sam-speech branch...

whaleygeek:
I want to use music and servos at the same time :)

ntoll:
@whaleygeek you can't even begin to imagine the amount of effort that's gone into getting things this far. That @dpgeorge and @markshannon have spent significant parts of their weekend demonstrates how awesome they are. I'm afraid you'll have to wait for a musical servo. It'll come and PR's are welcome.

whaleygeek:
Also can someone characterise and publish any details about shifts in
memory footprint as a result of this addition? I have a number of scripts
that 'only just fit' today, I'd like to upgrade when this is released but
would like to think that nothing breaks as a result of different RAM
footprint. Thanks.

ntoll:
My advice is to just test your scripts. 

As far as I can tell, there's been an adjustment to the heap size (see comments above for details), other than that I don't believe we've done anything else since it's all just new code (ergo in the flash memory).

dpgeorge:
> Also can someone characterise and publish any details about shifts in memory footprint as a result of this addition?

Yes we are keeping an eye on that and will let people know (somehow...) about any changes.

whaleygeek:
Auto detect of speaker configuration sounds like a potentially hardware
damaging feature. Is it optional?

dpgeorge:
> Auto detect of speaker configuration sounds like a potentially hardware damaging feature. Is it optional?

Yes, as long as you provide arguments for the pins for audio it won't auto-detect.

PR #320, opened 2016-07-17
Fix crash caused by allocation in interrupt.

markshannon:
Avoid allocation by:
1. Basing the repeat iterator on indexable sequences not iterators. That way we avoid allocating when reaching the end of the sequence.
2. Creating a "facade" object that makes a string appear to be a sequence of images.

Also fixes a timing issue that caused the first image in a sequence to sometimes only appear momentarily.


markshannon:
There is a very subtle change to the API, in that a custom iterable that is not a sequence and does no allocation will no longer work with `loop=True`.
I doubt that anyone has ever or is ever likely to make such an iterable.

ntoll:
Good to know. :-)

Many thanks and, as always, great work..!

ntoll:
So, happy for me to merge in speech? :-)

PR #321, opened 2016-07-17
Speech

markshannon:
"SAM" based speech module.
@dpgeorge @ntoll Can you check that the heap size is OK on your machines.


dpgeorge:
Builds ok for me, but has a few warnings (eg line 51 of sam/render.c, comparison is always false).  Heap size fits, and there are 44 bytes of RAM still unused (but we should leave them for room to grow later).

dpgeorge:
Firmware flash size for me is 237940.

ntoll:
Compiles fine. Works for me. Let's merge the damn thing (Damien?). I'll rebase my speech docs from master then merge once this is in. I get firmware size of 242216.

markshannon:
The "comparison is always false" warning is benign as I've made the buffer size configurable and it is currently set to 256. The indexing variable is a `uint8_t`. 
When (if?) I convert SAM to use streaming internally, I will be able to shrink these buffers a lot so the check should remain.

markshannon:
How are you measuring the firmware size?
I get:
```
$ wc -c build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython.bin
241864
```

ntoll:
OK... if no-one has any problems, I'll merge it! Speak now, of forever hold your peace (I'll give you 5 mins to hold me off!). Or just say "Aye" ;-)

dpgeorge:
Firmware size measurement: `size build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython`

Aye!

ntoll:
@markshannon `size build/firmware.hex`

ntoll:
Yay! It's in. I'll merge in the docs too.

PR #323, opened 2016-07-19
Changes  panic example code to less than 255

hlwill:
404 is more than the limit of 255 stated in the docs. This makes the previous example misleading.


dpgeorge:
Thanks!

In the latest version of the DAL the range has been extended to 0-999 (inclusive), so if/when we update we need to remember to change the docs.

PR #326, opened 2016-07-21
Add radio, os, speech to help() in the REPL

bfayers:
Added os, radio, speech to the help() in REPL


dpgeorge:
Great, thank you!

PR #340, opened 2016-08-14
Add missing comma to answer list

markbasham:
Does not cause an error, but does cause 2 elements to be unnecessarily concatenated.  Only spotted this as my first random line was effected.


dpgeorge:
Thanks!

Issue #337, opened 2016-08-09
pin0.write_analog stops working after music.play()


deshipu:
After the following code:

```
>>> from microbit import *
>>> import music
>>> pin0.set_analog_period(20)
>>> pin0.write_analog(512)
```

I get the following result:
![img_20160809_171645](https://cloud.githubusercontent.com/assets/1181357/17521950/8854a776-5e55-11e6-8cc7-1aad19040079.jpg)

Then after this:

```
>>> music.play(music.WAWAWAWAA)
>>> pin0.write_analog(512)     
```

I get no output on pin0:
![img_20160809_171731](https://cloud.githubusercontent.com/assets/1181357/17521972/9a17e860-5e55-11e6-83b1-adf8322b1576.jpg)

deshipu:
Using ``speech.say()`` in place of ``music.play()`` doesn't have that effect.

carlosperate:
I've just tested this with a speaker and oscilloscope with version 0.9 (the one on the online editor since October 2016) and  while #345 did fix this in a way (`pin0.write_analog()` works fine after `music.play()`), something has caused the `music.play(music.WAWAWAWAA)` to not reproduce the tune output on pin0, basically the first `pin0.write_analog()` continues uninterrupted.

I've tested the same with the latest master (with the version1 branch merged) and I can confirm this version works as expected:
*  `pin0.write_analog()` produces the correct output
* `music.play(music.WAWAWAWAA)` plays the tune as expected
* second `pin0.write_analog()` works again, however, this second `write_analog` uses the period last configured by the music module.

We haven't really defined any default period for the `write_analog` so I would consider this expected behaviour.

Both issues are fixed on master, so I'll close this one.

Issue #331, opened 2016-07-24
Strange behaviour of pins when switching modes


markshannon:
The various behaviours of pins interact in strange and non-obvious ways.

For example, playing audio on a pin that is already set to an analog value temporary controls the pin, then reverts to the analog value, but writing a digital value is effectively ignored (the value only persists as long of the analog period)
When you start considering that some pins are dedicated to i2c and buttons, things start getting a bit weird.
You might want to play audio on the i2c pins, but you really ought to explictly disable i2c first.
Writing digital values to the button pins has a strange effect:
pin5.write_digital(1) and button A seems to function correctly,
but pin5.write_digital(0) and button A claims to be always pressed.

And as a really strange example:

``` python
>>> pin5.set_analog_period(37)
>>> pin5.write_analog(512)
```

button_a flips between pressed and not pressed about once a second as the debouncing moves in and out of phase with the analog output.

What I would propose is:
- Button pins should be read-only
- The simpler modes, touch, digital, and analog, should be implicitly switched when a relevant method is called. This is supposed to happen already, but doesn't always.
- The more complex modes, audio, display, i2c, and spi, should need to be explicitly turned off before their pins can be used for anything else. 

So, for example, any attempt to read or write to one of the display pins should raise an exception unless the display has been turned off.

dpgeorge:
Isn't this an issue for upstream DAL?

markshannon:
It would be if didn't we implement our own display, pwm and audio modules.

markshannon:
Fixed by #345

PR #345, opened 2016-09-19
Safer behaviour when the mode of a pin changes. Fix up PWM implementation

### LINKED ISSUE #335: Using music.play after speech.say results in no sound or garbled sound
Using the mu editor for python / microbit. It reports: MicroPython v1.7-9-gbe020eb

The following program:

``` python
from microbit import *
import music
import speech
music.play(music.WAWAWAWAA, wait=True)
speech.say('Hello world')
sleep(2000)
music.play(music.WAWAWAWAA, wait=True)
```

Does not play the second WAWAWA... tune (with speaker attached to GND + PIN0). Remove speech.say() and it works as expected. Seems that speech.say leaves the microbit sound system in some altered state? I have tried music.reset() but does not seem to help.

markshannon:
Rework the pwm module to allow smooth transitions when period or duty cycles are changed.
Make music module use our pwm (not the DAL's pwm).
Modify the pin and related modules to ensure that attempting to use a pin when it is already used for another purpose, raises an exception.

For example:

``` python
>>> pin5.write_analog(300)
TypeError: Pin 5 in button mode
>>> pin6.write_analog(300)
TypeError: Pin 6 in display mode
>>> display.off()         
>>> pin6.write_analog(300)
>>> display.on()          
TypeError: Pin 6 in analog mode
>>> pin6.write_analog(0)  
>>> display.on()        
>>> 
```

Fixes #335, #337, #331, #288.


dpgeorge:
There are a lot of changes here.... what is the delta of code size and RAM usage?  I see you needed to reduce the heap by about 150 bytes, and that's going to potentially break some large scripts.

whaleygeek:
Hmm,

Changes in RAM usage are going to be a nightmare, especially if user
scripts that once worked now no longer fit or they crash when they run,
when a user upgrades to a more recent version of MicroPython. It will be a
terrible experience for users.

I know that the RAM utilisation story is complex, and especially
challenging with MicroPython due to the need to store the Python bytecodes
in RAM. But I think it needs good management now that the whole system is
out of beta (you did know that we're out of beta I presume? The beta tag
was removed from the website w/c 12th Sept). So this is now released
product!

Clearly the heap is one area we could set a high water mark on and signpost
in the code so that it never changes in the future. Other static RAM
utilisation could be managed by perhaps setting an overall static RAM high
water mark that indicates exactly how much RAM is left for user programs
(and us policing this guaranteed amount of static RAM in all future builds).

Actual runtime RAM utilisation of each feature is much harder to control. I
wonder if the only way we could control this is to have a representative
set of test scripts that we must run before committing features that affect
runtime RAM utilisation?

Damien you're doing a great job to spot this, but it feels to me like there
needs to be more signposts and checks in place so that others can police it
too, otherwise we are always highly reliant on you checking every single PR
carefully.

markshannon:
@dpgeorge Thanks for spotting the reduced heap size, I'd just chosen the lowest option when rebasing with the intention of fixing it up later, and then forgotten. 
The available heap has in fact *increased* by a tiny bit.

Filesystem size will have reduced, but that's only temporary as this takes us very close to being able to drop the DAL.

dpgeorge:
> Filesystem size will have reduced, but that's only temporary as this takes us very close to being able to drop the DAL.

If we also drop mbed then we can reclaim a whole 2k of RAM, which would be the idea outcome.

markshannon:
I plan to add a test-suite for this in the next few days. 
You may care to hold off reviewing this properly until then.

markshannon:
OK, I've added the test and it passes.

dpgeorge:
@markshannon very nice work!  I like that there are more "const" objects (ie less RAM usage) and that there's no longer a need for DAL Pin and Button, which take up lots of RAM.  It seems that the code size actually *decreases* by about 500 bytes (probably due to less functions from DAL needed in the link stage).

One point re the API: I think it should be allowed to do pin.write_digital even if the pin is in analog mode.  There's no real need to force users to do pin.write_analog(0) then pin.write_digital(x).

Also, please rebase :)

dpgeorge:
@markshannon ping, see above comment.

markshannon:
Found some time to finish this at last.
I've extended the tests to check that repeated switching of modes still leaves things in a sane state.
Uses `qstr`s rather than `int`s to identify modes, which should make things more robust when adding more modes (like neopixel or audio-record).

(No need to rebase, github can do that for us now)

dpgeorge:
> Uses qstrs rather than ints to identify modes, which should make things more robust when adding more modes

It's just as robust to add a new member to an enum, so I don't see a problem using ints here.

Why expose the `pin.get_mode()` method at all?  Users should already know the mode of the pin because they are using the pin for some purpose.

Regarding get/set pull: for the constants PULL_DOWN etc why not use the nrf register values directly?  Then there is no need for a lookup table to convert between (-1,0,1) and the NRF_GPIO_PIN_xxx value.  The values of these constants (exposed on the Python side) should anyway be opaque to the user, so it's ok to use the register values.

And from a consistency point of view, if get_mode is going to be there in the API then it should be similar to get_pull, ie return an integer which corresponds to some constant.

markshannon:
I should have been a bit clearer, it is not ints that are less robust than qstrs. It is using an external index which cannot be verified. Searching the array of modes, looking for a marker, whether it is a string or an int, is more robust than an unchecked indexing operation.
In other words we cannot tell whether `pin_modes[PINMODE_DISPLAY]` is in fact the mode for the display without inspecting it. So we are better off just searching the modes. 
We need a qstr in the mode for error reporting, so we might as well use it as an index.

Strings are also self describing.
```python
>>> p.get_mode()
"display"
```
is a lot use informative than
```python
>>> p.get_mode()
7
```

The reason for using -1, 0, 1 rather than the nordic-defined constants is similar.
The values are opaque when setting them, but not when getting them.
Which is of 0,2,3 is up or down? It is rather more obvious with -1, 0, 1.
We could change set_pull/get_pull to take strings, but enums do have the advantage of offering some static typing thanks to auto-completion.

dpgeorge:
> It is using an external index which cannot be verified. Searching the array of modes, looking for a marker, whether it is a string or an int, is more robust than an unchecked indexing operation

To validate an index you just check that it's within a certain range (eg `idx >= 0 && idx <= 10`).

My other main point is that you don't need to expose pin.get_mode().

> The reason for using -1, 0, 1 rather than the nordic-defined constants is similar.
> The values are opaque when setting them, but not when getting them.

They are also opaque when getting.  You should test against known values, eg `pin.get_pull() == pin.PULL_UP`.

markshannon:
You can't validate an index just by checking it is in range. E.g.
```C
enum {
     DISPLAY_T,
     SPEECH_T
};
pinmode_t modes = {
    { "speech", ... },
    { "display", ... },
};
```
which is awfully easy to do when adding new values to a larger array.
I want to expose `pin.get_mode()` for testing. Quite happy to leave it undocumented.
`pin.get_pull()` is not opaque if you use it in the repl :wink: 
What is the advantage of using the Nordic values? The lookup table is tiny.

dpgeorge:
It's possible to protect against incorrect ordering in an array using a C99 feature:
```
pinmode_t modes[] = {
    [DISPLAY_T] = {...},
    [SPEECH_T] = {...},
};
```
We use this quite a bit in uPy core.

Well, anyway, if you want to keep the index as a qstr then I'm fine by it, especially since it's used to print a nicer error message.

> What is the advantage of using the Nordic values? The lookup table is tiny.

The fact that you need an LUT at all.  And there's extra translation code in get_pull() as well that you wouldn't need.

markshannon:
I'd like to get this merged as it is blocking other work. I really don't think it is worth quibbling over a few bytes in a lookup table, as we will save kilobytes when we remove the superfluous parts of the DAL.

dpgeorge:
> I'd like to get this merged as it is blocking other work. I really don't think it is worth quibbling over a few bytes in a lookup table, as we will save kilobytes when we remove the superfluous parts of the DAL.

The more bytes we save now the better.  It's hard to take things away in the future (because users rely on them).

I still think that making artificial pull values (-1, 0, 1) is not a good idea and wastes bytes.  If I were a new user and did pin.get_pull() and got 0, I'd think it was pull down (since 0 is a low logic level).  If I got -1 I'd be confused and think it was perhaps pulling it to a negative voltage.

My proposal is as before: the constants `PULL_UP`, `PULL_DOWN` and `NO_PULL` should be the corresponding nordic register values, and users should use `pin.get_pull() == pin.PULL_UP` to do checks.

What do other people think?

deshipu:
Personally, I would expect `NO_PULL` to map to `None`, simply because that's the implied default value when no parameter is passed to the pin's initialization. As for the other two values, constants sound reasonable, as long as they evaluate to `True`.

ntoll:
Regarding @dpgeorge's question... 

I'm wearing my teacher's hat here - the most important thing is that it's obvious what's going on and that "what's going on" is related to current conventions and practices (so such skills and knowledge can be transferred).

Given the above and upon careful reflection I prefer Damien's suggestion for the following reasons:

* `PULL_UP`, `PULL_DOWN` and `NO_PULL` are all unambiguous in what they mean.
* The use of such terminology is a positive thing since it's educating users how to describe things in the expected way.
* That such constants correspond to Nordic register values is also a good thing since anyone who wants to "look behind the curtain" and see what's going on will be able to line up such values with any specifications from Nordic.

I hope this makes sense..? It would certainly make my life as a teacher and learner easier.

Also, if what Damien says is true, it's more efficient in terms of memory.

Just my 2c. :-) Happy to answer questions.

markshannon:
@ntoll 
@dpgeorge and I both agree that `PULL_UP`, `PULL_DOWN` and `NO_PULL` should be defined, we differ in what values they should hold.
The difference is what should this print:
```python
pin0.set_pull(pin0.PULL_DOWN)
pin1.set_pull(pin1.NO_PULL)
pin2.set_pull(pin2.PULL_UP)
print(pin0.get_pull(), pin1.get_pull(), pin2.get_pull())
```
I think it should be `-1, 0, 1`
@dpgeorge suggests the underlying Nordic values, which would be `1, 0, 3` (and would save a few bytes of flash).
If I understand correctly, @deshipu would be happy with either `-1, None, 1` or `1, None, 3`

It looks like @ntoll has the deciding vote. Choose wisely :smile:

dpgeorge:
> The difference is what should this print ...

Not exactly.  What I'm saying is that the return value of pin.get_pull() is an *opaque* value and has no direct meaning to the user (they should not rely on it being anything sensible, it could even change from one version to the next).  So it makes no sense to print it out.  Instead, the things you can do with the return value are: 1) save it to a variable, 2) use it as an argument to pin.set_pull(), 3) compare it with the constants PULL_UP, PULL_DOWN, NO_PULL.

The first question is: should pin.get_pull() return an opaque or transparent (ie user can take meaning from it) value?  I vote opaque.

If it's transparent, then the follow-up question is: what values should these transparent values be?

markshannon:
I think we agree that the documentation should recommend that the user treats the value as opaque, 
and I agree that the user should treat the value as opaque. But what does "opaque" really mean?

It must have some value, and that value will be observable.
You simply cannot prevent people from printing it out.
In fact, when using the repl I like to be able to inspect the state of the peripherals.

ntoll:
What should the values be? I believe they should be obvious and unsurprising.

Is there a precedent in this case? 

Are hardware engineers expecting the Nordic register values? It strikes me this may be a reasonable expectation.

Alternatively, is there an existing numeric representation of pull up, pull down or no pull? If there *isn't* we should be careful: what value do we add by making one? We may even confuse people with a non-standard convention.

On balance, if we use Nordic register value we can flag what this means in our documentation, we save a small amount of memory and we give learners an opportunity to ask, what exactly *is* a register value?

Hope this makes (careful) sense. ;-)

dpgeorge:
> But what does "opaque" really mean?

In this context, to me opaque means that it's not defined in the docs to be any particular value, the user should not rely on it being any particular value, and it can change at any notice and user code should not break.

As such we're free to use any values, and using nordic register values is the simplest and most efficient (in code size and code execution).

markshannon:
OK, I've changed the pull constants to be the nordic provided values.
The tests still pass.

markshannon:
I've tacked the changes to the pull values to the final commit, so this PR should be squashed when merging.

dpgeorge:
> OK, I've changed the pull constants to be the nordic provided values.

I don't see the changes...?

> this PR should be squashed when merging.

GtiHub only allows "squash all and merge" or "rebase and merge" (we want the latter).  So If you want to have it squashed please do it on your branch corresponding to this PR.

markshannon:
Rebased, squashed and retested.

dpgeorge:
It's done!

Issue #349, opened 2016-10-01
Missing input() builtin?


zejn:
Hi,

I'm trying to use input() from user when connecting over serial, but it seems it's not implemented in bbcmicrobit port. Am I reading the code correctly?

ZanderBrown:
Yes it isn't implemented. Look at the UART module, Personally haven't had much luck with it though

ntoll:
Docs are here:

https://microbit-micropython.readthedocs.io/en/latest/uart.html

What exactly are you trying to achieve..?

ZanderBrown:
What @zejn is trying to do i don't know but if you have a working `uart.readline()` sample that would be great. So far i have been only been able to get a single character from it.

zejn:
Ah, I see now. Instead of using input(), one can use uart.read() to read input over serial. Works great for me.

With this I was able to create a simple serial chat protocol where I can instruct one microbit to send something over the radio and make the other receive and display it.

Thanks. :) :+1:

ZanderBrown:
Just what I've been doing! 😄 
![image](https://cloud.githubusercontent.com/assets/17074021/19022561/cb168c1c-88d2-11e6-9a95-96f725e43a07.png) (The box is supposed to be a bell (`␇`) but is displaying as a box)

Complete with Animations for message received, and whether or not the mb just sent or received.

PR #353, opened 2016-10-03
Add builtin input() function.

dpgeorge:
This patch adds the builtin input() function.  It's pretty minimal because it uses the existing readline function to capture the input line.

It was requested in #349, and although using UART was found to be a solution to a missing input() function, I think input is useful enough (and has many applications in the class room) to add it.


ZanderBrown:
Without it the REPL can be kinda limited

source/microbit/input.c
```diff
@@ -0,0 +1,44 @@
+/*
+ * This file is part of the MicroPython project, http://micropython.org/
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2013-2016 Damien P. George
```
markshannon: 2013? This took a while to write :smiling_imp:

markshannon:
LGTM and tested. 
@dpgeorge do you want to rebase before I merge?

markshannon:
I forget that github can do fast-forward merges now. Merging.

source/microbit/input.c
dpgeorge: :) it's from upstream uPy and was written long ago.

Issue #355, opened 2016-10-11
Repeated write to the same file fails.


markshannon:
This fails with `OSError 28`.

``` bash
ufs put tests/test_files.py
ufs put tests/test_files.py
ufs put tests/test_files.py
ufs put tests/test_files.py
ufs put tests/test_files.py
ufs put tests/test_files.py
```

markshannon:
Fixed by https://github.com/bbcmicrobit/micropython/pull/358

PR #356, opened 2016-10-11
Fixup file system. Fixes #318  and #355

### LINKED ISSUE #318: TextIO should not have `readinto` method.
Also BytesIO shouldn't have a `readline` method.

markshannon:
Remove incorrect method on bytes/textio classes.
Fix repeated write to the same file.

I've also had to reduce the number and sizes of files in the tests to make up for reduced space as a result of larger executable. Mainly the speech module, I expect.

### DIFF
--- a/source/microbit/fileobj.c
+++ b/source/microbit/fileobj.c
@@ -52,7 +52,20 @@ STATIC mp_obj_t file___exit__(size_t n_args, const mp_obj_t *args) {
 }
 STATIC MP_DEFINE_CONST_FUN_OBJ_VAR_BETWEEN(file___exit___obj, 4, 4, file___exit__);
 
-static const mp_map_elem_t microbit_file_locals_dict_table[] = {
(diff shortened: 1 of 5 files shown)


ZanderBrown:
the ratio of memory used by Python to available program space is fast becoming an issue

markshannon:
If you mean space in RAM, that has always been as issue. It's not getting any worse.
Expect a significant improvement in the not too far future...

If you are referring to the filesystem, then that's not really an issue. We can store programs in the FS that are bigger than we can execute due to RAM constraints.

markshannon:
I'd rather do the "right thing" here and save flash elsewhere, but we are getting rather low on flash.
I'll leave the single method table for now. 

The text in help.c could be compressed and dropping the remnants of the DAL will help a lot.
Maybe when we have more flash we can revisit this.

I'll open a new PR as I want to keep this branch for possible future use.

Issue #361, opened 2016-10-20
No RSSI measurement on the radio module?


whaleygeek:
I'd like to do received signal strength indication measurements from the radio module on MicroPython on the micro:bit.

There doesn't seem to be anything in the API to retrieve this information, the most I can find from a github search is this single reference:

https://github.com/bbcmicrobit/micropython/blob/a20fbe82c919cea9519ce5a1140103f518b51caa/source/microbit/modradio.cpp#L78

Would it be possible for someone to add a feature to get the RSSI level of the last received radio packet? It seems like the structure record in the above line of code would surface this reading, providing it was then scaled to the same range that PXT offers.

I and others would like to use RSSI measurements with 3 micro:bits to explain how triangulation works in cell towers.

ukBaz:
Sorry, a bit of an aside. Is it triangulation or trilateration you are looking to do with RSSI?
http://circuitcellar.com/ee-tips/triangulation-trilateration-or-multilateration-ee-tip-125/

dpgeorge:
Yes, there is the capability for this, and I actually tested it when developing the radio module, but it never got exposed on the Python side.

Can someone please propose a method name and behaviour for this?

ukBaz:
My suggestion would be that the library present 'distance' to users rather than raw RSSI.
From my experience of doing a few Bluetooth beacon workshops where we worked on proximity, the students never got beyond just using the distance value from the library.

For my workshop I used:
distance = math.pow(10, (rssi - tx_power_at_1m) / (-10 * path_loss_exponent))

Which was based on these articles:
http://matts-soup.blogspot.co.uk/2013/12/finding-distance-from-rssi.html
And section two of:
http://www.intechopen.com/books/emerging-communications-for-wireless-sensor-networks/indoor-location-tracking-using-received-signal-strength-indicator

Just to set peoples expectation, there will be a lot of fluctuation on RSSI readings especially in an indoor environment with students moving around.

My thought on implementation is that it is a characteristic of the receive() function. Maybe something like this:
```
incoming = radio.receive()
if incoming['message'] == 'flash' and incoming['distance'] < 2:
```

However I appreciate that this would break backwards compatibility so maybe you have have to set parameter in `rado.receive(message_only=False)` to get the additional information. 
This would allow the library to expose RSSI and distance.

whaleygeek:
Thanks @dpgeorge 

[...]

PR #362, opened 2016-10-21
Update to radio API proposal (RSSI) as new documentation.

ntoll:
I have tried to distil all the requirements mentioned in #361 and express them in the simplest non-backwards-breaking and Pythonic manner (return a tuple for multi value results).

Please bear in mind:
- Any change should NOT break the existing API.
- Any change should be as simple as possible.
- Any change should work within the conceptual model of the existing API.

I realise naming stuff is hard which is why I opted for `receive_all` - it fits with the naming convention we currently use in the `radio` module, it's easy to type and it does exactly what it says on the tin.

This is a straw man. I expect this to change. Comments, constructive critique and ideas are most welcome. :-)

### DIFF
--- a/docs/radio.rst
+++ b/docs/radio.rst
@@ -9,12 +9,13 @@ networks.
 The radio module is conceptually very simple:
 
 * Broadcast messages are of a certain configurable length (up to 251 bytes).
-* Messages received are read from a queue of configurable size (the larger the queue the more RAM is used). If the queue is full, new messages are ignored.
+* Messages received are read from a queue of configurable size (the larger the queue the more RAM is used). If the queue is full, new messages are ignored. Reading a message removes it from the queue.
 * Messages are broadcast and received on a preselected channel (numbered 0-100).
 * Broadcasts are at a certain level of power - more power means more range.
 * Messages are filtered by address (like a house number) and group (like a named recipient at the specified address).
 * The rate of throughput can be one of three pre-determined settings.
 * Send and receieve bytes to work with arbitrary data.
+* Use `receive_full` to obtain full details about an incoming message: receiving signal strength and the device's `running_time` when the message arrived.
 * As a convenience for children, it's easy to send and receive messages as strings.
 * The default configuration is both sensible and compatible with other platforms that target the BBC micro:bit.
 
@@ -131,6 +132,28 @@ Functions
     the prepended bytes before converting to a string.
 
     A ``ValueError`` exception is raised if conversion to string fails.
+
+.. py:function:: receive_full()
+
+    Returns a tuple containing three values representing the next incoming
+    message on the message queue. If there are no pending messages then
+    ``None`` is returned.
+
+    The three values in the tuple represent:
+
+    * the next incoming message on the message queue as bytes.
+    * the RSSI (signal strength): a value between 0 (strongest) and -255 (weakest) as measured in dBm.
+    * a timestamp: the value returned by `microbit.running_time` when the message was receieved.
+
+    For example::
+
+        details = radio.receive_full()
+        if details:
+            msg, rssi, timestamp = details
+
+    This function is useful for providing information needed for triangulation
+    and/or triliteration with other micro:bit devices.
+
 
 Examples
 --------


docs/radio.rst
```diff
@@ -132,6 +133,27 @@ Functions
 
     A ``ValueError`` exception is raised if conversion to string fails.
 
+.. py:function:: receive_all()
+
+    Returns a tuple containing four values representing the next incoming
+    message on the message queue. If there no pending messages then ``None``
+    is returned.
```
deshipu: And removes that message from the queue.

docs/radio.rst
```diff
@@ -132,6 +133,27 @@ Functions
...
+    * the RSSI signal strength: a value between 0 (strongest) and 255 (weakest).
+    * the approximate distance: equivalent to ``math.pow(10, (rssi - tx_power_at_1m) / (-10 * path_loss_exponent))``
+    * a timestamp: when the signal was receieved.
+
+    For example::
+
+        msg, rssi, distance, timestamp = radio.receive_all()
+
+    This function is useful providing information needed for triangulation
+    and/or triliteration with other micro:bit devices.
```
deshipu: This somehow doesn't sound right grammar-wise, "useful in providing" perhaps?

docs/radio.rst
```diff
@@ -132,6 +133,27 @@ Functions
...
+
+    Returns a tuple containing four values representing the next incoming
+    message on the message queue. If there no pending messages then ``None``
+    is returned.
+
+    The four values in the tuple represent:
+
+    * the next incoming message on the message queue as bytes.
+    * the RSSI signal strength: a value between 0 (strongest) and 255 (weakest).
+    * the approximate distance: equivalent to ``math.pow(10, (rssi - tx_power_at_1m) / (-10 * path_loss_exponent))``
```
deshipu: in what units?

docs/radio.rst
```diff
@@ -132,6 +133,27 @@ Functions
...
+.. py:function:: receive_all()
+
+    Returns a tuple containing four values representing the next incoming
+    message on the message queue. If there no pending messages then ``None``
+    is returned.
+
+    The four values in the tuple represent:
+
+    * the next incoming message on the message queue as bytes.
+    * the RSSI signal strength: a value between 0 (strongest) and 255 (weakest).
```
deshipu: I assume "weakest" is the weakest signal that can still be received by the radio, so I guess no units there...

docs/radio.rst
```diff
@@ -132,6 +133,27 @@ Functions
...
+    * the approximate distance: equivalent to ``math.pow(10, (rssi - tx_power_at_1m) / (-10 * path_loss_exponent))``
+    * a timestamp: when the signal was receieved.
```
deshipu: What format is the timestamp?

deshipu:
This looks very reasonable. A few nitpicks.

docs/radio.rst
dpgeorge: I guess it would be the same result from `microbit.running_time()`.

docs/radio.rst
```diff
@@ -132,6 +133,27 @@ Functions
 
     A ``ValueError`` exception is raised if conversion to string fails.
 
+.. py:function:: receive_all()
```
ukBaz: Would receive_detailed() be clearer than receive_all()?

docs/radio.rst
ukBaz: I suspect the RSSI on this system is in dBm (rather than dB) but I can not find notes anywhere confirming this.
I see on the Lancaster runtime uBit.radio that the Transmission Power has eight user configurable settings from 0 (-30dbm) to 7 (+4dbm).

docs/radio.rst
ukBaz: I would suggest taking this formula out. The reason being that the calculation of distance with RSSI is done by calculating the difference between the power at transmit and the power at receive. Right now the library does not send the details of what power level the data was transmitted at.

docs/radio.rst
ntoll: Not sure I understand. I'm concerned that we can say 0-255 measured as [insert measurement scale here].

docs/radio.rst
ntoll: That's what I had in mind. We should mention this too (that it's the value of `running_time` when the message was received).

docs/radio.rst
ntoll: Yeah... it's a typo. Will fix.

docs/radio.rst
ntoll: I vote -1 because it's too wordy. I believe `receive_all` - while not precise - communicates what's going on and allows kids with underdeveloped typing skills a bit of a break. ;-)

docs/radio.rst
ntoll: Aha. I didn't realise, but now I see the tx_foo. I'll remove it when I revise (later today).

However, such a formula would be a fun exercise. :-) We should put something like it into our documentation.

docs/radio.rst
ntoll: Hmm... could this be somewhere in the docs for the NRF chip?

docs/radio.rst
deshipu: How about `receive_full`?

whaleygeek:
Ok, so the return value from the DAL is -dBm so 0 is 0dBm and 255 is -255dBm, as per jfinney comments here:

https://github.com/lancaster-university/microbit-dal/issues/226

whaleygeek:
I agree with @ukBaz about the distance. I think we should leave that to a module of example code, then the application at the transmitting end can put in the tx power setting, and the application at the receiving end can receive it and do the maths. It would make a nice tutorial.

Also, I think the maths while it looks correct, is a bit suspect in practice due to multipath interference and obstacles. My line-of-sight field testing showed I got a high signal strength for, say, 30 meters, but put a person in the way and the signal strength drops rapidly - so any estimation of distance based on RSSI will in practice be wildly inaccurate and misleading.

ukBaz:
@whaleygeek you are correct that the formula I gave was for signal loss in free space which typically is not what you get in a busy classroom.

For reference there is this article on the challenges of using RSSI for location.
https://www.beaconzone.co.uk/bluetoothleposition

docs/radio.rst
ntoll: So, my `dict` command tells me:

```
  61 Moby Thesaurus words for "details":
     brass tacks, bric-a-brac, cameo, catalog, cataloging, chaff,
     character, character sketch, characterization, chicken feed,
     chickenshit, delineation, depiction, description, essential facts,
     essentials, evocation, froth, gimcrackery, graphic account, image,
     imagery, impression, itemization, knickknackery, limning, minutiae,
     particularization, peanuts, photograph, picture, portrait,
     portraiture, portrayal, profile, rendering, rendition,
     representation, rubbish, sketch, small beer, small change,
     specification, the data, the details, the dope, the facts,
     the information, the particulars, the picture, the scoop,
     the score, the specifics, the whole story, trash, trifles, trivia,
     trumpery, vignette, vivid description, word painting
```

I've been racking my brain to think of something that's a simple alternative that'll make sense to kids and be easy to type. @deshipu I totally get your point and I think it a good one. Not sure we'll get a better alternative though. Perhaps a good alternative is that it's something we should highlight in the docs (see pending changes landing in a moment).

dpgeorge: `receive_with_info`?  Otherwise I like `receive_full`.

ntoll:
So my remaining concern is that I would have expected dBm values rather than -dBm as the RSSI value. I guess this is trumped by consistency with the DAL.

dpgeorge:
> So my remaining concern is that I would have expected dBm values rather than -dBm as the RSSI value. I guess this is trumped by consistency with the DAL.

If you look at the DAL issue that @whaleygeek links to above you'll see that @finneyj is supportive of changing the sign of the RSSI value so that it is indeed dBm (ie values are from -255 up to 0).  So we could go with that hoping that the DAL is also (eventually) changed.

ntoll:
@dpgeorge aha... I missed that (I skimmed Joe's reply, I'm at "real work" and multitasking). +1 on what you suggest. :-)

finneyj:
@dpgeorge @ntoll Just to record here that we'll be moving microbit-dal to return accurate -dbm values in the near future, so this should work out nicely.

ntoll:
@finneyj stonking! I was going to ask why it was like that, but no point now. :-)

ntoll:
OK, I've made a couple of minor changes to make the language clearer and rebased into a single commit.

Is everyone happy with the description of the API? Then someone could try implementing it given this specification. :-)

whaleygeek:
Should 'when the signal was received' be 'when the message was received'?

whaleygeek:
RSSI stands for 'received signal strength indication' so really should we change:

the RSSI signal strength

to 

the RSSI (signal strength)

i.e. it is not the RSSI signal strength (receive signal strength indication signal strength)

This also has the benefit of still (mostly) defining what the RSSI term means, as many readers won't know the acronym.

whaleygeek:
The function heading says 'receive_full' but the inline example says 'receive_all' - which are you settling on?

whaleygeek:
It would be useful to put a note somewhere in the PR, not sure where, but to say that we would like contributions in the examples section and in the tutorials, of doing something useful with this.

I can think of a couple of nice examples we could work up between us:

1. something a bit like @jamesadevine DAL demo of the heart on the screen that gets brighter as the micro:bits are closer, to show the concept working in a really simple way.

2. an example or tutorial where the application includes the tx-power in the payload, and uses @ukBaz example calculation to estimate distance. As an alternative to that, if the application knows the distance, perhaps it could instead work out 'amount of obstacles in the way' - e.g. the London Underground have sensors to tell them how full the stations are. So, if the RSSI from a beacon (or known beacons) at known distances reduces, this would indicate more obstacles in the way (e.g. how full is the playground, how full is the classroom).

3. an application that demonstrates how cell tower handovers work in mobile telecoms (uses the txpower and rssi to estimate radio channel performance, and negotiates a transfer to a different channel with the transmitter).

ntoll:
@whaleygeek fixed and thanks. Mainly due to my terrible editing skills.

Once the function is implemented we can add examples (your suggestions are great!).

docs/radio.rst
```diff
@@ -132,6 +133,26 @@ Functions
...
+
+    The three values in the tuple represent:
+
+    * the next incoming message on the message queue as bytes.
+    * the RSSI (signal strength): a value between 0 (strongest) and -255 (weakest) as measured in dBm.
+    * a timestamp: the value returned by `microbit.current_time` when the message was receieved.
+
+    For example::
+
+        msg, rssi, timestamp = radio.receive_full()
```
dpgeorge: This is not a good example if the method is going to return None when no messages are available.  Either receive_full should return a 3-tuple of None's indicating no message, or this example should assign to just a single variable.

ntoll: Doh.... good point (entirely my fault). I'll update.

markshannon:
Given the lack of precision of the underlying timer, is the timestamp worth it?
I don't see that it adds anything that calling `microbit.running_time()` immediately after `radio.receive_full()` doesn't provide.

dpgeorge:
> Given the lack of precision of the underlying timer, is the timestamp worth it?
> I don't see that it adds anything that calling microbit.running_time() immediately after radio.receive_full() doesn't provide.

Since the packets are buffered in a queue, if there's more than 1 outstanding one then it makes sense to store the time stamp with it.

DavidWhaleMEF:
Ping - is this ready to be deployed yet, looks like it has stalled somewhere? I have a number of users with applications where accessing RSSI would make those applications possible to implement (they are impossible to implement at the moment without this feature). Thanks.

markshannon:
Could we remove the timestamp from this PR, so that we can merge #410?
The main point of this is to add signal strength detection. 
IMO  the timestamp is unlikely to be useful, and not worth the cost, given the relatively large jitter in `running_time()`.

dpgeorge:
> IMO the timestamp is unlikely to be useful, and not worth the cost,

Yes, I agree, it's relatively costly in RAM (4 bytes per packet in the queue).  So it can go.

DavidWhaleMEF:
I'd be happy with a partial timestamp, so we can at least get a small delta between the receive times. It doesn't need the full resolution of a 4 byte timestamp. I reckon we could pack just enough time data in 1 byte per packet, with 255 indicating the timer overflowed. Mostly your python will get round quick enough to receive a packet within a second or two (if it is doing other stuff), but it's the fine timing that would be really useful to enable other simple location based use cases to be explored.

DavidWhaleMEF:
The nordic radio system offers radio packets to the language via DMA transfer, so it performs really well at the lowest level, so I think we could get good value from a single byte small delta between payloads. Apps that want more range on receive times can do their own timing at the packet consumption end, but the use case is if the queue has 10 messages in it because they come in quicker than the application is polling, some indication of time between packets will help. I'm happy for the maximum range of this timer to be 1 second or less. e.g. 1 second split into 255 segments gives a time resolution of about 4ms - clearly this needs to be matched to the fastest possible packet-to-queue processing overhead.

dpgeorge:
Now that radio.receive_full() is implemented I merged this in aa4cbd933c3811c291388433ac4cea836a39fce5 with some minor tweaks (notably that the timestamp is in microseconds, not milliseconds).

PR #369, opened 2016-10-29
Add receive_into function to radio module. Allows allocation-free radio reception.

markshannon:
To transmit data at higher speeds, predictable delays are needed and that means no GC.
This PR adds a `receive_bytes_into` function allowing packets to be copied into a pre-existing buffer.

For example, this allows wireless audio:

``` python
import audio
import radio
f = audio.AudioFrame()
radio.on()
def radio_stream():
    while True:
        if radio.receive_bytes_into(f) == 32:
            yield f
audio.play(radio_stream())
```

Also includes a couple of minor fixes for the AudioFrame class.


dpgeorge:
It looks good, I'm all for reducing heap allocations where possible.

Did you come across the need for this only from trying to do audio over radio, or where there other use-cases?

Also, how will it combine with #362, would we have a `radio.receive_full_into()` method?

markshannon:
So far only audio; I'm planning on making a walky-talky demo, as part of my `audio.record()` implementation.

`radio.receive_full_into()` sounds like overkill to me. `radio.receive_full` is useful for proximity sensing, which can be done at low speeds and with soft real-time constraints. `radio.receive_into` is useful when there are hard real-time constraints.

source/microbit/modaudio.cpp
```diff
@@ -620,7 +620,7 @@ const mp_obj_type_t microbit_audio_frame_type = {
     .buffer_p = { .get_buffer = audio_frame_get_buffer },
     .stream_p = NULL,
     .bases_tuple = NULL,
-    .locals_dict = (mp_obj_dict_t*)&microbit_audio_frame_locals_dict_table,
+    .locals_dict = (mp_obj_dict_t*)&microbit_audio_frame_locals_dict,
```
dpgeorge: What is this change?

source/microbit/modaudio.cpp
```diff
@@ -526,11 +526,11 @@ static microbit_audio_frame_obj_t *copy(microbit_audio_frame_obj_t *self) {
...
-    if (mp_obj_get_type(other) != &microbit_audio_frame_type) {
-        nlr_raise(mp_obj_new_exception_msg(&mp_type_TypeError, "Must be an AudioBuffer"));
-    }
-    for (int i = 0; i < AUDIO_CHUNK_SIZE; i++) {
-        self->data[i] = ((microbit_audio_frame_obj_t *)other)->data[i];
+    mp_buffer_info_t bufinfo;
+    mp_get_buffer_raise(other, &bufinfo, MP_BUFFER_READ);
+    uint32_t len = bufinfo.len > AUDIO_CHUNK_SIZE ? AUDIO_CHUNK_SIZE : bufinfo.len;
+    for (uint32_t i = 0; i < len; i++) {
+        self->data[i] = ((uint8_t *)bufinfo.buf)[i];
```
dpgeorge: So the idea here is to make it more general, to allow "other" to be any buffer-like object?

source/microbit/modradio.cpp
```diff
@@ -444,10 +449,17 @@ STATIC mp_obj_t mod_radio_send(mp_obj_t buf_in) {
 MP_DEFINE_CONST_FUN_OBJ_1(mod_radio_send_obj, mod_radio_send);
 
 STATIC mp_obj_t mod_radio_receive(void) {
-    return radio_receive(true);
+    return radio_receive(true, NULL);
 }
 MP_DEFINE_CONST_FUN_OBJ_0(mod_radio_receive_obj, mod_radio_receive);
 
+STATIC mp_obj_t mod_radio_receive_into(mp_obj_t buf_in) {
```
dpgeorge: Shouldn't it be called receive_bytes_into?

source/microbit/modaudio.cpp
markshannon: Yes. I can't see any reason not to and it makes the example Python cleaner.

source/microbit/modaudio.cpp
markshannon: `.local_dict` should point to a dict, not the table for a dict.

source/microbit/modradio.cpp
markshannon: I guess that would be clearer. 
Sure we can afford the extra 5 bytes? :wink:

markshannon:
@dpgeorge I've split this PR into 2 commits as asked and changed `receive_into` into `receive_bytes_into` as suggested.

tests/radio_audio.py
```diff
@@ -0,0 +1,74 @@
...
+def play():
+    display.clear()
+    radio.on()
+    radio.config(channel=90, queue=12)
+    count = -1
+    def gen():
+        recvd = audio.AudioFrame()
+        empty = audio.AudioFrame()
+        while True:
+            if radio.receive_into(recvd) == 32:
```
dpgeorge: This should be receive_bytes_into().

markshannon:
Done.

Issue #117, opened 2015-11-29
Higher level I²C interface


deshipu:
This is not a critical issue, I suppose, but I thought I will report it so that I don't forget about it.

Our I²C interface only has two commands, read bytes and write bytes. While this allows us to do everything that is needed (especially since we have control over the stop of transaction), it's a very low-level interface. If we want this to be actually used, it would be nice to have a higher level interface.

The most popular Python library for doing I²C communication is the `smbus` module. It's commonly used on the Raspberry Pi, and there are many tutorials online and examples of communicating with various devices using it. I think that if we mirrored the interface of that library (which in turn mirrors the interface of the C library that it wraps), we could count on reusing all of those examples, which is a big advantage.

I couldn't find any proper documentation of the library, but the docstrings in the code are explanatory:
https://github.com/bivab/smbus-cffi/blob/master/smbus/smbus.py

carlosperate:
Have the toch develop guys already implemented I2C/SPI/UART? While I do like the idea of having the same API as popular library, it's worth considering what the users might get familiarised with first while using the other high level microbit platforms.

markshannon:
smbus looks like a lot of code for our tiny device. 
Anyone volunteering to add a cut-down version to the `examples/` folder?

deshipu:
After some discussions about the `machine.I2C` interface in the upstream micropython, I think it would be good to implement an "i2c endpoint" class for this. I will take a stab at this.

deshipu:
Note that this will have an additional advantage -- any drivers created for regular micropython i2c devices will work without modification on the microbit.

deshipu:
Since the specification for the i2c endpoints is still a little unclear, I went and implement the SMBus commands after all. They should be familiar for many from the Raspberry Pi.

PR #376, opened 2016-11-07
examples/smbus.py: SMBus commands

deshipu:
Adds an example class which implements the basic SMBus commands, as
present in the SMBus Python library for the Raspberry Pi.

Note that Block Read and Block Process call transactions should be
done with just two repeated starts, not three, but there is no way
to do it with the current microbit I2C API.

Implements #117

### DIFF
--- a/examples/smbus.py
+++ b/examples/smbus.py
@@ -1,0 +1,75 @@
+class SMBus:
+    """Implement the basic SMBus commands."""
+
+    def __init__(self, i2c, addr, pec=False):
+        if pec:
+            raise NotImplementedError("PEC not supported")
+        self.i2c = i2c
+        self.addr = addr
+
+    def write_quick(self):
+        """Perform a QuickCommand write."""
+        return self.i2c.write(self.addr, b'')
+
+    def read_byte(self):
+        """Perform a Receive Byte."""
+        return self.i2c.read(self.addr, 1)[0]
+
+    def write_byte(self, val):
+        """Perform a Send Byte."""
+        return self.i2c.write(self.addr, val.to_bytes(1))
+
+    def read_byte_data(self, cmd):
+        """Perform a Read Byte."""
+        self.i2c.write(self.addr, cmd.to_bytes(1), repeat=True)
+        return int.from_bytes(self.i2c.read(self.addr, 1))
+
+    def write_byte_data(self, cmd, val):
+        """Perform a Write Byte."""
+        self.i2c.write(self.addr, cmd.to_bytes(1) + val.to_bytes(1))
+
+    def read_word_data(self, cmd):
+        """Perform a Read Word."""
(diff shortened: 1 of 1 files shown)


markshannon:
OOI, what would need to be added to the i2c API in order to do Block Read and Block Process properly?

deshipu:
I would need to be able to make a read without sending a start condition (or a stop condition), so that I can read the number of bytes to read, before actually reading those bytes.

deshipu:
Basically something like:
```
self.i2c.write(self.addr, cmd.to_bytes(1) + count.to_bytes(1) + vals, repeat=True)
count = int.from_bytes(self.i2c.read(self.addr, 1, repeat=True))
return self.i2c.read(None, count)
```

deshipu:
If I could also do a similar write, the expensive bytestring adding operation could be also replaced with three writes.

dpgeorge:
> Note that Block Read and Block Process call transactions should be
> done with just two repeated starts, not three, but there is no way
> to do it with the current microbit I2C API.

I think you can do it by reading the maximum (32 bytes) and then truncating the result at the end.  Eg:
```python
    def read_block_data(self, cmd):
        """Perform a Block Read."""
        self.i2c.write(self.addr, cmd.to_bytes(1), repeat=True)
        data = self.i2c.read(self.addr, 32)
        return data[1 : (1 + data[0])]
```
The slave has no say as to how many bytes the master reads from it.  The master keeps reading until it sends a nack.

deshipu:
Interesting, I will test that, thanks.

abencomo:
Will this allow me to use the Pyboard with this VL53L0X sensor?

https://www.tindie.com/products/onehorse/vl53l0x-time-of-flight-ranging-sensor/

The interface is I2C, but the pyboard complains when I try to do "import smbus" and "import struct"

https://github.com/popunder/VL53L0X/blob/master/VL53L0Xtest.py

deshipu:
@abencomo it is certainly possible to talk to that sensor from MicroBit, however, the library that you are trying was most likely written for Raspberry Pi, which has much more resources available (both in terms of memory and cpu speed). It is probably best to write a dedicated driver for MicroPython, that will take the limited resources into account. I wrote a driver for a very similar sensor, https://bitbucket.org/thesheep/micropython-vl6180/src/tip/vl6180.py -- you might be able to adapt it for your use using the information from the datasheet for your sensor.

Unfortunately I don't have a VL53L0X to test.

On another note, the "struct" module is available (in a limited form) on the pyboard as "ustruct". Unfortunately, "smbus" module is written specially for the microbit and will not work on pyboard.

abencomo:
Thanks Radomir.  I'll try to adapt the vl6180.py and let you know if I got it working with VL53L0X sensor.

dbrgn:
Any news on smbus support? :)

Would be great to have a `smbus` shim available that would make it possible to share driver code for the Raspberry Pi and for MicroPython boards.

PR #377, opened 2016-11-08
Add i2c.init to change the pins and frequency

### LINKED ISSUE #375: I2C on pins 0, 1 or 2
I think that i2c is a great protocol for communicating with peripherals, and it's especially nice for the microbit, where you only have very limited number of pins easily accessible -- so the ability to connect hundreds of devices with just two pins is very good. However, the pins for the hardware i2c on the microbit are in a very awkward place, and they are very hard to use in any project that doesn't utilize the bulky edge connector (and who wants to leave their edge connector in their projects, too). So I'm wondering about some way of doing i2c on the pins that are easily accessible.

The obvious way would be to have software (bit-banging) i2c, working on any pins. The upstream micropython has an implementation of this that works reasonably well -- I wonder how much sense it would make to make this available in the microbit? It will cost some flash memory, certainly, but it will make it much easier to use things like the 8x8 LED matrix backpacks, OLED displays, servo and motor controllers and sensors.

Should I try and get this to work?

deshipu:
Fixes #375


docs/i2c.rst
```diff
@@ -21,6 +21,17 @@ should be described separately in that device's documentation.
 Functions
 =========
 
+.. py:function:: init(freq=None, sda=None, scl=None)
```
dpgeorge: This is different to spi.init: the spi always sets all parameters and has defaults for everything, so that calling spi.init() resets everything to the defaults.  I'd say i2c.init should follow that behaviour.

docs/i2c.rst
```diff
@@ -21,6 +21,17 @@ should be described separately in that device's documentation.
...
+
+    Re-initialize peripheral with the specified frequency ``freq`` on the
+    specified ``sda`` and ``scl`` pins. If pins are not specified, ``pin20``
+    and ``pin19`` are used. If frequency is not specified, it's not changed.
+
+    .. warning::
+
+        Changing the I²C pins from defaults will make the accelerometer and
+        compass stop working, as they are connected internally to those pins.
```
dpgeorge: There are actually 2 peripherals which can act in either I2C or SPI mode.  One of those periphs are configured as I2C for compass and accelerometer.  The other is uninitialised and will be used in SPI mode if the user ever calls spi.init().

So, while it's possible to have 2 I2C periphs (eg 1 for accel/compass, another for the user) I think it's simpler to have one peripheral reserved for I2C and the other for SPI.  This maps to the current microbit module where there's 1 i2c object and 1 spi object.

The alternative is to allow the user to specify which periph they want and in what mode and with what pins.  But that will really complicate the API.

docs/i2c.rst
dpgeorge: We should also discuss if freq is the right name to use, cf baudrate for spi.

source/microbit/microbiti2c.cpp
```diff
@@ -25,17 +25,69 @@
...
+    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
+    mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
+
+
+    PinName p_sda = MICROBIT_PIN_SDA;
+    PinName p_scl = MICROBIT_PIN_SCL;
+
+    // set sda/scl pins if they are given
+    if (args[1].u_obj != mp_const_none || args[2].u_obj != mp_const_none) {
+        if (args[1].u_obj == mp_const_none) {
```
dpgeorge: This should be !=

source/microbit/microbiti2c.cpp
```diff
@@ -25,17 +25,69 @@
...
+            p_sda = microbit_obj_get_pin_name(args[1].u_obj);
+        }
+        if (args[2].u_obj == mp_const_none) {
+            p_scl = microbit_obj_get_pin_name(args[2].u_obj);
+        }
+        ((mp_I2C*)self->i2c)->set_pins(p_sda, p_scl);
```
dpgeorge: This line should be outside the "if" so the pins can be reset if not specified.

docs/i2c.rst
deshipu: I went for the simplest change and tried to make it consistent with `spi.init` and `uart.init`.

Having two i2c's could be nice, but I can't think of an api for this that would both be clear and stay compatible with what is there so far (and with other languages on the micro:bit).

docs/i2c.rst
deshipu: Updated to use the defaults when no value is specified.

I used "freq" because:

- the `machine` API uses that on other micropython ports
- the mbed internals call it "frequency" internally
- "frequency" seems to be the standard term used anywhere when talking about I2C

Of course changing it to "baudrate" is trivial to do, if needed.

deshipu:
Copying the comments here, as they got hidden by the new commit:

>> We should also discuss if freq is the right name to use, cf baudrate for spi.
￼  ￼  ￼￼
> I used "freq" because:
>
> - the machine API uses that on other micropython ports
> - the mbed internals call it "frequency" internally
> - "frequency" seems to be the standard term used anywhere when talking about I2C
> - Of course changing it to "baudrate" is trivial to do, if needed.

deshipu:
To be honest, I'm perfectly fine with either `baudrate` or `frequency`, even when the former is not technically right, due to clock stretching and multi-master arbitration. However, it does save a few bytes, and makes the i2c, spi and uart init method look more consistent...

dpgeorge:
This PR looks good now.

I'd say it's better to be consistent with the usual I2C terminology (ie use freq) rather than consistent with the other APIs like spi and uart.

@ntoll @markshannon please give your vote/opinion on using baudrate vs freq in the i2c init method, thanks!

ntoll:
It's a +1 on `freq` for the reasons listed in the fragment @deshipu copied into the thread. :+1:

markshannon:
I don't have a strong opinion about `freq/baudrate`, but I do prefer `freq[uency]`

source/microbit/microbiti2c.cpp
```diff
@@ -25,17 +25,62 @@
...
+    // parse args
+    microbit_i2c_obj_t *self = (microbit_i2c_obj_t*)pos_args[0];
+    mp_arg_val_t args[MP_ARRAY_SIZE(allowed_args)];
+    mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(allowed_args), allowed_args, args);
+
+
+    PinName p_sda = MICROBIT_PIN_SDA;
+    PinName p_scl = MICROBIT_PIN_SCL;
+
+    if (args[1].u_obj == mp_const_none) {
```
dpgeorge: The == here should be a !=

source/microbit/microbiti2c.cpp
```diff
@@ -25,17 +25,62 @@
...
+        p_sda = microbit_obj_get_pin_name(args[1].u_obj);
+    }
+    if (args[2].u_obj == mp_const_none) {
```
dpgeorge: Should be !=

deshipu:
Fixed and squashed.

dpgeorge:
Great!

PR #382, opened 2016-12-09
Add details of set_pull method and default pull states.

### LINKED ISSUE #381: Docs do not list the set_pull methods
The docs here:

https://microbit-micropython.readthedocs.io/en/latest/pin.html#pin-functions

do not list the set_pull() method.

From python-microbit mailing list...

Hi David,

Looking at the code the default pull states seem to be:

read_analog: configured by mbed analogin_init() function, looks like
there is no pull
read_digital: PULL_DOWN
is_touched: PULL_UP

To change the value, use: pin.set_pull(pin.PULL_UP), or PULL_DOWN or NO_PULL.

ntoll:
Fixes #381

### DIFF
--- a/docs/pin.rst
+++ b/docs/pin.rst
@@ -116,9 +116,9 @@ Classes
 =======
 
 There are three kinds of pins, differing in what is available for them. They
-are represented by the below classes. Note that they form a hierarchy, so that
-each class has all the functionality of the previous class, and adds its own
-to that.
+are represented by the classes listed below. Note that they form a hierarchy,
+so that each class has all the functionality of the previous class, and adds
+its own to that.
 
 .. note::
     Those classes are not actually available for the user, you can't create
@@ -134,6 +134,12 @@ to that.
     .. py:method:: write_digital(value)
 
         Set the pin to high if ``value`` is 1, or to low, if it is 0.
+
+    .. py:method::set_pull(value)
+
+        Set the pull state to one of three possible values: ``pin.PULL_UP``,
+        ``pin.PULL_DOWN`` or ``pin.NO_PULL`` (where ``pin`` is an instance of
+        a pin). See below for discussion of default pull states.
 
 .. py:class:: MicroBitAnalogDigitalPin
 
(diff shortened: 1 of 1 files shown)


docs/pin.rst
```diff
@@ -170,3 +176,10 @@ to that.
         whatever is connected to it. Human body has quite a large capacitance,
         so touching the pin gives a dramatic change in reading, which can be
         detected.
+
+
+The default pullup/down states for each type of pin are as follows:
+
+* ``read_analog`` - there is no pull.
+* ``read_digital`` - ``pin.PULL_DOWN``
+* ``is_touched`` - ``pin.PULL_UP``
```
dpgeorge: It's a bit confusing listing it like this.  First, read_analog/read_digital/is_touched are not pins but rather states of a pin (ie a particular mode).  Second, you can't actually change the pull for any mode except read_digital.

So I'd suggest some wording like this: "The pull mode for a pin is automatically configured when the pin changes to an input mode.  Input modes are when you call read_analog/read_digital/is_touched.  The pull mode for these is, respectively, NO_PULL, PULL_DOWN, PULL_UP.  Only when in read_analog mode can you call set_pull to change the pull mode from the default."

ntoll:
Fixed and rebased.

whaleygeek:
Also note that the micro:bit has external weak (10M) pull-ups fitted on P0 P1 and P2 only, in order for the touch sensing to work. See the edge connector data sheet here: http://tech.microbit.org/hardware/edgeconnector_ds/

docs/pin.rst
```diff
@@ -170,3 +176,9 @@ to that.
         whatever is connected to it. Human body has quite a large capacitance,
         so touching the pin gives a dramatic change in reading, which can be
         detected.
+
+The pull mode for a pin is automatically configured when the pin changes to an
+input mode. Input modes are when you call ``read_analog`` / ``read_digital`` /
+``is_touched``. The pull mode for these is, respectively, ``NO_PULL``,
+``PULL_DOWN``, ``PULL_UP``. Only when in ``read_analog`` mode can you call
```
dpgeorge: Sorry, my bad, read_analog should be read_digital on this line.

ntoll:
I think it 's probably worth adding @whaleygeek's comment somewhere in this documentation too.

dpgeorge:
Thanks @ntoll!  Good that you added the link to the edge-connector details.

PR #385, opened 2016-12-19
Fix touch sensing for pins.

markshannon:
Use NO_PULL for touch pins.
The internal 11k resistor is a better conductor than I am, so `pin0.is_touched()` always returns false.
Using the external 1MΩ means that a sufficiently firm press while earthed does work (just).

http://tech.microbit.org/hardware/schematic/

### DIFF
--- a/source/microbit/microbitbutton.cpp
+++ b/source/microbit/microbitbutton.cpp
@@ -46,7 +46,7 @@ static bool debounced_high[8] = { true, true, true, true, true, true, true, true
 mp_obj_t microbit_button_is_pressed(mp_obj_t self_in) {
     microbit_button_obj_t *self = (microbit_button_obj_t*)self_in;
     /* Button is pressed if pin is low */
-    return mp_obj_new_bool(!debounced_high[self->pin->number]);
+    return mp_obj_new_bool(!debounced_high[self->pin->number&7]);
 }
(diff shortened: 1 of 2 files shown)


deshipu:
I thought the touch detection is based on capacitance, and not resistance? Then you need a resistor to discharge the capacitor?

markshannon:
It's just resistance. We detect whether the (debounced) pin input is low or high.
Essentially it is the same as the buttons, but with higher resistances.

whaleygeek:
Confirmed. It is resistive touch sense, which is why you have to also hold
the GND pin.

There is a 10 meg external pullup fitted on P0 P1 P2 specifically for that
purpose, as detailed on the tech info pages here:

http://tech.microbit.org/hardware/edgeconnector_ds/

The standard DAL wraps a debounced button class around these pins and turns
off any internal pull resistors when touch mode is enabled on the pad. The
P pads are therefore pulled high with a very weak pullup, and connecting
them via the human body to gnd will circumvent that pullup just enough for
it to fall below the pin VIL (voltage input low) threshold and sense as a
low.

David

dpgeorge:
I tested it before and after this patch and indeed this patch does fix touch!

carlosperate:
do we support `is_touched` on pins other than 0, 1, and 2?

ntoll:
@carlosperate no.

Issue #388, opened 2016-12-21
clear() method for Image objects


jbenc:
When using an Image object as an off-screen buffer (see #387), it would be useful to recycle the same buffer instead of creating a new object many times a second (granted, the garbage collector seems to do a good job but still...). A clear() method that sets all pixels to 0 (or, perhaps, to a specified value) would come very handy for this.

dpgeorge:
The idea is that you would create a new Image(5,5) for each frame.  But considering that images are in fact mutable and have the set_pixel method, I'd say a fill method would be a very sensible addition.  Eg:
```
img.fill(0) # clear the entire image
img.fill(5) # set the image half-bright
```

dpgeorge:
The fill() method is implemented by #390.

PR #390, opened 2016-12-30
Make image created with no arguments mutable and add fill method.

### LINKED ISSUE #387: Usage of Image() without parameters
Currently, Image() without parameters creates an empty immutable image. This seems not to be that useful behavior; displaying such image is equivalent to calling display.clear().
When programming simple games, it's necessary to draw to an off-screen buffer and display the buffer with display.show(buffer) when done to avoid flickering. A logical way to write this is:

buffer = Image()
...do what is necessary using buffer.set_pixel()...
display.show(buffer)

However, this raises an exception. The following code works as intended:

buffer = Image(5, 5)
...do what is necessary using buffer.set_pixel()...
display.show(buffer)

Given that the first snippet looks very natural (and is what I tried first), I wonder whether it wouldn't be better if the default width and height were 5. Took a bit of effort (not significant, though) to explain to kids that they need to write Image(5, 5).

markshannon:
This now works
```python
i = Image()
i.set_pixel(2,2,9)
```
The fill method sets all the pixels to the same value

Fixes #387 and #388


markshannon:
Done

docs/image.rst
```diff
@@ -100,6 +100,13 @@ Classes
         Return a new image by inverting the brightness of the pixels in the
         source image.
 
+    .. py:method:: fill(value)
+
+        Set the brightness of all the pixels in the image to the
+        ``value``, which has to be between 0 (dark) and 9 (bright).
+
+        This method will raise an exception when called on any of the build-in
```
dpgeorge: Shouldn't it be "built-in"? (note t)

markshannon: The perils of copy-and-paste. Fixed.

dpgeorge:
Great!

PR #397, opened 2017-01-06
Change default pin states to allow for digital use.

ntoll:
This is probably a bodge. *PLEASE CHECK*

When using a build of the latest master.  I try this in the REPL:

```
>>> from microbit import *
>>> pin14.write_digital(1)
```

I *should* be able to use pin14 as a digital pin (I've been sent a robotics kit that expects to be able to use `pin14` in this manner and our docs appear to say so too - see that table here: https://microbit-micropython.readthedocs.io/en/latest/pin.html). 

But instead I get:

```
ValueError: Pin 14 in spi mode
```

Looking at `microbitpin.cpp` I notice that the `microbit_pin_write_digital` function attempts to acquire the pin in write digital mode. Unfortunately, I the `release` function assigned to pins in SPI mode appears to be `mode_error`. Hence the traceback we get above.

Put simply you can't transition out of the SPI state. This PR attempts to *bodge* the simplest solution by putting the SPI and I2C pins into a state from which one can transition.

*My concern is there's no way for the pins to get into SPI of I2C mode* that I can find. I guess this should happen when / if those modules are initialised with either referenced or default pins. However, a cursory look at the source appears to suggest such modes are not set against the pins even when non-default pins are used.

I understand why we probably don't want a state transition from spi to something else: you've got some sort of hardware necessarily plugged in. However, I guess if you don't you should be able to use the pins as their capabilities allow. This perhaps also applies to pins for buttons and display columns.

I don't expect this PR to be merged but it gives us two things:

1) The robot kit works (I've tried it).
2) It's somewhere for a discussion of the above.

Thoughts..? @markshannon @dpgeorge

### DIFF
--- a/source/microbit/microbitpin.cpp
+++ b/source/microbit/microbitpin.cpp
@@ -50,14 +50,14 @@ static const qstr initial_modes[] = {
     MP_QSTR_display, /* pin 10 */
     MP_QSTR_button,  /* pin 11 - button B */
     MP_QSTR_unused,  /* pin 12 */
-    MP_QSTR_spi,     /* pin 13 */
-    MP_QSTR_spi,     /* pin 14 */
-    MP_QSTR_spi,     /* pin 15 */
+    MP_QSTR_unused,  /* pin 13 */
+    MP_QSTR_unused,  /* pin 14 */
+    MP_QSTR_unused,  /* pin 15 */
     MP_QSTR_unused,  /* pin 16 */
     MP_QSTR_3v,      /* pin 17 - 3V */
     MP_QSTR_3v,      /* pin 18 - 3V */
-    MP_QSTR_i2c,     /* pin 19 */
-    MP_QSTR_i2c,     /* pin 20 */
+    MP_QSTR_unused,  /* pin 19 */
+    MP_QSTR_unused,  /* pin 20 */
     MP_QSTR_stop     /* Sentinel value to mark end of array */
 };
 


markshannon:
"This is probably a bodge". It is :smile: 
We should add an `off` method to release the pins, instead.

dpgeorge:
> We should add an off method to release the pins, instead.

Yes, or deinit() to match init().  So that'd be i2c.deinit() and spi.deinit(), which would release the pins.

But I do agree with the part of @ntoll's patch that makes the SPI pins unused by default: this is backwards compatible behaviour (back to the time before pin acquire code was introduced).  I2C pins are already in I2C mode on boot-up (due to the accelerometer and compass), but SPI pins need not be used for SPI by default.

So I'd suggest having the SPI pins unused by default, and then when/if any SPI method is used then the pins are acquired (if they are already not acquired by SPI).

markshannon:
Allowing arbitrary reading or writing to the SPI pins has the potential to damage hardware, so I think it should raise an exception if there is any chance that the pins are in SPI mode.
@dpgeorge's suggestion for the spi methods to lazily acquire the pins, sounds like the best idea.

I can't say that I like "deinit" much, I'd prefer "off".

Any suggestions for cheap hardware to test it with?

deshipu:
I think memory chips are as cheap as they get, especially the small ones.

ntoll:
So I've got reports that others have run into the same problem with pin states. I've asked them to add details to this PR. Would be good to get this fixed then I can update the firmware in Mu, uFlash and the browser based editor.

obrasier:
We had the same problem with the pin states, not with SPI but with the button pins.

After flashing to the latest firmware (as of 2017-01-06)

`pin5.read_digital()`

Throws an error as it is a button pin and can't be set to digital mode. Whereas in the old firmware this was allowed. Sorry I can't give you a specific error message right now, I don't have a windows machine to run the build handy. Our version (windows) is here though: http://ncss.edu.au/embedded

Additionally, for the pin mode, users found it confusing that they had to do a `pin0.read_digital()` in order to set the pin mode to digital implicitly, in order to then use:

`pin0.set_pull(pin0.NO_PULL)`

I think a good solution would be to add a `set_mode()` in addition to the `get_mode()` function that exists already. Just so it is more explicit in how to set the mode of each pin.

dpgeorge:
> I think a good solution would be to add a set_mode() in addition to the get_mode() function that exists already.

We definitely need to fix these problems with pin modes.  But note that the original idea of the pin API was that it should be very easy to use and not require explicit mode setting.  Users should be able to do `pin5.read_digital()` and expect it to "just work".

obrasier:
> Users should be able to do pin5.read_digital() and expect it to "just work".

Definitely, however to set the internal pull up resistors _requires_ the pin mode to be set, and it is not necessarily intuitive to have to do a dummy `digital_read()` in order to set the required pin mode.

microbit-carlos:
As https://github.com/bbcmicrobit/micropython/pull/407 this one has been superseded by e04006d67764da0e12251f9621d3254b7fe9b23a.

PR #399, opened 2017-01-11
Fix ValueError exception that occurs when playing a second song.

### LINKED ISSUE #398: ValueError: pin0 is in music mode
With code that looks like the following, we have occasionally seen `ValueError: pin0 is in music mode`:

```python
while True:
  music.play(music.NYAN, wait=False)
```

tsukasa-au:
We now explicitly stop the current song that is playing before
attempting to start a new song.

Fixes #398


markshannon:
LGTM.
Confirmed that this works.

dpgeorge:
This doesn't look correct to me.  It frees the pin that is passed to play, instead of the pin that was passed to the previous invocation of play (ie the pin that's currently playing music, if any).

Issue #395, opened 2017-01-03
Memory-less Image.crop()


deshipu:
I recently gave an example for how to make an image scroll across the display at http://forum.micropython.org/viewtopic.php?f=2&t=2833&p=16840 but the code I wrote, which I thought was intuitive, will not work, because when animating an iterator, you can't allocate memory inside it, and Image.crop() creates a new image and necessarily allocates memory for it.

I wonder if we could somehow work around this. The obvious solution would be to create a list of all possible scrolling stages, and feed that to the animation function, but somehow that feels wrong, and also wastes a lot of memory.

Another solution would be to allow an optional argument to methods like crop or shift_*, that would provide an image object to use instead of creating a new image. Of course that image would need to have the right dimensions.

Another solution would be to add a "blit" method, that would let us copy a fragment from one image into another image, and then we could use it in place of crop. This would be also consistent with how PyGame works, which is often taught in Python classes, so it could be familiar to kids.

PR #401, opened 2017-01-15
Add blit method to image class.

markshannon:
I've also simplified the rest of the image code by implementing crop and shift_xxx in terms of blit.

Fixes https://github.com/bbcmicrobit/micropython/issues/395


source/microbit/microbitimage.cpp
```diff
@@ -347,27 +271,45 @@ STATIC mp_obj_t microbit_image_make_new(const mp_obj_type_t *type_in, mp_uint_t
...
-    if (w > intersect_x1 - intersect_x0 || h > intersect_y1 - intersect_y0) {
-        result->clear();
+    mp_int_t intersect_x0 = max(0, xdest);
+    mp_int_t intersect_y0 = max(0, ydest);
+    mp_int_t intersect_x1 = min(dest->width, xdest+w);
+    mp_int_t intersect_y1 = min(dest->height, ydest+h);
+    for (int i = intersect_x0; i < intersect_x1; ++i) {
+        for (int j = intersect_y0; j < intersect_y1; ++j) {
+            dest->setPixelValue(i, j, 0);
+        }
```
deshipu: Why are you clearing the target rectangle? Won't the copy overwrite it all the pixels anyways?

source/microbit/microbitimage.cpp
```diff
@@ -439,6 +381,39 @@ mp_obj_t microbit_image_fill(mp_obj_t self_in, mp_obj_t n_in) {
...
+            "must specify both offsets"));
+    }
+    mp_int_t x = mp_obj_get_int(args[2]);
+    mp_int_t y = mp_obj_get_int(args[3]);
+    mp_int_t w = mp_obj_get_int(args[4]);
+    mp_int_t h = mp_obj_get_int(args[5]);
+    if (w < 0 || h < 0) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_ValueError,
+            "size cannot be negative"));
+    }
```
deshipu: Should we have a check that we are not blitting onto itself? (Or, better, if we are, that the areas don't overlap.) The results of that could be funky...

source/microbit/microbitimage.cpp
deshipu: Ah, I see, this is for the case where the source rectangle sticks outside the source image.

markshannon: If I'd remembered to commit the test, that would have been more obvious :man_facepalming:

source/microbit/microbitimage.cpp
markshannon: Or just iterate in the right direction, so we do it correctly.
Blitting onto the same image might be handy for scrolling.
I'll add some tests and tweak the iteration order.

markshannon:
Now supports blitting an image onto itself.

source/microbit/microbitimage.cpp
```diff
@@ -439,6 +410,39 @@ mp_obj_t microbit_image_fill(mp_obj_t self_in, mp_obj_t n_in) {
...
+    if (n_args == 7) {
+        nlr_raise(mp_obj_new_exception_msg(&mp_type_TypeError,
+            "must specify both offsets"));
+    }
+    mp_int_t x = mp_obj_get_int(args[2]);
+    mp_int_t y = mp_obj_get_int(args[3]);
+    mp_int_t w = mp_obj_get_int(args[4]);
+    mp_int_t h = mp_obj_get_int(args[5]);
+    if (w < 0 || h < 0) {
```
deshipu: This is probably a crazy idea, but whenever I see a `ValueError` raised I begin to wonder whether that case could actually make sense and be useful, and in this case I think it would be natural to simply copy the region as mirrored (with negative width) or flipped up-side-down (with negative height), or both. I know this is probably putting too much functionality into a single method, and makes it really complex, but I thought I would just throw it here anyways. Just ignore if you think it's silly.

source/microbit/microbitimage.cpp
```diff
@@ -347,27 +271,74 @@ STATIC mp_obj_t microbit_image_make_new(const mp_obj_type_t *type_in, mp_uint_t
...
+            dest->setPixelValue(i+xdest-x, j+ydest-y, val);
         }
     }
+    // Adjust intersection rectange to dest
+    intersect_x0 += xdest-x;
+    intersect_y0 += ydest-y;
+    intersect_x1 += xdest-x;
+    intersect_y1 += ydest-y;
+    // Clear four rectangles in the cleared area surrounding the copied area.
+    clear_rect(dest, clear_x0, clear_y0, intersect_x0, intersect_y1);
```
dpgeorge: Wouldn't it be simpler to unconditionally clear the dest region first, then just copy across those pixels from the src that are in range (which might be no pixels in the case of "nothing to copy")?  This would eliminate the need for these 4 clears.

markshannon: Not if you are blitting an image onto itself.

dpgeorge: True!

dpgeorge:
It's merged!

PR #402, opened 2017-01-16
Fix radio docs: change default power to "6", which is 0 dBm.

dpgeorge:
The default power level for radio is 0 dBm, which is "power=6" in the API.  This patch updates the docs to reflect this.

To whoever reviews this: please double check the code in source/microbit/modradio.cpp to check that I got this right, thanks!

### DIFF
--- a/docs/radio.rst
+++ b/docs/radio.rst
@@ -71,7 +71,7 @@ Functions
     via this channel will be put onto the incoming message queue. Each step is
     1MHz wide, based at 2400MHz.
 
-    The ``power`` (default=0) is an integer value from 0 to 7 (inclusive) to
+    The ``power`` (default=6) is an integer value from 0 to 7 (inclusive) to
     indicate the strength of signal used when broadcasting a message. The
     higher the value the stronger the signal, but the more power is consumed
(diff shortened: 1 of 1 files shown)


ntoll:
Morning @dpgeorge!

Just to be very explicit about this, the important lines are:

https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/modradio.cpp#L39 (where the default power is defined as 0 [dBm]).

https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/modradio.cpp#L287 (where the initial / reset radio state is set to the default power).

https://github.com/bbcmicrobit/micropython/blob/master/source/microbit/modradio.cpp#L336 (where the different available power settings are defined with 0 dBm in position 6).

Ergo, the default `power` setting in Python is correctly described in this patch as "6" since 0 dBm is in the sixth slot as demonstrated above.

ntoll:
To be clear, I've just kicked off a new documentation build on RtD.

dpgeorge:
> Just to be very explicit about this, the important lines are: ...

Yes, exactly!  Thanks for merging.

PR #405, opened 2017-01-21
Sort out pin modes and music module

### LINKED ISSUE #398: ValueError: pin0 is in music mode

markshannon:
Fixes #398

This PR:

1. Changes pin-mode implementation to use pointers to structs, rather than strings. This should avoid any race conditions when releasing a pin in an interrupt handler.
2. Documents which pin-mode functions can and cannot be called in an interrupt handler.
3. Improves the pin-mode API doing the sensible thing when passed a NULL or attempting to re-acquire the same pin.
4. Fix music module's pin releasing.
5. Adds a test for the music module (just tests that it doesn't crash, not that it makes the right noises :musical_note: ).

Includes a revert of #399 and supersedes #403

### DIFF
--- a/tests/test_music.py
+++ b/tests/test_music.py
@@ -1,0 +1,46 @@
+
+from microbit import display, Image, pin0, pin1, pin2, pin8
+import music
+
+PINS = pin0, pin1, pin2, pin8
+FREQS = 400, 500, 600, 700, 800, 900
+
+def test_rapid_pin_switch():
+    for i in range(20):
+        for pin in PINS:
+            music.play(music.NYAN, pin, wait=False)
+
+def test_rapid_pitch_switch():
+    for i in range(20):
+        for freq in FREQS:
+            music.pitch(freq, wait=False)
+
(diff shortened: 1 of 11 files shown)


markshannon:
I've also added a fix for passing zero to the `pitch` function. For pitch 0, we just turn the pin off.
Negative pitch remains an error.

tests/test_music.py
```diff
@@ -0,0 +1,50 @@
...
+
+def test_repeated_stop():
+    for i in range(20):
+        music.stop()
+
+def test_repeated_reset():
+    for i in range(20):
+        music.reset()
+
+def test_repeated_stop():
```
dpgeorge: This function is duplicated!

markshannon: removed

dpgeorge:
Great work @markshannon!

PR #407, opened 2017-01-21
SPI acquires pins only when initialized and methods raise if SPI is not initialized before use.

### LINKED ISSUE #406: Calling `spi.read(1)` before `spi.init()` cause device to freeze
Trivial to reproduce

```python
>>> spi.read(1)
```
freezes.

But,
```python
>>> spi.init()
>>> spi.read(1)
b'\xff'
```

`write` and `write_readinto` exhibit the same problem.

markshannon:
Fixes #406 
Also fixes the problem that #397 was meant to address.

**WARNING** this hasn't been tested properly as I don't have the necessary hardware.
I have tested that pins can be used as expected, but I am not able test that the SPI still works correctly.

This PR needs #405 to be merged first (and possibly then rebased), I am putting it up here in case anyone is feels like testing it.

### DIFF
--- a/inc/microbit/microbitpin.h
+++ b/inc/microbit/microbitpin.h
@@ -69,6 +69,28 @@ extern const microbit_pinmode_t microbit_pinmodes[];
 #define microbit_pin_mode_i2c           (&microbit_pinmodes[MODE_I2C])
 #define microbit_pin_mode_spi           (&microbit_pinmodes[MODE_SPI])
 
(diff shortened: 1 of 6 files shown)


dpgeorge:
@markshannon I guess this is next on the list?

dpgeorge:
Superseded by e04006d67764da0e12251f9621d3254b7fe9b23a

PR #410, opened 2017-02-13
Implement radio.receive_full() function.

dpgeorge:
This PR adds the radio.receive_full() function, discussed in #362.

Note: I did not include a timestamp in the return value!  (So it returns just (bytes, rssi).)  We kind of discussed it in #362 but I'm not sure what the consensus was.  And adding a timestamp means using an extra 4 bytes per packet to store it (which is quite a bit if you have a large queue of packets).

So we need to decide if timestamp is worth the extra 4 bytes of storage overhead.


Neon22:
The timestamp is useful in situations like multiplayer games where the microbits signal each oher.
Knowing when a packet comes in is useful to determine the 'winner' in such cases.

But we can always use recieve_bytes in any situation to craft what is needed. So the answer probaly lies in what is the purpose of receive_full()

deshipu:
I think in case of multiplayer games and such, you really want to include the timestamp in the message, as you want to know when the message was sent, not when it was received.

ntoll:
I believe the timestamp issue is a red herring given that there's no clock on the device. All we can tell is the number of milliseconds since the device started. Ergo, sending a timestamp in a packet wouldn't be useful. I believe having the time of receipt of the message would be useful for some of the use cases discussed in #362.

ZanderBrown:
Of course if you felt the need you could do time synchronization using a microbit connected to a PC.

deshipu:
>> I believe the timestamp issue is a red herring given that there's no clock on the device. All we can tell is the number of milliseconds since the device started. Ergo, sending a timestamp in a packet wouldn't be useful

> Of course if you felt the need you could do time synchronization using a microbit connected to a PC.

Or synchronize the timestamps of the two microbits at the beginning of your communication session (calculating the offsets). But that's definitely outside of the scope of the standard library.

ZanderBrown:
> > >  believe the timestamp issue is a red herring given that there's no clock on the device. All we can tell is the number of milliseconds since the device started. Ergo, sending a timestamp in a packet wouldn't be useful

> > Of course if you felt the need you could do time synchronization using a microbit connected to a PC.

> Or synchronize the timestamps of the two microbits at the beginning of your communication session (calculating the offsets). But that's definitely outside of the scope of the standard library.

It's also probably a bit excessive.

markshannon:
Looks good, but the docs in #362 need updating to remove `timestamp`.
Could #362 then be included in this PR?

source/microbit/modradio.cpp
```diff
@@ -449,17 +464,28 @@ STATIC mp_obj_t mod_radio_send(mp_obj_t buf_in) {
...
 }
 MP_DEFINE_CONST_FUN_OBJ_1(mod_radio_receive_bytes_into_obj, mod_radio_receive_bytes_into);
 
+STATIC mp_obj_t mod_radio_receive_full(void) {
+    int rssi;
+    mp_obj_t bytes = radio_receive(false, NULL, &rssi);
+    if (bytes == mp_const_none) {
+        return mp_const_none;
+    }
+    mp_obj_t t[2] = {bytes, MP_OBJ_NEW_SMALL_INT(rssi)};
```
markshannon: According to #362 `result[1]` should be the negative of rssi

dpgeorge: See line 281 above: it's negating the raw value there.

dpgeorge:
> but the docs in #362 need updating to remove timestamp.

Yes, if we agree that timestamp is not worth adding.  Probably for cases where timing is important one can just make a loop which constantly polls for packets and then calls microbit.running_time() to get the timestamp for it.

dpgeorge:
radio.receive_full() was implemented in e6f9ab13baa86cab58cbd093c7f2adf91ce56af7

PR #412, opened 2017-02-26
Add i2c.init to help

deshipu:
Without this, it doesn't appear in the code completion API, and people
who use the Mu editor don't know it's there.

### DIFF
--- a/source/microbit/help.c
+++ b/source/microbit/help.c
@@ -138,6 +138,7 @@ STATIC const mp_doc_t help_table_instances[] = {
     {&microbit_i2c_obj, "Communicate with one or more named devices connected to micro:bit. Each named\ndevice has an 'address', communicates using I2C, and connects to the I/O pins.\n"},
     {&microbit_i2c_read_obj, "Use read(address, n) to read 'n' bytes from the device with this address.\n"},
     {&microbit_i2c_write_obj, "Use write(address, buffer) to write to the 'buffer' of the device at this 'address'.\n"},
+    {&microbit_i2c_init_obj, "Use init(frequency, scl, sda) to set the bus frequency and pins.\n"},
     // Image
     {&microbit_image_type, "Create and use built-in IMAGES to show on the display. Use:\nImage(\n  '09090:'\n  '99999:'\n  '99999:'\n  '09990:'\n  '00900:')\n...to make a new 5x5 heart image. Numbers go from 0 (off) to 9 (brightest). Note\nthe colon ':' to set the end of a row.\n"},
     {&microbit_image_width_obj, "Return the width of the image in pixels.\n"},


source/microbit/help.c
```diff
@@ -138,6 +138,7 @@ STATIC const mp_doc_t help_table_instances[] = {
     {&microbit_i2c_obj, "Communicate with one or more named devices connected to micro:bit. Each named\ndevice has an 'address', communicates using I2C, and connects to the I/O pins.\n"},
     {&microbit_i2c_read_obj, "Use read(address, n) to read 'n' bytes from the device with this address.\n"},
     {&microbit_i2c_write_obj, "Use write(address, buffer) to write to the 'buffer' of the device at this 'address'.\n"},
+    {&microbit_i2c_init_obj, "Set the bus frequency pins.\n"},
```
dpgeorge: How about "Set the bus frequency and pins to use."?

deshipu:
Apparently this file: https://github.com/mu-editor/mu/blob/7c171e9c894657293c3f8d27196fd92dae58df94/mu/resources/api.py
is generated from the help, and then used for code completion in Mu.

source/microbit/help.c
deshipu: Thanks, updated.

source/microbit/help.c
```diff
@@ -138,6 +138,7 @@ STATIC const mp_doc_t help_table_instances[] = {
     {&microbit_i2c_obj, "Communicate with one or more named devices connected to micro:bit. Each named\ndevice has an 'address', communicates using I2C, and connects to the I/O pins.\n"},
     {&microbit_i2c_read_obj, "Use read(address, n) to read 'n' bytes from the device with this address.\n"},
     {&microbit_i2c_write_obj, "Use write(address, buffer) to write to the 'buffer' of the device at this 'address'.\n"},
+    {&microbit_i2c_init_obj, "Set the bus frequency and pins to use.\n"},
```
dpgeorge: Sorry, I just realised we should follow existing messages.  Eg: "Use init(freq, sda, scl) to set the bus 'freq' and pins to use.\n".

deshipu:
I updated the message -- but not all of them follow this pattern, so I wasn't sure at the beginning if it's worth it. This is an advanced function after all.

I also learned that the Mu file I mentioned is not actually generated automatically anymore -- it was initially, but now is updated manually. So it's not necessary to merge this to have it in that file. Since it's an advanced function, I'm not sure if we want to have it here. I will make a separate pull request for Mu for adding it there.

dpgeorge:
> Since it's an advanced function, I'm not sure if we want to have it here.

I think it's ok to add it here, for completeness.  Afterall, uart.init and spi.init are documented here.

Issue #370, opened 2016-10-30
Traceability to source


whaleygeek:
It would be really helpful to be able to trace an arbitrary running version of MicroPython on an arbitrary micro:bit, back to the github source that generated it. 

This would be especially useful when diagnosing problems that users have, to say 'tell me the version string that you get (when you type this in) or (when you press the reset button), as anyone could then look through github for that version and see if a particular feature or bug-fix is, or is not, implemented.

Also, it's quite important to be able to setup an environment when reproducing issues, to be using the same version of micro:bit MicroPython as others are using. So, if there is an easy(ish) process by which that identifying information can be followed to a cached released .hex that could just be downloaded and used to reproduce the issue, that would additionally help.

The startup message that you get when you press the reset button seems to identify the root MicroPython version, but not the actual build of the micro:bit version that was based on that code.

carlosperate:
Also, as I was mentioning in https://github.com/bbcmicrobit/micropython/issues/199#issuecomment-256902093 it would be ideal if we could identify the version from the hex file itself.

whaleygeek:
Yes @carlosperate that would be a great facility too - someone could email a .hex file of an 'errant script' and a simple tool could be used to work out the version string and therefore the exact build of micropython for microbit that generated the interpreter inside it.

whaleygeek:
It sounds like we need the magic of @ntoll to help us work up an initial spec for this, as that is what normally makes these things happen in the end!

dpgeorge:
What we need is:
- proper tags in this repository for each version we release
- the tag exposed to the user, in the start-up banner and in os.uname()
- the git hash from this repo that was used to build the firmware exposed to the user
- the upstream micropython version and git hash exposed to the user

markshannon:
IMO providing git hash, tags and other such artifacts can be fiddly and confusing.
If we just updated the version number more often, it should solve @whaleygeek's problem.
Ideally a new version of the flash would be uploaded to mu only when we add a new version.
Version 1.1 now?

carlosperate:
Not to sound like a broken record, but having the version in the hex file in a known and/or identifiable location would also help things like the online editor to chose the correct version of micropython when a user hex file is loaded. So, if future features reduce the memory available for user programs, old scripts that were on that limit could still be loaded, edited and run.

dpgeorge:
> IMO providing git hash, tags and other such artifacts can be fiddly and confusing.

[...]

Issue #392, opened 2016-12-31
microbit hangs when using radio and PWM


DFBREB:
When two micro:bits communicate using radio and the transmitting micro:bit sends a string to the receiving micro:bit and the reciving microbit converts it to an integer from 0 to 1023 which it hen sends as an output PWM signal via a pin (in my case P0 and P8) after 20 seconds to 200 seconds the receiving micro:bit freezes/hangs.  The receiving micro:bit cannot be reset with the reset button in 90% of the cases.

The Beta Editor with Radio found at microbit.org is being used

However if the write analog lines are commented out the radio part of the system continues to work indefinitely.  

Damien George is looking into this issue at this time.  

Since this site does not support .hex or .py files I am unable to send scripts.
David Booth

whaleygeek:
You can zip the files first then attach the zip files

On 31 Dec 2016 11:10 pm, "DFBREB" <notifications@github.com> wrote:

When two micro:bits communicate using radio and the transmitting micro:bit
sends a string to the receiving micro:bit and the reciving microbit
converts it to an integer from 0 to 1023 which it hen sends as an output
PWM signal via a pin (in my case P0 and P8) after 20 seconds to 200 seconds
the receiving micro:bit freezes/hangs. The receiving micro:bit cannot be
reset with the reset button in 90% of the cases.

The Beta Editor with Radio found at microbit.org is being used

However if the write analog lines are commented out the radio part of the
system continues to work indefinitely.

Damien George is looking into this issue at this time.

Since this site does not support .hex or .py files I am unable to send
scripts.
David Booth

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/bbcmicrobit/micropython/issues/392>, or mute the thread

jezdean:
Does the same issue happen on the [microbit.co.uk ](https://www.microbit.co.uk/app/#create:xyelfe)editor? It uses a different micropython hex file to the `.co.uk` and `mu` editor. 

If the issue is not present, the problem might be a buggy out-of-date `.hex` file generated by the `.co.org` editor. A similar problem was reported in [issue 391](https://github.com/bbcmicrobit/micropython/issues/391).

 It has also been [reported to the uFlash ](https://github.com/ntoll/uflash/issues/33)repository.

DFBREB:
Hi Jez
Thanks for the suggestion.  I will try it.
David

dpgeorge:
[...]

PR #420, opened 2017-03-23
Add additional version info to os.uname().version.

dpgeorge:
This additional info can be used to tell the exact git commit from this repository that was used to build the firmware.

Exampe output:
```
MicroPython v1.7-9-gbe020eb on 2016-04-18; micro:bit with nRF51822
>>> import os
>>> os.uname().version
"micro:bit v1.0-b'64a61db' (dirty); MicroPython v1.7-9-gbe020eb on 2016-04-18"
>>> 
```

This gives the git hash of this repo, as well as the upstream uPy version.

If one wanted to extract this info from a hex file it would be pretty straightforward: just write a script to scan the .hex file for the string "micro:bit v" and then extract the version info.

See #370 for discussion around this point.

### DIFF
--- a/source/microbit/modos.c
+++ b/source/microbit/modos.c
@@ -29,6 +29,10 @@
 #include "py/objtuple.h"
 #include "py/objstr.h"
 #include "genhdr/mpversion.h"
+#include YOTTA_BUILD_INFO_HEADER
+
+#define _MP_STRINGIFY(x) #x
+#define MP_STRINGIFY(x) _MP_STRINGIFY(x)
 
 #define RELEASE "1.0"
 
@@ -39,7 +43,14 @@ STATIC const qstr os_uname_info_fields[] = {
 STATIC const MP_DEFINE_STR_OBJ(os_uname_info_sysname_obj, MICROPY_PY_SYS_PLATFORM);
 STATIC const MP_DEFINE_STR_OBJ(os_uname_info_nodename_obj, MICROPY_PY_SYS_PLATFORM);
 STATIC const MP_DEFINE_STR_OBJ(os_uname_info_release_obj, RELEASE);
-STATIC const MP_DEFINE_STR_OBJ(os_uname_info_version_obj, MICROPY_GIT_TAG " on " MICROPY_BUILD_DATE);
+STATIC const MP_DEFINE_STR_OBJ(os_uname_info_version_obj,
+    "micro:bit v" RELEASE
+    "-" MP_STRINGIFY(YOTTA_BUILD_VCS_DESCRIPTION)
+    #if YOTTA_BUILD_VCS_CLEAN == 0
+    "-dirty"
+    #endif
+    " on " MP_STRINGIFY(YOTTA_BUILD_YEAR) "-" MP_STRINGIFY(YOTTA_BUILD_MONTH) "-" MP_STRINGIFY(YOTTA_BUILD_DAY)
+    "; MicroPython " MICROPY_GIT_TAG " on " MICROPY_BUILD_DATE);
 STATIC const MP_DEFINE_STR_OBJ(os_uname_info_machine_obj, MICROPY_HW_BOARD_NAME " with " MICROPY_HW_MCU_NAME);
 
 STATIC MP_DEFINE_ATTRTUPLE(


whaleygeek:
'microbit v' as a string sounds a little brittle.

For PXT we have a binary signature that identifies a metadata record, and that is pretty robust.

I appreciate this comment here is a retro-fit, but I wonder if longer term we need a more robust method of embedding simple metadata into the .hex so that it can be retrieved by an external tool?

I'm reminded of similar systems I developed years ago, where a specific binary tag and record termination protocol was used to allow any module to embed metadata into the final object - and I wrote a tool that scanned and extracted all tags, which was great for post-analysis of 'any old .hex file that happened to be lying around'. Allowing any module to use this tagging format was very powerful. The number of bytes required for a tag can be quite small, providing there are additional protections around the record, such as a defined protocol for termination of the record, it makes it quite robust.

dpgeorge:
> 'microbit v' as a string sounds a little brittle.

You can scan for more than that, you can also check for the "; MicroPython v" string and then verify that the whole thing has the right format (eg digits in the correct place, etc).

> I appreciate this comment here is a retro-fit

Yes this is a short term solution.  When (if?) we move to the mainline MicroPython repository then things will be different.

markshannon:
This should be very useful pulling apart issues like #392 and #421 in the future.
Happy to merge unless @whaleygeek has any objections.

whaleygeek:
No objections - anything that improves traceability to the precise source
files used to generate that .hex is always a good thing, thanks.

It will significantly improve our ability to send you much better bug
reports in the future (and filter out spurious ones that just need people
to upgrade to the latest deployed editors, or wait for planned fixes to be
deployed into those editors).

markshannon:
I tried this out, but it makes no difference for me. `YOTTA_BUILD_VCS_DESCRIPTION` is not defined.

My `YOTTA_BUILD_INFO_HEADER` file looks this:
```c
#ifndef __YOTTA_BUILD_INFO_H__
#define __YOTTA_BUILD_INFO_H__
// yotta build info, #include YOTTA_BUILD_INFO_HEADER to access
#define YOTTA_BUILD_YEAR 2017 // UTC year
#define YOTTA_BUILD_MONTH 4 // UTC month 1-12
#define YOTTA_BUILD_DAY 9 // UTC day 1-31
#define YOTTA_BUILD_HOUR 16 // UTC hour 0-24
#define YOTTA_BUILD_MINUTE 26 // UTC minute 0-59
#define YOTTA_BUILD_SECOND 57 // UTC second 0-61
#define YOTTA_BUILD_UUID 9ffbacea-d18a-45dc-a0ca-0418b53ce567 // unique random UUID for each build
#define YOTTA_BUILD_VCS_ID e90c2dbc9273a55ad4053404565b2e27fd4e82fc // git or mercurial hash
#define YOTTA_BUILD_VCS_CLEAN 1 // evaluates true if the version control system was clean, otherwise false
#endif // ndef __YOTTA_BUILD_INFO_H__
```
Should we be using `YOTTA_BUILD_VCS_ID` instead?

dpgeorge:
That's strange, I have the same data in build/bbc-microbit-classic-gcc-nosd/yotta_build_info.h plus one extra line:
```
#define YOTTA_BUILD_VCS_DESCRIPTION b'9aa6dc9' // git describe or mercurial equivalent
```
Maybe I have a newer version of yotta?  yt --version gives 0.16.4 for me.  Also, my version has the VCS_ID as a Python bytes object:
```
#define YOTTA_BUILD_VCS_ID b'9aa6dc9ebf12700b9f282ac9d6ef6fd8a16ad26e' // git or mercurial hash
```

markshannon:
I had an ancient version of yotta, so I've upgraded (to 0.18.2) and it now works.
I think we should remove the `#if defined(YOTTA_BUILD_VCS_DESCRIPTION)` line, that way we can't accidentally produce a build without the git-tag (as was happening for me before I upgraded).

dpgeorge:
> I think we should remove the #if defined(YOTTA_BUILD_VCS_DESCRIPTION) line,

Ok, I've removed it (the downside is that you must have used git to check out the code).

I also added the build date to the version info.

deshipu:
You are potentially going to have a hard time with distro packagers with this
if you ever want it to get packaged into any distribution. The packages
are always built from released archives, and never from a git repo.

dpgeorge:
> The packages are always built from released archives, and never from a git repo.

We don't have release archives, and using yotta makes it hard to do that (unless there's a special yotta command to make a release archive...).  So, as long as we are tied to yotta (which hopefully won't be for much longer) I don't think it's a problem to use this build info.

markshannon:
LGTM

@dpgeorge, @deshipu are you both happy for me to merge this?

deshipu:
I see no further problems other than the one I mentioned, and as @dpgeorge said, that is a problem that we have anyways, at least for now. we can always change this if we decide to change the current build system.

dpgeorge:
Should be ok.

PR #422, opened 2017-03-28
Remove a few more DAL dependencies.

markshannon:
Removes a few "#includes" and replaces `panic()` and `temperature()` with non-DAL versions.

### DIFF
--- a/source/microbit/modantigravity.cpp
+++ b/source/microbit/modantigravity.cpp
@@ -26,9 +26,6 @@
  * THE SOFTWARE.
  */
 
-#include <stdio.h>
-#include "MicroBit.h"
-
 extern "C" {
 
 #include "py/mphal.h"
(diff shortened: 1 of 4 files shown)


dpgeorge:
This looks ok, except that `panic()` has changed behaviour: it used to loop indefinitely and require a reset to restart the device, whereas now it restarts automatically after 1 second.  We should keep the original behaviour because this function is only ever called by the user and they must have wanted an infinite loop if they called the function.  We really don't want to change behaviour when we don't need to.

markshannon:
I've made `panic()` act exactly the same as before, even using the same image.

dpgeorge:
> I've made panic() act exactly the same as before, even using the same image.

I don't see the changes.... did you push them?

markshannon:
Apparently not. I have now.

dpgeorge:
Thanks!

PR #428, opened 2017-04-05
Increase width of 2 variables that should be larger than a byte.

### LINKED ISSUE #427: Incorrect music playback speed at BPM<=58
```
from microbit import *
import music

#bpm = 60
bpm = 58
tune = ['C5:3', 'D5:3', 'E5:3', 'F5:3']
music.set_tempo(ticks=4, bpm=bpm)
music.play(tune, wait=True)
```

Works fine at 60BPM
Plays blindingly fast at 58BPM and below

dpgeorge:
If bpm is small then the delay between notes (ms_per_ticks) will be larger
than 255 so needs more than a byte for the computation.

The variable current_position counts bytes so should be a size_t.

Should fix #427.


markshannon:
LGTM and works for me.

PR #431, opened 2017-04-09
DAL-free accelerometer, compass and I2C.

markshannon:
Updates accelerometer on ticks, rather than on demand allowing us to use jerk (deriviative of acceleration) to determine shake, giving improved gesture detection.
Extend range from 2g to 8g, using the 2 least significant bits that were previous ignored.
Compass is now updated on ticks, rather than on demand. This is necessary to avoid i2c conflicts with
the accelerometer.


dpgeorge:
This is a large change and I didn't go through it in detail, but a few high-level comments are:
- the removal of 3g/6g/8g constants (qstrs) indicates that there is a breaking-change from a user's point of view, which we should avoid
- if accel and compass are updated on ticks instead of on demand then that will use more power.  In particular for users that don't use these sensors, their microbits are going to suddenly use a lot more power, so this would be considered a breaking change.
- the heap is decreased by about 30 bytes -- why is that necessary?

markshannon:
Sadly it turns out that this is broken. We just can't use I2C asynchronously and synchronously, even with a mutex. The problem is that i2c read/writes aren't really atomic. Often we want to send a command, consisting of a write and read. Interrupting between the two won't work. 
Unfortunately gesture detection doesn't work when calling the accelerometer synchronously, unless an accelerometer method is called at regular intervals.

So, I've reverted to synchronous operation (still using the integral of |jerk| instead of zero-crossing) and made it clear it the docs that gesture detection won't work unless an accelerator method is called at regular intervals.

> the removal of 3g/6g/8g constants (qstrs) indicates that there is a breaking-change from a user's point of view, which we should avoid

Behaviour is unchanged. `accelerometer.get_gesture() == "3g"` continues to be a synonym for `False`.
Given the hardware was limited to 2g the "gestures" 3g/6g/8g were never possible, so any use of them was erroneous.
And what does 3g as a "gesture" mean? 3g for how long? Tilt and shake/jerk are easily explained. Impulse is a lot harder.

> the heap is decreased by about 30 bytes -- why is that necessary?

Ask the linker :wink: My guess is that although this PR duplicates some the data in the DAL's accelerometer and compass classes, the linker still thinks we need the data in those classes.
See https://github.com/bbcmicrobit/micropython/pull/434 for improved heap size.

markshannon:
All cleaned up and ready to review.
I've reverted to only updating the compass and accelerometer on demand. This makes it fully backwards compatible at the API level. Gesture detection is not as good as the background update version, but it is a bit better than on master (from my somewhat adhoc testing).

markshannon:
Anyone feel like reviewing this?

dpgeorge:
It's tough to review such a huge PR.  It would be a lot easier if it were broken down into manageable pieces, or, better still, done in stages.

Does the vector type really need to be exposed to everyone?  It would be cleaner if it were contained to the file that needed it.  If it does need to be public, I'd say the type name should be made a bit less generic and changed to something like `vec3d_16int_t` (to make it clear it's 3 components and they are ints).

markshannon:
Since both the accelerometer and compass depend on i2c, it is much simpler to do it all at once. Sorry.

I've renamed `vector` to `microbit_vector3_t` so it doesn't mess with the main µPy namespace. The vector maths is used by both the accelerometer and the compass, so it can't be static.

examples/magic8.py
```diff
@@ -35,4 +35,5 @@
         display.clear()
         sleep(1000)
         display.scroll(random.choice(answers))
-    sleep(10)
+    #Accelerometer updates at 50Hz
```
ZanderBrown: Perhaps explain why this is relevent to `sleep(20)`?

markshannon: 50Hz = 20ms per cycle

ZanderBrown: I understand that but the comment doesn't necessarily expalin it

dpgeorge:
A few comments on this PR (I didn't check everything yet):
- accelerometer x and y values are negated compared to original behaviour (and hence the compass calibration is all backwards)
- get_gestures() doesn't seem to work very well, eg the following loop just prints "shake" gesture no matter what you do to the microbit:
```
while 1:
    g = accelerometer.get_gestures()
    if g:
        print(g)
```
- when adding copyright to the header of the file, I think it's better to add a completely new line (following after existing ones), of the form " * Copyright (c) 2017 Mark Shannon"
-

dpgeorge:
Unfortunately this is rather stale and can no longer be merged.  In particular because the micro:bit accel and compass may now use different parts the code here will not work on newer micro:bits.  The code relies on the DAL so the DAL can abstract away these hardware changes.

PR #432, opened 2017-04-09
Move music module's data into a heap-allocated struct.

markshannon:
Saves 20bytes RAM when music module is not in use.

### DIFF
--- a/source/microbit/microbitmusic.cpp
+++ b/source/microbit/microbitmusic.cpp
@@ -43,14 +43,23 @@ extern "C" {
 #define DEFAULT_DURATION 4 // Crotchet
 #define ARTICULATION_MS  10 // articulation between notes in milliseconds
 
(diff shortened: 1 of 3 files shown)


source/microbit/microbitmusic.cpp
```diff
@@ -43,97 +43,94 @@ extern "C" {
...
 
     // store these to simplify the writing process
     uint8_t last_octave;
     uint8_t last_duration;
-} music_state_t;
+
+    // Asynchronous parts.
+    volatile uint8_t async_state;
+    uint32_t async_wait_ticks;
+    bool async_loop;
```
dpgeorge: This entire struct can be made smaller (by 4 bytes) by moving this async_loop member up one, to sit just after async_state.

markshannon: Done.

PR #433, opened 2017-04-09
Remove DAL dependency for random module.

markshannon:
Also fixes 1-in-4-billion bug when the hardware RNG produces a 32 bit 0.

### DIFF
--- a/source/microbit/random.cpp
+++ b/source/microbit/random.cpp
@@ -1,0 +1,139 @@
+/*
+The MIT License (MIT)
+
+Copyright (c) 2016 British Broadcasting Corporation.
+This software is provided by Lancaster University by arrangement with the BBC.
+
+Permission is hereby granted, free of charge, to any person obtaining a
+copy of this software and associated documentation files (the "Software"),
+to deal in the Software without restriction, including without limitation
+the rights to use, copy, modify, merge, publish, distribute, sublicense,
+and/or sell copies of the Software, and to permit persons to whom the
+Software is furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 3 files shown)


dpgeorge:
It looks good BUT... if code is cut-and-paste verbatim (or nearly verbatim) from the DAL then the DAL's license and copyright must also be included.

markshannon:
Good point. I've copied the DAL code into its own file (retaining the copyright) and modified that to apply the fix.

dpgeorge:
This doesn't build for me because it uses 4 bytes more of RAM for the state of the PRNG.

See #435 for a permanent fix to the changing-heap-size issue.

source/microbit/random.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+
+/**
+  * Seed our a random number generator (RNG).
+  * We use the NRF51822 in built cryptographic random number generator to seed a Galois LFSR.
+  * We do this as the hardware RNG is relatively high power, and use the the BLE stack internally,
+  * with a less than optimal application interface. A Galois LFSR is sufficient for our
+  * applications, and much more lightweight.
+  */
+void MicroBit_seedRandom(void)
```
ZanderBrown: Any particular reason for this being `MicroBit_seedRandom` rather than `micro_bit_seed_random` or `microbit_seed_random`?

I know this has been 'ported' from C++ but i though C convertion was to avoid capitals in function names?

markshannon: I just wanted to make the minimum change possible from the original.

ZanderBrown: okay, just looked a little odd

dpgeorge:
Ok, this one is merged.

PR #434, opened 2017-04-15
Dalek to me (pronounced DAL-ectomy)

markshannon:
NOT READY FOR MERGING.
https://github.com/bbcmicrobit/micropython/pull/431 needs merging first.

Frees up 3k of RAM, upping the heap size to 12k and frees up about 30k of flash, so we have plenty of space for the file-system.

Fixes #107

PR #435, opened 2017-04-18
Automatically allocate the uPy heap using all the left-over RAM.

dpgeorge:
This change removes the fixed-size uPy heap and replaces it with automatic allocation so that the heap is as large as can be, using all the left-over RAM.  The location of the uPy heap is after the mbed 2k heap and before the 2k stack.

This automatic allocation should prevent any build problems when there are small changes in the amount of data/bss being used, or when the code is built with a different version of the compiler, and will ensure that the entire RAM is always used to its full potential.

### DIFF
--- a/source/microbit/mprun.c
+++ b/source/microbit/mprun.c
@@ -98,21 +98,10 @@ void mp_run(void) {
     mp_stack_ctrl_init();
     mp_stack_set_limit(1800); // stack is 2k
 
-    // allocate the heap statically in the bss
-    static uint32_t heap[9732 / 4];
-    gc_init(heap, (uint8_t*)heap + sizeof(heap));
-
-    /*
-    // allocate the heap using system malloc
-    extern void *malloc(int);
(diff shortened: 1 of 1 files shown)


markshannon:
LGTM.

Longer term we want to use most, if not all, of the mbed heap as well. Also, we ought to choose a fixed size heap with enough headroom for new code. See https://github.com/bbcmicrobit/micropython/pull/434/files#diff-ca2db8d00aff7c82ce7787291acc5181R103 for a possible approach.

For now though, this saves fiddling with each PR to get the heap use exactly right.

DavidWhaleMEF:
Nice fix @dpgeorge - I'd really like to see a published (fixed) amount of RAM for user code/vars that we never go below - otherwise any change could be a breaking change for users.

dpgeorge:
This fix doesn't work.  The underlying newlib uses _sbrk and assumes it can have all of the "unused" RAM.  This patch will need to be reverted.

PR #436, opened 2017-04-18
Fix music module so the ticker does nothing if module wasn't imported.

dpgeorge:
Otherwise the microbit can (and does for me) randomly crash just when sitting at the REPL.

### DIFF
--- a/source/microbit/microbitmusic.cpp
+++ b/source/microbit/microbitmusic.cpp
@@ -74,6 +74,11 @@ extern uint32_t ticks;
 STATIC uint32_t start_note(const char *note_str, size_t note_len, const microbit_pin_obj_t *pin);
 
 void microbit_music_tick(void) {
+    if (music_data == NULL) {
+        // music module not yet imported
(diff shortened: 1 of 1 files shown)


markshannon:
I'm surprised (and a bit worried) that 
a) I didn't catch that, and
b) it crashes randomly.
It ought to crash the instant `microbit_music_tick` is called.

dpgeorge:
> b) it crashes randomly.

Yeah, I didn't get to the bottom of it.  But note that on bare-metal there's no such thing as a seg-fault, you're allowed to read from pretty much any memory location.  So it was probably reading the current `async_state` from location 0x00000008 and that location would eventually become non-zero and then it would go into the rest of the music ticker callback and then crash.

ZanderBrown:
I see @dpgeorge has been busy today! Congrats on all the great work!

markshannon:
Location 0x00000008 is in flash, so shouldn't be able to change. It's a mystery...

PR #443, opened 2017-05-07
Enable __class__ and __name__ to allow introspection.

markshannon:
Without these µPython on the µBit feels a bit too "micro" :smile: 
See https://github.com/bbcmicrobit/micropython/issues/36#issuecomment-299694729 for motivating example.

Increases the size of the executable by just under 300 bytes. Once the DAL is gone, there is still plenty of room for the filesystem.


deshipu:
In that example, you could instead either: pass a string as the source in the first place, since you will want a string pretty much every time; or compare the src to microbit.button_a directly, without having to use strings at all.

markshannon:
Although the macro is called `MICROPY_CPYTHON_COMPAT`, `__name__` and `__class__` are not a CPython implementation detail, they are part of the language. I think we ought to support them.

Just because you can use strings doesn't mean that you should. Using strings prevent type-checking.
`src == microbit.buton_a` will raise an `AttributeError`, but `src = "buton_a"` will just fail silently.

markshannon:
Although the macro is called `MICROPY_CPYTHON_COMPAT`, `__name__` and `__class__` are not a CPython implementation detail, they are part of the language. I think we ought to support them.

Just because you can use strings doesn't mean that you should. Using strings prevent type-checking.
`src == microbit.buton_a` will raise an `AttributeError`, but `src == "buton_a"` will just fail silently.

PR #455, opened 2017-06-22
Fix some mistakes in the in-code documentation of neopixel.h

mathisgerdes:
Item order was wrong and indices were falsely claimed to start with index 1 (start with 0).


dpgeorge:
Thanks for the contribution.  Note that this code comes straight from an upstream source https://github.com/lavallc/nrf51-neopixel so ideally these issues would be fixed there first, and then we could pull them in to this repository.  Otherwise maintenance of the code here becomes more difficult if/when the neopixel driver needs updating from upstream.

PR #457, opened 2017-06-30
Update documentation for PWM period minimum microseconds value

carlosperate:
Noticed the documentation wasn't correct, so just a very quick edit to fix that.
Link to the source code clamp value: [source/lib/pwm.c#L181](https://github.com/bbcmicrobit/micropython/blob/9cef1292df38649ad7f11e0c4b8c27bbacda6c3d/source/lib/pwm.c#L181)


dpgeorge:
Thanks!

PR #468, opened 2017-09-14
Slightly reduce the heap size when building with newlib >2.4.0

### LINKED ISSUE #363: build error: "region RAM overflowed with stack"
Running the latest available gcc-arm-none-eabi 5.4.1 for debian-based systems (https://launchpad.net/~team-gcc-arm-embedded/+archive/ubuntu/ppa), I am basically running out of memory on a clean master:

```
/usr/bin/../lib/gcc/arm-none-eabi/5.4.1/../../../../arm-none-eabi/bin/ld: region RAM overflowed with stack
collect2: error: ld returned 1 exit status
ninja: build stopped: subcommand failed.
error: command ['ninja'] failed
```

I basically have to reduce 320 bytes from the heap:

``` c
// allocate the heap statically in the bss
static uint32_t heap[9692 / 4];
```

to:

``` c
static uint32_t heap[9372 / 4];
```

This is quite a lot of memory, has anybody else experienced this?

jimmo:
Fixes #363

### DIFF
--- a/source/microbit/main.cpp
+++ b/source/microbit/main.cpp
@@ -123,7 +123,7 @@ int main(void) {
 #if __NEWLIB__ > 2 || (__NEWLIB__ == 2 && (__NEWLIB_MINOR__ > 4 || (__NEWLIB_MINOR == 4 && __NEWLIB_PATCHLEVEL > 0)))
         // newlib >2.4.0 uses more RAM for locale data.
         // https://github.com/bbcmicrobit/micropython/issues/363
-        static uint32_t mp_heap[10052 / sizeof(uint32_t)];
+        static uint32_t mp_heap[10020 / sizeof(uint32_t)];
 #else
         static uint32_t mp_heap[10240 / sizeof(uint32_t)];
 #endif
--- a/source/microbit/main.cpp
+++ b/source/microbit/main.cpp
@@ -6,6 +6,7 @@
 #include "microbit/microbitobj.h"
 #include "microbit/microbitbutton.h"
 #include "microbit/modmusic.h"
+#include "_newlib_version.h"
 
 // Global instances of the DAL components that we use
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks @jimmo for this, it's a neat way to support older and newer newlib versions.  But this really would be a temporary solution because it's such a waste of RAM!  Can you confirm that the size you have for the heap leaves a little bit of room in RAM, eg around 32 bytes?  So that things don't have to be changed again if/when more RAM is needed for the data/bss sections.

Ping @jaustin we really should fix this issue with newlib...

jimmo:
Good point - 10052 was exactly what it needed to fit, so have adjusted to 10020.

Definitely temporary, but I figured that for someone new to the project being able to build without a weird error was worth 220 bytes of heap.

Alternatively, we could put a useful `#error` instead with a link to the bug?

Either way, perhaps a `#warn` would be appropriate?

dpgeorge:
> Good point - 10052 was exactly what it needed to fit, so have adjusted to 10020.

Great, thanks!

> Either way, perhaps a #warn would be appropriate?

Yes, that's a good idea.  Can you please add one and verify it gives a warning for you?  (and doesn't error out, ie the build still completes)

jimmo:
On 15 September 2017 at 18:20, Damien George <notifications@github.com>
wrote:

> Good point - 10052 was exactly what it needed to fit, so have adjusted to
> 10020.
>
> Great, thanks!
>
> Either way, perhaps a #warn would be appropriate?
>
> Yes, that's a good idea. Can you please add one and verify it gives a
> warning for you? (and doesn't error out, ie the build still completes)
>
Done.

dpgeorge:
Thanks, merged in 498125d2619f5f7a48026f12d6c984dd476643d7

PR #473, opened 2017-10-09
Fix end-string in uart.rst.

MinoruInachi:


dpgeorge:
Thanks for that!  Merged.

PR #485, opened 2018-01-14
Allow floats and integers in display.show and arbitrary objects in display.scroll

mpbagot:
This is implemented in display.scroll and display.show by calling the `mp_obj_str_make_new` function on the objects which aren't normally accepted.

### DIFF
--- a/source/microbit/microbitdisplay.cpp
+++ b/source/microbit/microbitdisplay.cpp
@@ -31,6 +31,7 @@ extern "C" {
 
 #include "py/runtime.h"
 #include "py/gc.h"
+#include "py/objstr.h"
 #include "lib/iters.h"
 #include "lib/ticker.h"
 #include "microbit/modmicrobit.h"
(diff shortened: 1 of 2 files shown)


dpgeorge:
+1 from me.  It's backwards compatible and helps to eliminate things like `display.show(str(123))` which is quite common.

@jaustin what do you think?

ZanderBrown:
Seems a good addition, I often think of `display.show`/`scroll` be equivalent to `print` so the fact it errors on non-strings has always been a little odd

source/microbit/microbitdisplay.cpp
```diff
@@ -432,7 +439,13 @@ mp_obj_t microbit_display_scroll_func(mp_uint_t n_args, const mp_obj_t *pos_args
     mp_arg_val_t args[MP_ARRAY_SIZE(scroll_allowed_args)];
     mp_arg_parse_all(n_args - 1, pos_args + 1, kw_args, MP_ARRAY_SIZE(scroll_allowed_args), scroll_allowed_args, args);
     mp_uint_t len;
-    const char* str = mp_obj_str_get_data(args[0].u_obj, &len);
+    mp_obj_t object_string;
+    if (!MP_OBJ_IS_TYPE(args[0].u_obj, &mp_type_str)) {
```
dpgeorge: I think that scroll should behave the same as show here, in the sense that it should only accept ints, floats and strings (not arbitrary objects).  Converting objects like lists, tuples and dicts to a string to show them is arguably not what you want as the programmer, and conflicts with show which would take a list/tuple and show each element of the list/tuple.

PR #493, opened 2018-02-18
Update storage.rst

Spadgerian:
print does nothing on the micro:bit. Replaced with display.scroll()
display.scroll(str(len(...))) or display.scroll(str(...)) added to aid displaying result

### DIFF
--- a/docs/tutorials/storage.rst
+++ b/docs/tutorials/storage.rst
@@ -59,7 +59,7 @@ The best way to make sure of this is to use the ``with`` statement like this::
 
     with open('story.txt') as my_file:
         content = my_file.read()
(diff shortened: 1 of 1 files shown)


deshipu:
You are mistaken. `print` does something: it outputs the text in the REPL, where it is much easier to read than on the scrolling display of the device.

whaleygeek:
print() on micro:bit sends data to the serial port. This comes out in the
REPL window in Mu.

Spadgerian:
Ok, sorry. I must have missed the REPL descriptions in the introduction.

deshipu:
Thank you for the time you took to make those fixes though!

PR #495, opened 2018-02-19
Update storage.rst

Spadgerian:
Is this demo code any good here?


dpgeorge:
Thanks for the contribution.  But I don't think this really adds anything that's not already in the tutorial.  All the functions are already explained and have examples.

So I'd suggest to close this PR.

Issue #237, opened 2016-03-13
Beeps and Boops - Piezo's need a square wave


descartes:
That sort of buzzer would need to be energised by a square wave around 2kHz - follow the link you've given for SparkFun to confirm.

You can either use a low current buzzer - I'll have a look for one. Or use the PWM output on a pin to drive the piezo. Or use suitable sleep values in a loop to switch the pin on & off at some reasonable rate (see below).

For this easy to understand example, the buzzer would be the best route.

The piezo will actually run at a wide range of frequencies but the element is tuned for a particular frequency - I use 2kHz piezo's and can get a reasonable tune out of them (middle C is 262Hz), but the volume goes right up when you hit the tuned frequency.

dpgeorge:
I think this "Beeps and Boops" tutorial is talking about a buzzer that just requires power to make a (fixed freq) tone.  You are right that the image of the piezo refers to something that is not a "buzzer".  @ntoll do you know which piece of hardware you used for this tutorial?

ntoll:
Greeting from Bratislava! OK... this is my mistake, and many apologies. I Googled for Piezo buzzer and used the (liberally licensed) image found as a result. It certainly looks like the hardware I used for this tutorial but I'll need to check the details when I return home. Mine is certainly a fixed pitch buzzer.

cefn:
I was just about to file this as an issue and found this thread. To clarify, the difference is whether the component is a Piezo transducer/speaker/passive buzzer (voltage input translates to movement) or an active buzzer (voltage input drives an oscillating circuit, which generates a voltage which translates to movement).

This example is especially problematic since housed Piezo transducers are a common way of wiring up a circuit suitable for e.g. http://microbit-micropython.readthedocs.io/en/latest/tutorials/music.html as we do with our Alarm Clock... http://start.shrimping.it/project/alarmclock/build.html#step12 despite the misleading warning in the Music tutorial...

```Do not attempt this with a Piezo speaker - such speakers are only able to play a single tone.```

...which should probably also be removed as using a Piezo transducer is fine and probably recommended as they are cheap, not too demanding of current, and ideal for classroom wiring.

If you wanted to reference a Piezo transducer in the music example, we reference our suppliers for every kit component e.g. links from http://start.shrimping.it/kit/alarmclock.html From experience, there seems to be very little consistency in naming among cheap suppliers. In the past I've bought elements which I expected to be transducers and they turned out to be active buzzers

[...]

PR #520, opened 2018-06-04
Unified and expanded PR for piezo buzzer issues.

ScottDWebster:
Here's a single pull request encompassing my previous requests plus additional text.


dpgeorge:
Thanks for the contribution.  Actually, this was already "fixed" once, see #237 and #285.  But if it's still causing confusion then maybe it does need further clarification.

dpgeorge:
Squashed and merged in e8df727c7233f35f183a8405a52525122b25102b

PR #521, opened 2018-06-05
Added note about ufs get and put

ScottDWebster:
I was scratching my head for a while trying to figure out why the next example utilizing hello.py and main.py wasn't working until I realized that "ufs put hello.py main.py" was only copying hello.py but saving it as main.py on the remote filesystem.  This note will hopefully help others avoid such confusion.

### DIFF
--- a/docs/tutorials/storage.rst
+++ b/docs/tutorials/storage.rst
@@ -210,6 +210,9 @@ Finally, the ``put`` sub-command puts a file from your computer onto the
 connected device (it's named after the ``put`` command that's part of FTP that
 serves the same function).
 
+.. note::
+    The ufs get and put commands only operate on one file at a time.
+
 Mainly main.py
 ++++++++++++++
 


dpgeorge:
> the next example utilizing hello.py and main.py wasn't working until I realized that "ufs put hello.py main.py" ...

Is there an example in the docs/tutorial that uses this format, `ufs put <src> <dest>`?  Why did you try using this form?  It might be better to improve the docs surrounding the extended form of ufs-put instead of adding this note here.

ZanderBrown:
I think @ScottDWebster expected put to work like unix `cp`

i.e.
```
cp sourcefile... targetdirectory
```
But found it actually works as:
```
ufs put sourcefile targetfile
```

dpgeorge:
> I think @ScottDWebster expected put to work like unix cp

Ok I see.  In that case I think it's better to expand the docs a bit and actually state this extended behaviour of ufs put/get, with the additional arg being the target file name.

ScottDWebster:
Yes, I had cp more in mind that ftp.

The microfs documentation spells out the get/put behavior:

The put and get commands optionally take a further argument to specify the name of the target file:
~~~~
$ ufs put /path/to/local.txt remote.txt
$ ufs get remote.txt local.txt
~~~~

I was trying to keep it succinct, but combining a note about the fact that ufs only acts on one file at a time with an expanded explanation of local/remote paradigm would be even better.

I'll think about how best to word that and submit a new PR.

dpgeorge:
This note is useful in its own right, so merged in f524840ae831f36cef5abfd19240de2054496939

Issue #525, opened 2018-07-10
Priority order to execute user code (embedded vs main.py)


microbit-carlos:
With the introduction of the filesystem we had the option to run a main.py file on startup. The current implementation prioritises a script saved at the known flash location (0x3E_0000) and I think we should probably reverse this order to prioritise main.py.

The reason this has come to light is because editors like Mu have started to flash the micro:bit using the filesystem (thanks @ntoll!), and we can have a situation where a micro:bit has MicroPython + embedded script, then a main.py file is then transferred, and the user would still see the old code running on the micro:bit.

It would also be really hard for an editor to determine if the micro:bit contains an embedded script in flash.

I've created a proposed PR for this change and gave it a quick test (embed script, load main.py and erase main.py), and seems to do the trick.

@dpgeorge what do you think? Would there be any other implications if this was changed?

dpgeorge:
> and we can have a situation where a micro:bit has MicroPython + embedded script,

How does this situation arise?  I thought Mu had a bare image in it (ie no appended script).  Does it come about when users download firmware with an appended script (maybe from the browser), and then reuse that device with Mu in filesystem-mode?

> Would there be any other implications if this was changed?

It does change behaviour, and likely there's someone out there relying on this behaviour (maybe unintentionally).  But I doubt it's a common thing, so it would be good to make it work more "as expected".  Since adding an appended script is only something that can be done when you flash the entire image, but changing main.py on the filesystem can (must!) be done after flashing the image, and can be done many times, it does make sense that main.py overrides the appended script.  After all, downloading main.py is an explicit, user-requested operation so they really must want to run the new main.py rather than the old one.

> It would also be really hard for an editor to determine if the micro:bit contains an embedded script in flash.

Using recent firmware it is actually very easy:
```python
import machine
have_appended_script = machine.mem16[0x3e000] == 0x504d
```
So this could be used as a way to warn the user that there exists an appended script.

But in the interest of simplicity it's probably a good idea to just make the filesystem main.py take precedence over any appended script.

ntoll:
That's a nice trick with the `machine` module.

microbit-carlos:
> How does this situation arise? I thought Mu had a bare image in it (ie no appended script). Does it come about when users download firmware with an appended script (maybe from the browser), and then reuse that device with Mu in filesystem-mode?

Yes, that would be one of the main ways this could happen. 

[...]

PR #526, opened 2018-07-10
Change the order of precedence for user code execution

microbit-carlos:
Proposed change from https://github.com/bbcmicrobit/micropython/issues/525.

The old behaviour prioritises code embedded in the known flash location over a main.py file in the file system. This change reverses that order.

### DIFF
--- a/source/microbit/main.cpp
+++ b/source/microbit/main.cpp
@@ -166,11 +166,11 @@ int main(void) {
         // mode.  If we are in "raw REPL" mode then this will be skipped.
         if (pyexec_mode_kind == PYEXEC_MODE_FRIENDLY_REPL) {
             file_descriptor_obj *main_module;
-            if (APPENDED_SCRIPT->header[0] == 'M' && APPENDED_SCRIPT->header[1] == 'P') {
+            if (main_module = microbit_file_open("main.py", 7, false, false)) {
+                do_file(main_module);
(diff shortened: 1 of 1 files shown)


source/microbit/main.cpp
```diff
@@ -166,11 +166,11 @@ int main(void) {
         // mode.  If we are in "raw REPL" mode then this will be skipped.
         if (pyexec_mode_kind == PYEXEC_MODE_FRIENDLY_REPL) {
             file_descriptor_obj *main_module;
-            if (APPENDED_SCRIPT->header[0] == 'M' && APPENDED_SCRIPT->header[1] == 'P') {
+            if (main_module = microbit_file_open("main.py", 7, false, false)) {
```
dpgeorge: Please use extra parenthesis around this expression, like it was done below.

dpgeorge:
I think this is a good change.

dpgeorge:
Should we put this on the master branch or the auto-detect-sensors branch?

source/microbit/main.cpp
carlosperate: Done.

microbit-carlos:
I guess that depends on what branch we'll use to create the RC once MicroPython is feature-frozen.
Are we going to merge to master? or continue working on the auto-detect-sensors branch?

Also, as a side note, could we git tag the commit used for v1.0.0-beta.1?

carlosperate:
Thanks Damien!

dpgeorge:
Squashed and merged in 5673bb09c0a77670ace70b425826fea47ad79684

PR #528, opened 2018-07-11
Make sure the display is active upon soft reset.

### LINKED ISSUE #527: The state set by display.off() isn't reset on soft reboot.
Steps to recreate:

```
>>> from microbit import *
>>> display.scroll("Test")
>>> display.off()
```

Press CTRL-D for a soft reset.

```
>>> from microbit import *
>>> display.scroll("Test")
```

At this point nothing happens, and I was expecting to see "Test" scrolled across the display.

For reference, first reported here: https://github.com/mu-editor/mu/issues/537

dpgeorge:
This fixes #527.

But it raises a few more issues:
- Should the display be completely cleared on soft reset?  At the moment it retains what it had on it before the soft reset.
- Should pin modes be reset on soft reset?  Currently they are not and this can lead to some funky things like having music and display on the same pin, eg:
```python
import microbit, music
microbit.display.show('a')
microbit.display.off()
music.pitch(440, pin=pin3)
# do a soft reset with ctrl-D
import microbit, music
music.pitch(440, pin=pin3)
# now the "a" still shows and music tries to play on pin3
```

### DIFF
--- a/source/microbit/microbitdisplay.cpp
+++ b/source/microbit/microbitdisplay.cpp
@@ -529,11 +529,10 @@ mp_obj_t microbit_display_on_func(mp_obj_t obj) {
     microbit_obj_pin_acquire(&microbit_p7_obj, microbit_pin_mode_display);
     microbit_obj_pin_acquire(&microbit_p9_obj, microbit_pin_mode_display);
     microbit_obj_pin_acquire(&microbit_p10_obj, microbit_pin_mode_display);
-    /* Make sure all pins are in the correct state */
+    /* Make sure all pins are in the correct state, and also re-enable the
+     * display loop.  This will resume any animations in progress and
+     * display any static image. */
     microbit_display_init();
-    /* Re-enable the display loop.  This will resume any animations in
-     * progress and display any static image. */
-    self->active = true;
     return mp_const_none;
 }
 MP_DEFINE_CONST_FUN_OBJ_1(microbit_display_on_obj, microbit_display_on_func);
@@ -668,6 +667,8 @@ microbit_display_obj_t microbit_display_obj = {
 void microbit_display_init(void) {
     //  Set pins as output.
     nrf_gpio_range_cfg_output(MIN_COLUMN_PIN, MIN_COLUMN_PIN + COLUMN_COUNT + ROW_COUNT);
+    // Allow the display to update itself on the ticker callback.
+    microbit_display_obj.active = true;
 }
 
 }


ntoll:
Heh.... love the "funky" things.

After talking to users of Mu, they were expecting a soft reset to actually reset the state of the board, hence their original bug report to Mu since they thought this behaviour problematic. Ergo, from my PoV and Mu users' points of view the two "should X be reset on soft reset?" questions are answered "yes, absolutely".

Hope this helps.

dpgeorge:
If the display is cleared on soft reset this may have an impact on users' existing expectations.  For example, if they have something on the display and Mu transfers files to/from the device then it will clear the display because Mu does a soft reset behind the scenes (I assume).  I don't know if this would be an issue but worth thinking about.

ntoll:
Correct, Mu does a soft reset (see: https://github.com/mu-editor/mu/blob/master/mu/modes/microbit.py#L466).

OK... so given the behaviour of "Flash" up until a few days ago was to completely re-flash and reset the device, I believe the users' expectations will remain the same. If they have something on the screen and "new Flash" merely copies the file over and soft-resets the device, then (just as before) the device should start from a "clean" fully reset state, IYSWIM. In fact, if it doesn't (as in this case) we're likely to get bug reports into Mu asking why the behaviour has changed. ;-)

dpgeorge:
> OK... so given the behaviour of "Flash" up until a few days ago was to completely re-flash and reset the device, I believe the users' expectations will remain the same.

Ok, so it sounds like there are 2 separate (but related) issues here:
1. How soft reset should behave, should it reset everything or just some things.
2. How Mu behaves when it updates the device.

For (2), if you really want to make it act like it did before, then I'd suggest issuing a `microbit.reset()` command after copying over main.py.  That will give a truly clean state (and will execute main.py which I guess is what you want).

ntoll:
Aha... I'd forgotten about `microbit.reset()` I'll use that instead. This is probably the best solution since it keeps the soft-reset behaviour the same while allowing Mu to retain the same behaviour. Thanks!

dpgeorge:
So, do we still need the fix in PR?  I'd suggest we do because it will potentially lead to confusion that after soft reset the display doesn't work.

ntoll:
+1

dpgeorge:
Mu now uses a full reset to get its expected behaviour after copying a main.py to the microbit.

So there is no need for the patch in this PR, and since it introduces a change in behaviour it's not going to be merged.

PR #533, opened 2018-07-15
Allow to change pin in touch mode to other mode without error.

dpgeorge:
Since touch mode is pretty much just digital input, it makes sense to be able to change from touch mode to other standard GPIO modes (read digital, write digital, read analog) without an error (you can already change from read digital to write digital without error).

Without a patch like this there's no way to get a pin out of touch mode once `pin.is_touched()` is called.


carlosperate:
Haven't had a chance to test this yet, but the change looks good to me.

dpgeorge:
This was merged in eeb9a2d2e4ec7d49354913fcafc990822430c662

Issue #536, opened 2018-07-18
Adafruit 4x14-segment I2C display code no longer working


SteveJSmythe:
Mu 1.0.0.rc.1
Windows 10

[Code that I wrote](http://www.smythe-consulting.com/2017/04/driving-adafruit-4x14-segment-i2c.html) to drive the Adafruit 4x14-segment I2C display no longer works. 

I rebuilt the circuit as per the instructions on my blog and it works with Mu 0.9.13. If I reflash with Mu 1.0.0.rc.1, the display does not light up at all. No error messages are generated.

This will be a tricky one to diagnose as you will need an identical display to test it. 

Whether it is relevant or not, I do not know but using an almost identical circuit with the Adafruit 8x8 LED matrix (the difference being that the 4x14-segment display has an additional power connection - to the micro:bit's 3.3v logic level). In that case, the software is written slightly differently as I load the code AND a separate module for the HT16K33 LED driver/controller.

I am happy to conduct any tests or measurements to help you solve this one!

dpgeorge:
Thanks for the report.  Can you please post here the exact code that no longer works?  Maybe by inspection of the code we can work out what is wrong.

dpgeorge:
I had a long look at the above I2C script, was able to reproduce the error and see exactly the problem.  The issue is again memory fragmentation, and that the pattern of memory allocations has changed with v1.0.0-rc.1 compared with Mu v0.9.3.  The script is large but it's not directly running out of memory, it's just that the memory is fragmented.

I've managed to tune some of the settings of MicroPython, and reduce memory consumption of the parser by a little, and that got the script running.  In fact the same modifications also fixed #535.  But I would be cautious about using these new settings because it may stop other things from working.  With some more work I may be able to find better settings to improve things overall.

There was also another "bug" I ran across: an error is not scrolled on the display if it occurred during the compilation of the script, which is why the above I2C script was failing without showing anything on the display.  This is easy to fix and shouldn't have any side effect.

SteveJSmythe:
Great debugging!

dpgeorge:
Fixed by #539

SteveJSmythe:
And my display now works again with the original code compiled with Mu 1.0.0! Thanks again.

microbit-carlos:
Glad it works again :)
And thank you for testing Steve, your large scripts helped tremendously!

PR #538, opened 2018-07-19
microbit: Scroll any exception on the display even if no traceback info.

dpgeorge:
If an exception happens during script compilation there will not be any traceback information, yet it's still important to display such an error on the display.

I made this patch have as few lines changed as possible, hence the `if (1)`.


microbit-carlos:
PR LGTM 👍 
Tested the patch with https://github.com/bbcmicrobit/micropython/issues/536 and works like charm, thanks Damien!

Issue #535, opened 2018-07-18
Memory allocation error in Script that run in older versions of MicroPython


microbit-carlos:
From: https://github.com/mu-editor/mu/issues/565

> Mu 1.0.0.rc.1
> Windows 10
> 
> The attached code generates the following error "Line 16 Memory Error memory allocation failed allocating 2816 bytes". No error is generated when flashing the same code using Mu 0.9.13. The program is test code for a dual 4x7-seg display with MAX7219 driver.
> 
> As the program is quite long by micro:bit standards, it is possible that the new version of Mu just takes up slightly more memory than the old version and runs out of space.
> 
> [7-seg test no library with functions 2.zip](https://github.com/mu-editor/mu/files/2205154/7-seg.test.no.library.with.functions.2.zip)

Thank you @SteveJSmythe for the testing and reporting the issue!

microbit-carlos:
I have tested the code in 3 versions of MicroPython, nothing attached to the micro:bit:
* MicroPython v1.0.0-rc.1
    * From Mu v1.0.0.rc.1
    * I can replicate the allocation error:
        ```
        Traceback (most recent call last):
          File "__main__", line 16, in <module>
        MemoryError: memory allocation failed, allocating 2816 bytes
        ```
* MicroPython v0.9
    * From http://python.microbit.org/v/1
    * I can replicate a similar allocation error:
        ```
        Traceback (most recent call last):
          File "__main__", line 16, in <module>
        MemoryError: memory allocation failed, allocating 3840 bytes
        ```
* MicroPython from unknown commit (slightly newer than v0.9) 
    * From Mu v0.9.13
    * MicroPython from ntoll/uflash@fe26fe8
    * Not sure which MicroPython commit was used, based on the date (23rd of October 2016). It's likely slightly ahead of the MicroPython v0.9 (effc07c2571bf8c388bf25ee1392cc8a1f535aac 3rd October 2016), possibly 525d88c73f2b81d319efa3889cf5424ebfdc1770?
![image](https://user-images.githubusercontent.com/29712657/42877484-b5e269c6-8a81-11e8-9322-08445db0538f.png)
    * This version, as indicated in the bug report, doesn't get an allocation error.

dpgeorge:
I can confirm that with Mu v0.9.13 the 7-seg script works, but with v1.0.0-rc.1 it doesn't.

[...]

PR #539, opened 2018-07-19
Tune parser memory usage

dpgeorge:
There are 2 commits here:

### first commit

py/parse2: Make rule_stack_t take only 8 bytes per entry, not 12.

Takes less memory when parsing.  Restrictions introduced by this change
are:
- maximum parse tree size is 65535 bytes (was 4G)
- maximum number of expressions in a tuple, list, dict, etc is 65535
  (was 4G)

Given the amount of memory on the device is so small, these new limits
should never be reached.

### second commit

microbit: Tune parser memory allocation policy.

Combined with the previous change to rule_stack_t, this makes sure that
the number of bytes allocated for the parser's rule stack is unchanged for
the initial allocation, and for subsequent resizes of the stack.

What does change is that now the rules stack can hold more entries for the
same number of bytes.  The result is that there will be less resizes of the
rule stack.

### summary

This should make the memory allocation patterns of code running under the latest version similar to v0.9 (before updating to upstream MicroPython 1.9.2).

This should be a relatively low risk change, but would still need some testing.

### DIFF
--- a/inc/microbit/mpconfigport.h
+++ b/inc/microbit/mpconfigport.h
@@ -33,6 +33,8 @@
 
 // memory allocation policies
 #define MICROPY_ALLOC_GC_STACK_SIZE (32)
+#define MICROPY_ALLOC_PARSE_RULE_INIT (96)
+#define MICROPY_ALLOC_PARSE_RULE_INC (24)
 #define MICROPY_ALLOC_PATH_MAX      (64)
 #define MICROPY_QSTR_BYTES_IN_HASH  (1)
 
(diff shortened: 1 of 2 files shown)


dpgeorge:
With this patch, both #535 and #536 are fixed (the scripts run without a MemoryError).

microbit-carlos:
What kind of testing would this feature need to be considered ready?
Would that be finding more user scripts that are on the edge of memory usage? (Like the ones @SteveJSmythe has kindly tested for us, thanks again!) 
Or are there some tests we could write ourselves?

dpgeorge:
I've now done some quite extensive testing of these patches.  The upstream MicroPython test suite (tests/basics/) runs just fine.  All the micro:bit specific tests in this repo run fine.  All the example scripts I have for the micro:bit seem to work okay, even the really large ones.  I couldn't find anything that fails, that didn't already fail on Mu v0.9.3.

So I merged this PR in 4c3cf36ab65a3168ebc7099403ae1a7404dac236 and d2311ae104b51d9b9de4d7cb894902f1a9b6ca5a

PR #550, opened 2018-09-01
microbit: Configure and enable handling of soft reset button.

### LINKED ISSUE #546: Reset button is ignored on battery power
When on battery power pressing reset does nothing with 
versions 1.7.9 that was shipped with mu 1.0.0 beta 12 and
MicroPython v1.9.2-34-gd64154c73 on 2017-09-01; micro:bit v1.0.0-rc.2 with nRF51822
Whilst using site https://python.microbit.org/v/1 or the equivalent program in microsoft makecode blocks.  The reset button works on battery

dpgeorge:
Should fix #546
 
Note: I used mbed's `gpio_t` and `gpio_irq_t` directly to minimise RAM usage (otherwise using `InterrutpIn costs 80 bytes of RAM).

### DIFF
--- a/source/microbit/main.cpp
+++ b/source/microbit/main.cpp
@@ -4,8 +4,11 @@
 #include "microbit/filesystem.h"
 #include "microbit/microbitdal.h"
 #include "_newlib_version.h"
+#include "MicroBitButton.h"
 
-// Global instances of the DAL components that we use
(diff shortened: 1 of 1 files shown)


microbit-carlos:
Tested it and works as discussed in #546 (same as DAL, on battery reset happens immediately on button press down).
LGTM, thanks Damien! 👍

Issue #556, opened 2018-09-09
Using Microbit pins 5 and 11 as inputs gets Error pin in button mode


4tronix:
Bit:Bot uses pins 5 and 11 as digital inputs and all existing code simply calls the read digital functions. This now gives an error stating the pin is in button mode, but it has worked fine for the last 2 years. How do we change pin modes? I can’t find documentation for it anywhere

DavidWhaleMEF:
@carlosperate @dpgeorge any ideas? If you could point us at the general area or change that caused this, I can probably bake up a workaround for @4tronix so that their line following buggy can be made to work again, thanks!

DavidWhaleMEF:
@4tronix Just a side note, we get a number of tickets logged about bit:bot keep going into pairing mode. This is because pairing mode is entered by holding buttons A and B when powering up or resetting. So if the bit:bot is in a state where both inputs are active on power up (presumably on the floor) and a .hex from MakeCode has been loaded (which has the Bluetooth DFU present), it enters the Bluetooth pairing mode app.

If not done already, it might be worth you signposting this in your docs? Thanks!

4tronix:
Yes, it is in big red letters at the top of 4tronix.co.uk/bitbot
For the last iteration (v1.2) I have also delayed the activation of the sensors for 2 seconds to prevent this.

Best regards,
Gareth

dpgeorge:
> This now gives an error stating the pin is in button mode, but it has worked fine for the last 2 years

Actually, it was about 2 years ago that this was changed!  See: ce36e0dea1c23f0608493978c839707386f34ce6

So it's strange that this issue only just came up now.

> How do we change pin modes?

It's not possible, at least not for pins attached to buttons.  The only way to read the current digital value of a button is to use `button_a.is_pressed()`.

---

Probably this behaviour should be changed back to the old behaviour that did allow to read a button like a digital IO pin.  I made a PR for that, see #557

4tronix:
I guess it is because this version has only now been incorporated into a 
released version of mu editor which people use.

It would be good if the behaviour could be changed back so that the 
existing line follower code can continue working.

In the meantime I will tell customers to convert to 
button_a.is_pressed() usage.

------ Original Message ------

microbit-carlos:
Thanks Damien for the quick turnaround on the fix. The PR adds the ability to use `read_digital` (and `set/get_pull`) on a pin in button mode.
[...]

PR #557, opened 2018-09-10
microbit: Allow pins in button mode to work as a digital-in pin would.

dpgeorge:
So that read_digital(), set_pull() and get_pull() work on pin5 and pin11.

See issue #556


dpgeorge:
@carlosperate I think we should merge this change for the next rc (or final?) release.

Hopefully there are no other valid uses of pins that still raise an error about the pin mode.

microbit-carlos:
Agreed, we should merge this for the next release.
We also have updates to the DAL, with the changes in the allocator already included: https://github.com/lancaster-university/microbit-dal/compare/dal-integration-2...dal-integration-5

Tested this PR and all looks good 👍 
Although, as expected, if the user sets the pull down the digital reads are a bit wonky since the external pull is 10k and internal 13ishK, so the pin ends up at 1.8ishV.

dpgeorge:
Thanks @microbit-carlos for testing.

> Although, as expected, if the user sets the pull down the digital reads are a bit wonky since the external pull is 10k and internal 13ishK, so the pin ends up at 1.8ishV.

I think this is ok.  The user should know what they are doing if they use `set_pull()`.  And it can't break the hardware :)

Issue #547, opened 2018-08-21
The links in  your README.md are incorrect


rhubarbdog:
I'm trying to join the mailing list  microbit@python.org (https://mail.python.org/mailman/listinfo/microbit).
when i click on the mailman link the python.org site reports 'no such list microbit'
The informaion i get back from python.org is direct my enquiries to postmaster@python.org when i try this they also report there's no such list.
How do i get involved?

deshipu:
That mailing lists has been closed.

rhubarbdog:
how do i 'register' to become a contributor?
I can program in python and c, my c++ is at a beginners level to program, but i read it quite well.  Happy to help with testing

deshipu:
You don't need to register, simply submit a pull request. If you are not sure about how to solve a particular issue or whether a given feature is desired, create an issue first to discuss it.

Isioviel:
I have submitted a pull request to remove the links to the mailing list, let me know if there are any issues! #560

microbit-carlos:
Changes have been merged (thanks Damien!) so I'll close this issue.

Thank you @rhubarbdog  for reporting the old link and thank you @Isioviel for submitting PR #560 to update it!

@rhubarbdog if you still have any question about how to contribute to this project please let us know and we can help through the process :)

PR #560, opened 2018-09-17
edit readme as issue #547

Isioviel:
deleted link to closed mailing list

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,6 @@ MicroPython for the BBC micro:bit
 
 This is the source code for MicroPython running on the BBC micro:bit!
 
-To get involved with the community subscribe to the microbit@python.org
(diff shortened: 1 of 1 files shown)


carlosperate:
Thank you very much for the submission Sarah!

Perhaps we could replace the link with the form to join the Slack channel: https://tech.microbit.org/get-involved/where-to-find/

What do you think @dpgeorge ?

dpgeorge:
> Perhaps we could replace the link with the form to join the Slack channel:

Yes that's a good idea, as long as that link will be active for the foreseeable future :)

@Isioviel would you mind updating this PR to point to that link instead?  Thanks.

dpgeorge:
Thanks a lot!  Merged in 89278a75218636865a6f921b192ce5a8f545f0e3 and b71b549c3566dcb73af46d7741fb5a16c2f916d8

PR #562, opened 2018-09-17
Docs: Update README to build using Makefile

carlosperate:
Makefile now created the microbitversion.h file needed for yt build to complete successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ sudo apt-get install cmake ninja-build gcc-arm-none-eabi srecord libssl-dev
 pip3 install yotta
 ```
 
-Once all packages are installed, use yotta to build.  You will need an ARM
-mbed account to complete the first command, and will be prompted to create one
-as a part of the process.
+Once all packages are installed, use yotta and the provided Makefile to build.
+You might need need an Arm Mbed account to complete some of the yotta commands,
+if so, you could be prompted to create one as a part of the process.
 
 - Use target bbc-microbit-classic-gcc-nosd:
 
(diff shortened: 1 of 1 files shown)


carlosperate:
Alternatively we could also create a dummy header file to be overwritten by the makeversionhdr.py script.

dpgeorge:
> Alternatively we could also create a dummy header file to be overwritten by the makeversionhdr.py script.

I don't think that would work well because git will always see the dummy file as modified after a build.  I think we've accepted by now that we need to use `make` to build the firmware fully.

carlosperate:
>I don't think that would work well because git will always see the dummy file as modified after a build. I think we've accepted by now that we need to use make to build the firmware fully.

That's a very good point, we'll have to use make.
Does the PR need any changes?

README.md
```diff
@@ -43,16 +43,16 @@ as a part of the process.
...
-repository.
+The resulting firmware.hex file to flash onto the device can be
+found in the build/ directory from the root of the repository.
 
-There is a Makefile provided that does some extra preprocessing of the source,
-which is needed only if you add new interned strings to qstrdefsport.h.  The
-Makefile also puts the resulting firmware at build/firmware.hex, and includes
+The Makefile provided does some extra preprocessing of the source,
+it adds version information to the UICR region, and process new interned
+strings to qstrdefsport.h.
```
dpgeorge: The last part of the sentence here about qstrdefsport.h is no longer true, so can be removed.

carlosperate: 👍 
Commit updated.

dpgeorge:
Thanks for updating, merged in 55e906c4abb629933085fa06693897435823ce62

PR #563, opened 2018-09-18
Tools: Update makecombinedhex to add code between MicroPython and UICR

carlosperate:
Nothing fancy (for instance it expects the UICR data to be placed after MicroPython), but it does the job.


dpgeorge:
I tested this and it works for me (on a microbit with old DAPLink 0234).  Merged in 418b60a2f00096af384065e91f532730c8aad9cf

Thanks!

PR #564, opened 2018-09-19
Change 'import' to 'important'.

ScottDWebster:
In the first paragraph after the code listing of radio.py, the word 'import' was used where I believe the word 'important' was intended, so I have changed it as it is potentially confusing.


microbit-carlos:
Merged, thanks @ScottDWebster for the PR and @deshipu for reviewing!

PR #571, opened 2018-10-04
docs: Update uart.rst to include note on internal buffer size

### LINKED ISSUE #503: Document uart buffer size
Separating this specific request from https://github.com/bbcmicrobit/micropython/issues/472#issuecomment-352729683 to keep better track.

@DavidWhaleMEF wrote:
> Would it be possible to update the docs to say that the buffer is hard coded at 64 bytes?

microbit-rosslowe:
Add note on internal uart buffer size to docs; closes #503

### DIFF
--- a/docs/uart.rst
+++ b/docs/uart.rst
@@ -7,7 +7,8 @@ The ``uart`` module lets you talk to a device connected to your board using
 a serial interface.
 
 .. note::
-    The internal uart buffer is hard coded to 64 bytes.
+    The internal UART RX buffer is 64 bytes, so make sure data is read before 
+    the buffer is full or some of the data might be lost.
 
 Functions
 =========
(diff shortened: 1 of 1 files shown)


docs/uart.rst
```diff
@@ -6,6 +6,8 @@ UART
 The ``uart`` module lets you talk to a device connected to your board using
 a serial interface.
 
+.. note::
+    The internal uart buffer is hard coded to 64 bytes.
```
microbit-carlos: Could we change this to `The internal UART RX buffer is 64 bytes, so make sure data is read before the buffer is full or some of the data might be lost`.

It clarifies a bit what having a fix size buffer means in this case.

microbit-carlos:
Thanks Ross!
LGTM 👍

microbit-rosslowe:
Now changed in #586, so closing this PR.

Issue #416, opened 2017-03-08
music module docs do not match code (music.stop)


DavidWhaleMEF:
The docs say that the music.stop() function has a pin=parameter

![screen shot 2017-03-08 at 20 47 28](https://cloud.githubusercontent.com/assets/25079586/23723290/7eeb8ae2-0440-11e7-9442-277a23e7885a.png)

However, if I use a pin parameter, I get an error.

>>> music.stop(pin=pin1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: function does not take keyword arguments

If I however just put the pin name, it works

music.stop(pin1)

The docs don't match the code.

Apparently using: http://codewith.mu/help/0.9.12

ZanderBrown:
I believe the docs is showing that it has a default value

DavidWhaleMEF:
Thanks. Although it confused me, and I'm supposed to know what I'm doing :)

DavidWhaleMEF:
Remember that this documentation is for a target age group of 11 year old children. Anything we can do to make it more accessible and obvious to them is a good thing :)

ntoll:
Hi @DavidWhaleMEF, to be clear, this is *NOT* documentation for 11yo children, it's documentation for developers, adults and teachers who may support 11yo children and other beginner programmers. 

I've always said (to the BBC at least) that we trust these supervising / supporting people will know their students best and know how to most effectively explain Python. We're just giving them to technical know-how to create their own resources. Stuff like the microbit-playground is more geared to resources beginner programmers may actually end up seeing.

Thanks for all the feedback, keep it coming! :-)

DavidWhaleMEF:
Okay, but we will need documentation for 11 year old children at some point, as they are the target audience. Add it to the list, and I'll help out with it when I can :)

And to be clear, adults (teachers and volunteers) find the whole (*kwargs) style Python documentation very confusing (they tell me as such), so it's not really written for them either?

ntoll:
I hear you!

Here's how things are organised:

* Teachers and volunteers should read the tutorials - they are written for adults who have a basic understanding of computers and if started at the beginning and completed to the end, they'll pick up quite a lot of Python (as well as micro:bit) knowledge too.

* Developers should read the API documentation (which is what this ticket is about). This is for experienced Python developers and assumes familiarity with Python and its idioms. We follow the same documentation style as the main Python docs on python.org.

[...]

PR #572, opened 2018-10-05
docs: Update music.rst to clarify music.stop()

microbit-mark:
As per https://github.com/bbcmicrobit/micropython/issues/416 duplicate help text from Mu and add example.

### DIFF
--- a/docs/music.rst
+++ b/docs/music.rst
@@ -112,9 +112,9 @@ Functions
     frequency is set or ``stop`` is called (see below).
 
 .. py:function:: stop(pin=microbit.pin0)
-
-    Stops all music playback on a given pin. If no pin is given eg.
-    ``music.stop(pin1)``, Pin0 is assumed.
+    
+    Stops all music playback on a given pin, eg. ``music.stop(pin1)``. 
+    If no pin is given, eg. ``music.stop()`` pin0 is assumed.
 
 .. py:function:: reset()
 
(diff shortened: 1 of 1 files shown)


docs/music.rst
```diff
@@ -113,7 +113,8 @@ Functions
 
 .. py:function:: stop(pin=microbit.pin0)
 
-    Stops all music playback on a given pin.
+    Stops all music playback on a given pin. If no pin is given eg.
+    ``music.stop(pin1)``, Pin0 is assumed.
```
microbit-carlos: The second sentence is about not using arguments, but the listed example shows one, which might be a bit confusing. Could we change this to either:

```
Stops all music playback on a given pin, eg. ``music.stop(pin1)``. If no pin is given Pin0 is assumed.
```
Or
```
Stops all music playback on a given pin. If no pin is given eg.``music.stop()``, Pin0 is assumed.
```

I think I prefer the second one.

microbit-carlos:
Thanks Mark, I've added a minor comment, let me know what you think.

docs/music.rst
microbit-mark: I'm not sure the second one resolves the difficulty that David had in the original ticket though. The first one provides an explicit example. How about a combo?

`Stops all music playback on a given pin, eg. ``music.stop(pin1)``. If no pin is given, eg. ``music.stop()`` pin0 is assumed.`

microbit-carlos: Sounds good, thanks Mark!

PR #574, opened 2018-10-05
docs: update music.rst to include clarification on music:pitch

### LINKED ISSUE #417: music.pitch only allows one pitch to be played at a time
The music.pitch module apparently allows you to play a pitch on multiple pins simultaneously.

music.pitch(440, pin=pin0)
music.pitch(600,pin=pin1)
music.stop(pin0)
music.stop(pin1)

At line 1, the tone 440 is generated on pin 0
However, at line 2, the tone 600 is generated on both pin0 and pin1.
At line 3, the tone on pin0 is stopped, and 600 is still generated on pin1
At line 4, the tone on pin1 stops.

It's not clear to me from the docs, which indicate by the fact that you can pitch and stop any arbitrary pins, whether this is *supposed* to allow generation of different tones on different pins or not. If it is not supposed to support it, could you please add some clarification to the docs saying that MicroPython can only generate one tone at a time, and generating tones on two or more pins will only generate the last requested tone. If however this is a bug and it *should* be able to generate different tones on different pins, can you mark it as a bug and add it to the queue? Thanks.

The micro:bit nRF51 processor supports up to three simultaneous PWM modules, but I understand from talking to the Lancaster guys that the period of these units is shared throughout the system. So I *suspect* it is impossible to play multiple pitches simultaneously, but I think the documentation should be clarified to make this point more obvious - the fact that you can start a pitch playing on different pins simultaneously leads the user to 'expect' that the pitch generated on each pin can be different, and it was a surprise to me when it wasn't possible (which caused me to clarify the PWM architecture with the Lancaster team).

Apparently using this editor: https://codewith.mu/help/0.9.12/

microbit-rosslowe:
Add clarification to music:pitch in docs, closes #417 

"Note that you can only play one pitch on one pin at any one time."

### DIFF
--- a/docs/music.rst
+++ b/docs/music.rst
@@ -105,6 +105,8 @@ Functions
     milliseconds. For example, if the frequency is set to 440 and the length to
     1000 then we hear a standard concert A for one second.
 
+    Note that you can only play one pitch on one pin at any one time.
+
     If ``wait`` is set to ``True``, this function is blocking.
 
     If ``duration`` is negative the pitch is played continuously until either the


microbit-carlos:
Thanks Ross!
LGTM 👍

PR #580, opened 2018-10-12
microbit: Allocate all DAL objects on the DAL heap, to save BSS.

dpgeorge:
Then the MicroPython heap can be increased by 128 bytes.


dpgeorge:
This may be useful in the future to reclaim some memory if another feature/fix needs it.  But for now it will not be merged.

PR #584, opened 2018-10-17
docs: Add accelerometer range and units

microbit-sam:
The measurement is given in milli-Newtons, for example +/- 2g will return +/- 2000mN.

### DIFF
--- a/docs/accelerometer.rst
+++ b/docs/accelerometer.rst
@@ -8,6 +8,9 @@ also provides convenience functions for detecting gestures. The
 recognised gestures are: ``up``, ``down``, ``left``, ``right``, ``face up``,
 ``face down``, ``freefall``, ``3g``, ``6g``, ``8g``, ``shake``.
 
+By default MicroPython sets the accelerometer to +/- 2g, and this can only be
+modified in the DAL layer below. The accelerometer returns a value in the range 
+0..1024 for each axis, which is then scaled accordingly.
 
 Functions
 =========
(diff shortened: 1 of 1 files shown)


microbit-carlos:
As you mention, since the range of values will be dependent on the accelerometer sample range configured, could we add a short note about that relationship between the sample range and the range of values returned by these methods?

https://github.com/lancaster-university/microbit-dal/blob/v2.0.0-rc10/source/drivers/MicroBitAccelerometer.cpp#L263-L283

It's true that MicroPython doesn't currently expose a way to change it, so the default of 2g is used, but it would be useful to have in the docs in case the method is added.

Also looking at the DAL docs it sounds like the units is milli-g?

microbit-sam:
👍

docs/accelerometer.rst
```diff
@@ -8,31 +8,36 @@ also provides convenience functions for detecting gestures. The
 recognised gestures are: ``up``, ``down``, ``left``, ``right``, ``face up``,
 ``face down``, ``freefall``, ``3g``, ``6g``, ``8g``, ``shake``.
 
+By default MicroPython sets the accelerometer to +/- 2g, and this can only be
+modified in the DAL layer below. The accelerometer returns a value in the range
```
microbit-carlos: We haven't really introduced the concept of the DAL in these docs, so we should probably rephrase it to something along the lines of:
`and changing this functionality is not currently exposed in MicroPython.`

docs/accelerometer.rst
```diff
@@ -8,31 +8,36 @@ also provides convenience functions for detecting gestures. The
...
 Functions
 =========
 
 .. py:function:: get_x()
 
     Get the acceleration measurement in the ``x`` axis, as a positive or
-    negative integer, depending on the direction.
-
+    negative integer, depending on the direction. The measurement is given in 
+    milli-g, for example +/- 2g will return +/- 2000mg.
```
microbit-carlos: `for example +/- 2g will return +/- 2000mg.`
I know it's been explained at the top of the page, but some people will probably miss it, and the accelerometer range is not really something that has been properly explained in general (it'll probably be new to a lot of people).
Could we rephrase this one (and the others below) to `by default the accelerometer is configured with a range of +/- 2g, and so this method will return a value between +/- 2000mg`?

microbit-carlos:
Thank you Sam!

PR #586, opened 2018-10-19
docs: Update uart.rst with additional detail

### LINKED ISSUE #477: Need more detail in uart docs
Relevant to this issue: https://github.com/bbcmicrobit/micropython/issues/472

The UART docs need more details, such as:
- what is the receive timeout duration
- can you change the receive timeout duration?
- the fact that sending an 0x03 via USBSER will KeyboardInterrupt your program
- whether the CTRL-C processing is enabled or disabled when the UART is redirected to the pins.
- how/if you can change the CTRL-C processing to enable binary data transfer in all of the above (and if it is not possible to change, state the fact that it cannot be changed)

microbit-rosslowe:
- add note on receive timeout duration
- add note on 0x03 keyboard interrupt
- add reference to micropython.kbd_intr() in relation to keyboard interrupts

Closes #477

### DIFF
--- a/docs/uart.rst
+++ b/docs/uart.rst
@@ -13,8 +13,8 @@ Functions
 .. method:: init(baudrate=9600, bits=8, parity=None, stop=1, \*, tx=None, rx=None)
 
     Initialize serial communication with the specified parameters on the
-    specified ``tx`` and ``rx`` pins. Note that for correct communication, the parameters
-    have to be the same on both communicating devices.
+    specified ``tx`` and ``rx`` pins. Note that for correct communication, the 
+    parameters have to be the same on both communicating devices.
 
     .. warning::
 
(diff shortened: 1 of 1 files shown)


docs/uart.rst
```diff
@@ -35,6 +35,11 @@ Functions
         * 57600
         * 115200
 
+    .. note::
```
microbit-carlos: Could we move this note to the read method?
It'd be nice to include as well the calculation (unit is milliseconds) `microbit_uart_timeout_char = 13000 / baudrate + 1`

docs/uart.rst
```diff
@@ -89,3 +94,8 @@ Functions
    Write the buffer of bytes to the bus.
 
    Return value: number of bytes written or ``None`` on timeout.
+
+    .. note::
```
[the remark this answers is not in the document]
microbit-carlos: It's probably better to keep the patch diff small, also not to conflict with the other open PRs.

docs/uart.rst
```diff
@@ -61,39 +58,45 @@ Functions
...
-   Read characters.  If ``nbytes`` is specified then read at most that many
-   bytes.
+    Read characters.  If ``nbytes`` is specified then read at most that many
+    bytes.
+
+    .. note::
+
+        The timeout for all UART reads depends on the baudrate and is otherwise
+        not changeable via Python. The timeout, in milliseconds, is given by: 
+        ``microbit_uart_timeout_char = 13000 / baudrate + 1``
```
microbit-carlos: Actually, could you also combine here the text from https://github.com/bbcmicrobit/micropython/pull/571 ?
We have quite a few notes/warnings in this document, and that note is only really relevant to the read method as well.

microbit-carlos:
Thanks Ross!

PR #587, opened 2018-10-22
docs: Add back EMF warning for speakers

microbit-sam:


microbit-carlos:
We need to revisit the wording and location for these docs. We'll close this for now and reopen/update when we reach a conclusion.

Issue #368, opened 2016-10-29
sending and receiving strings via uart methods


whaleygeek:
microbit.org-ticket: 197

Is there a 'pythonic' way to send and receive strings via the uart object?

e.g. is this the most normal way to encode and decode string data, as the uart methods requires a 'bytes':

msg = "ABC"
uart.write(bytes(msg, "UTF-8"))

m = uart.readall()
msg = str(m, "UTF-8")

ntoll:
I usually just do `msg.encode('utf-8')` etc...

@whaleygeek a quick heads up...

This *isn't* the best place for these questions. Please only use GitHub for raising bugs, pointing out missing docs, asking for features etc... relating to MicroPython on the micro:bit. ;-)

For questions of Python style, best ask on the microbit@python.org mailing list, in either the micropython or microbit channels on IRC (Freenode) or on the micropython room on the official Slack site created by microbit.org.

It's a great question - I just want to make sure it's getting asked in the right place. :-)

whaleygeek:
Okay, so I'll change this into a request to add a note in the docs on the uart page, explaining how to send and receive strings, as it is a common question I get asked. Thanks!

whaleygeek:
@ntoll I couldn't get your example that you gave above to work?

On my Mac:

Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 23 2015, 02:52:03) 
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> msg = 'hello'
>>> msg.encode('UTF-8')
b'hello'
>>> 

Inside the Mu REPL

>>> msg = 'hello'
>>> msg.encode('UTF-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'str' object has no attribute 'encode'
>>>

dpgeorge:
@whaleygeek the way you did it at the start of this issue is correct.  Use `bytes` and `str` to convert between bytes and str objects.  The encode and decode methods are deprecated (or at least legacy) in Python and not implemented in MicroPython.

But note that you can also write the str object directly, like `uart.write('hello')`.

microbit-carlos:
@whaleygeek Do you think a note about this still needs to be added to the docs? (using `bytes` and `str`)
Any suggestions on what to clarify would be most welcomed :)

whaleygeek:
[...]

Issue #504, opened 2018-03-20
Feature request: Flag to indicate uart data buffer overflowed


microbit-carlos:
Separating this feature request from https://github.com/bbcmicrobit/micropython/issues/472#issuecomment-352729683 to keep better track.

@DavidWhaleMEF wrote:
> I was wondering, Is there any opportunity to add something that sets a flag when the UART interrupt tries to write to the 64 byte buffer and there is no more space and junks a byte, and a uart.was_overflow() method that reads and clears the flag? That way, application code would at least have a chance to detect data loss and junk that data and get round to re-syncing and processing the next payload quickly? Just a thought.

PR #588, opened 2018-10-29
docs: Add info about UART read/write data formatting

carlosperate:
Fixes https://github.com/bbcmicrobit/micropython/issues/368.

Adds info about how to format data to send and receive via UART.
It also edits the "Keyboard Interrupt" info to display it as a warning.

@whaleygeek Based on the discussion #368, could you review when you get a chance?

### DIFF
--- a/docs/uart.rst
+++ b/docs/uart.rst
@@ -42,7 +42,7 @@ Functions
     this board.
 
     If ``tx`` and ``rx`` are not specified then the internal USB-UART TX/RX pins
-    are used which connect to the USB serial convertor on the micro:bit, thus
+    are used which connect to the USB serial converter on the micro:bit, thus
     connecting the UART to your PC.  You can specify any other pins you want by
     passing the desired pin objects to the ``tx`` and ``rx`` parameters.
 
@@ -60,24 +60,32 @@ Functions
 
 .. method:: uart.read([nbytes])
 
-   Read characters.  If ``nbytes`` is specified then read at most that many
-   bytes.
+    Read characters.  If ``nbytes`` is specified then read at most that many
+    bytes, otherwise read as many bytes as possible.
 
-   .. note::
+    The return value is a bytes object. You can convert this object to a
+    string, and if there are non-ASCII characters you can also specify the
+    encoding::
+
+        msg_bytes = uart.read()
+        msg_str = str(msg, 'UTF-8')
+
+    .. note::
 
         The timeout for all UART reads depends on the baudrate and is otherwise
         not changeable via Python. The timeout, in milliseconds, is given by:
         ``microbit_uart_timeout_char = 13000 / baudrate + 1``
 
-   .. note::
+    .. note::
 
-        The internal UART RX buffer is 64 bytes, so make sure data is read 
+        The internal UART RX buffer is 64 bytes, so make sure data is read
         before the buffer is full or some of the data might be lost.
 
-   .. note::
+    .. warning::
 
-        Sending ``0x03`` will KeyboardInterrupt your program. You can enable or
-        disable this using :func:`micropython.kbd_intr()`.
+        Receiving ``0x03`` will stop your program by raising a Keyboard
+        Interrupt. You can enable or disable this using
+        :func:`micropython.kbd_intr()`.
 
 .. method:: uart.readall()
 
@@ -102,6 +110,10 @@ Functions
 
 .. method:: uart.write(buf)
 
-   Write the buffer of bytes to the bus.
+    Write the buffer to the bus, it can be a bytes object or a string::
 
-   Return value: number of bytes written or ``None`` on timeout.
+        uart.write('hello world')
+        uart.write(b'hello world')
+        uart.write(bytes([1, 2, 3]))
+
+    Return value: number of bytes written or ``None`` on timeout.
--- a/docs/uart.rst
+++ b/docs/uart.rst
@@ -56,16 +56,23 @@ Functions
 
 .. method:: uart.any()
 
-   Return ``True`` if any characters waiting, else ``False``.
+   Return ``True`` if any data is waiting, else ``False``.
 
 .. method:: uart.read([nbytes])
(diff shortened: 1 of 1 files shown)


DavidWhaleMEF:
LGTM. The examples are short, simple, and to the point, which I think is perfect for these type of docs.

One minor comment - the docs (already) refer to the uart in one place as a 'bus' - strictly it's not a bus, as it is point to point. Should we call this a port instead, perhaps?

DavidWhaleMEF:
Just a minor note, the micropython.kbd_intr() link goes nowhere, where can I read about the parameters that this method takes? If I search for kbd_intr() in the readthedocs search, this is the only page that appears, so I am left dangling without an answer?

screen shot 2018-10-30 at 12 09 20

DavidWhaleMEF:
Also, if I search for 'micropython' in readthedocs, there doesn't seem to be a module of that name documented, and while the word appears in the contents page, it doesn't tell me anything about that specific module, what it's methods are (and do) and what their parameters are.

screen shot 2018-10-30 at 12 12 29

If this is functionality inherited from mainline micropython, is it worth linking over to their docs? Or I though the micropython we have is a special version that only runs on the micro:bit, so cross linking to docs in the core repo could be a bad move?

DavidWhaleMEF:
This is what I get at the REPL when interacting with the V1.0 image:

```>>> help(modules)
NameError: name 'modules' is not defined
>>> import micropython
>>> help(micropython)
object <module 'micropython'> is of type module
  __name__ -- micropython
  const -- <function>
  opt_level -- <function>
  mem_info -- <function>
  qstr_info -- <function>
  stack_use -- <function>
  heap_lock -- <function>
  heap_unlock -- <function>
  kbd_intr -- <function>
>>> 
```

I did a readthedocs search for mem_info and nothing was found. So I suspect the micropython module is undocumented on the micro:bit (and without help(modules) there is no way for a user to discover it exists, so it is unicorn dust (magic!))

screen shot 2018-10-30 at 12 16 54

Shall I log a different ticket for that, would that help to keep this PR focused?

DavidWhaleMEF:
I think it would be bad to link to the core micropython readthedocs, as that page includes other methods that are not supported on the micro:bit, and will confuse our users.

http://docs.micropython.org/en/latest/library/micropython.html?highlight=python

screen shot 2018-10-30 at 12 20 43

carlosperate:
Thanks David!

The upstream MicroPython docs also refer to it as a bus, so I'd be inclined to leave it, but it had never occurred to me that it could be an incorrect use of the term "bus".

The `micropython` module and `kbd_intr` function are documented in this PR still under work: https://github.com/bbcmicrobit/micropython/pull/578
When they get merged, sphinx will add a link to that specific section to the docs, as the restructured text formatting is already in place here.

DavidWhaleMEF:
Regarding... 
 Read characters.  If ``nbytes`` is specified then read at most that many
    bytes, otherwise read as many bytes as possible.

1. If it returns a bytes object, then it is reading bytes, not characters.

2. Returning 'as many bytes as possible' - is this capped at the 64 byte internal buffer?

3. What happens if characters are received at the UART interface but the buffer is full? Are bytes dropped (silently) and/or is there any way to query the uart object to sense that a buffer overflow has occurred? Just mentioning this as I am implementing binary payload transfer at the moment in a project, and this uart object makes error handling in buffer overflow conditions very hard!

Thanks

microbit-carlos:
Thanks for the feedback!

> If it returns a bytes object, then it is reading bytes, not characters.

I guess technically a byte and an (ASCII) character are interchangeable in this context, but could lead to confusion. The last commit has edited the working a bit, does that make it more clear?

> Returning 'as many bytes as possible' - is this capped at the 64 byte internal buffer?

I would think so, but I am not sure. Perhaps @dpgeorge can confirm.

> What happens if characters are received at the UART interface but the buffer is full? Are bytes dropped (silently) and/or is there any way to query the uart object to sense that a buffer overflow has occurred? Just mentioning this as I am implementing binary payload transfer at the moment in a project, and this uart object makes error handling in buffer overflow conditions very hard!

There is no way to detect the data has been dropped, but we do have this open feature request: https://github.com/bbcmicrobit/micropython/issues/504

DavidWhaleMEF:
Yes, that wording is clearer and more correct, thanks!

dpgeorge:
> > Returning 'as many bytes as possible' - is this capped at the 64 byte internal buffer?
> 
> I would think so, but I am not sure. Perhaps @dpgeorge can confirm.
> 

It's possible to get more than 64 bytes if the data is coming in at the same time as reading.  Eg, buffer is nearly full, `uart.read()` starts draining it and copying it to the buffer to return, and while it is draining more bytes are added, which will then be drained into the buffer.

DavidWhaleMEF:
@dpgeorge Interesting. So, does this mean that if data is coming in faster than MicroPython can process it and you ask for 'all data' (no limit), it will keep reading and filling a user buffer until memory is exhausted, and then the user script will crash?

dpgeorge:
> does this mean that if data is coming in faster than MicroPython can process it

If that's the case then you need to redesign something in your protocol/device :)

> and you ask for 'all data' (no limit)

For UART protocols it's generally not a good idea to ask for "all data" because there is no well defined end to the incoming data.

>  it will keep reading and filling a user buffer until memory is exhausted, and then the user script will crash?

Yes.  It will keep reading and then raise a MemoryError, so the script could catch this and possibly recover.

DavidWhaleMEF:
Sadly the device I am connecting to has it's own defined protocol with long return lines, sometimes longer than 64 characters. Hence another request to have a flag or something that indicates the buffer has overflowed and data has been lost.

Thanks for info. So, reading 'all available data' is not really a useful feature in this API then? Should it be removed?

DavidWhaleMEF:
Ok, good point. With a slower baud-rate it might still work anyway, so worth hanging on to.

LGTM for a merge.

Issue #592, opened 2018-10-30
Lack of a method deprecation process creates poor user experience


DavidWhaleMEF:
Regarding this example: https://github.com/bbcmicrobit/micropython/issues/591

I think we should have a simple but controlled process for managing breaking changes to the language/API.

DavidWhaleMEF:
Comments welcomed @carlosperate @jaustin

Removing methods will break existing users code. We all know that, but its worth remembering that, as it provides a bad experience for users who expect their code to always work on the latest editors and languages.

Should we mark removed methods in the docs as 'deprecated, please use x instead', and have a deprecation process a bit like Java, where one major release marks it as deprecated, the next major release actually removes it?

And, any method that once existed that is now deprecated and removed should be listed at the base of the readthedocs page for that module, so that users that miss the deprecation steps can still work out themselves what has gone wrong (and don't think 'how did this code ever work, like I just did??!!)

microbit-carlos:
> Should we mark removed methods in the docs as 'deprecated, please use x instead', and have a deprecation process a bit like Java, where one major release marks it as deprecated, the next major release actually removes it?

Yes, I think that would be good. And since a deprecated method would be a breaking change it makes sense to trigger a major version (`readall` was deprecated from v0.9 to v1.0).

Because `readall` was deprecated before we had a good process I think we should keep the deprecation notice in the docs for v1.0 for a while, but for how long? The default answer would be to only remove it on the next major release (v2.0), but our aim is to never do a breaking change again, and so we are not planning to do a major release (all new features would trigger a a new minor version).

Issue #591, opened 2018-10-30
readall method does not exist


DavidWhaleMEF:
The currently published readthedocs for the uart module refer to a 'readall' method.

https://microbit-micropython.readthedocs.io/en/latest/uart.html

![screen shot 2018-10-30 at 12 55 17](https://user-images.githubusercontent.com/25079586/47719431-164f1080-dc43-11e8-92c3-9cb6d79f8a38.png)

The current v1.0.0 MicroPython image does not have this method:

```
MicroPython v1.9.2-34-gd64154c73 on 2017-09-01; micro:bit v1.0.0 with nRF51822
Type "help()" for more information.
>>> dir(uart)
['init', 'any', 'read', 'readline', 'readinto', 'write', 'ODD', 'EVEN']
```

DavidWhaleMEF:
Note, it looks like this functionality can be achieved with readinto(), the main difference being that you provide the buffer, rather than having the buffer created for you (as the old readall did).

I recall reading that @dpgeorge had removed this a while ago, but it's still in the live docs, and the live docs don't match the live language (I appreciate we're all working to correct that at the moment though!)

DavidWhaleMEF:
Note: Removing methods will break existing users code. We all know that, but its worth remembering that, as it provides a bad experience for users who expect their code to always work on the latest editors and languages.

Should we mark removed methods in the docs as 'deprecated, please use x instead', and have a deprecation process a bit like Java, where one major release marks it as deprecated, the next major release actually removes it?

Comments welcomed @carlosperate @jaustin

microbit-carlos:
Completely agree that we should have a good deprecation process, although the aim is to never have to use it.

The `readall` method was removed in https://github.com/bbcmicrobit/micropython/commit/d231463dd9a4c9fe366314c2a59d5ff507a22893, which was over 2 years ago. At this point I think we should add a note to the uart docs to indicate it has been deprecated. I've created this PR, let me know what you think:  https://github.com/bbcmicrobit/micropython/pull/595

As far as achieving the same functionality goes, you can use the `read()` method without any arguments to do the same. PR https://github.com/bbcmicrobit/micropython/pull/588 clarifies the method behaviour if an argument is not provided.

microbit-carlos:
Thanks David, the docs have been updated in PR #595.

PR #595, opened 2018-10-31
docs: Deprecate uart.readall()

carlosperate:
Based on the conversation from issue #591, this PR adds a deprecation notice to `uart.readall()`.

This is rendered as:
image

Had to add a custom css file to be able to strike the text.

@dpgeorge could you review this PR? Wondering what are your views on this deprecation in the docs.

Related discussion: https://github.com/bbcmicrobit/micropython/issues/592

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -352,10 +352,3 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
     html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
 
 # otherwise, readthedocs.org uses their theme by default, so no need to specify it
-
-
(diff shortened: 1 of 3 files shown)


dpgeorge:
Since this function is completely removed I don't think "deprecated" is the right wording.  It should be "removed".  I also don't think it's worth adding custom css and striking out the existing text.  I would suggest something like the following:
```
uart.readall()
Removed since version 1.0.
Instead, use uart.read() with no arguments, which will read as much data as possible.
```

microbit-carlos:
Sounds good. Updated in the latest commit: 
image

dpgeorge:
That looks good.

PR #606, opened 2018-11-28
lib: Add "bx lr" instruction to end of assembly neopixel function.

dpgeorge:
Luckily in the output firmware this function was followed by a simple
function that just returned 0, so that's why it worked without this return.


microbit-carlos:
Good catch, I don't have any neopixels at hand to test, but the fix looks good to me for merge.

dpgeorge:
Ok, merged in e2294590a314b90463a598d3deebd30843f2f536

PR #607, opened 2018-11-28
lib: Make sure that PWM continues to work even if it misses an IRQ.

### LINKED ISSUE #600: neopixel effects music
i've bought <a href="https://www.kitronik.co.uk/5626-game-zip-64-for-the-bbc-microbit.html">this joypad (kitronik game zip)</a>.
It has a little speaker and an array of "zip LEDs" (neopixel compatible addressable LEDs)
when i run the following program the music plays once at the start and then about every 5 loops randomly.
```
from microbit import *
import music
import neopixel

zip_led = neopixel.NeoPixel(pin0, 64)

while True:
    display.show('-')
    music.play(music.BA_DING, pin2, False)
    display.clear()

    sleep(1000 * 1)
    
    zip_led.clear()
    zip_led[1] = (20, 20, 0)
    zip_led.show()

    sleep(1000 * 2)
```

i've tried it on the <a href="https://python.microbit.org/v/1">'old' web editor</a> and the music plays, but there are issues with neopixel and all LEDs are lit bright white rather than 1 yellow LED.

dpgeorge:
If the soft PWM driver misses an IRQ then the next CC value (computed as CC[x] += delay) could already be behind the timer counter value, and hence the next compare IRQ won't fire until the timer counter value wraps around.

This needs some testing because the timing of the PWM will now be slightly different compared to before, since the next event will be computed from the *current counter value* plus delay, rather than the counter value at the *time of the previous IRQ*.  But I can't really think of any other way to fix the issue of missed IRQs.

Should fix #600

### DIFF
--- a/source/lib/ticker.c
+++ b/source/lib/ticker.c
@@ -37,6 +37,23 @@
 #define LowPriority_IRQn SWI4_IRQn
 #define LowPriority_IRQHandler SWI4_IRQHandler
 
+// Ticker is configured to TIMER_BITMODE_BITMODE_24Bit.
+#define TICKER_MASK 0xffffff
+
+// Schedule the next CC value, taking care of the case where an IRQ could have been
+// missed, so CC[x]+dt is already behind the timer counter value.  In such a case
+// schedule it as COUNTER[x]+dt.
+#define TICKER_SCHEDULE_NEXT_CC(reg_id, dt) \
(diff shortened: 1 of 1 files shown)


dpgeorge:
This fix was updated so that it retained precise PWM timing in the case when no IRQs were missed.

PR #618, opened 2018-12-11
module.json: Update to DAL release version 2.1.1.

microbit-carlos:
DAL changes can be seen here: https://github.com/lancaster-university/microbit-dal/compare/v2.1.0...v2.1.1

The main things relevant to MicroPython are:
- The delayed error for accelerometer/magnetometer (panic is only raised when the acc/mag methods are used for the first time, instead of during init)
- Radio added 2 new defines to configure the channel max (83) and min(0) frequencies
    - These are the values already used in MicroPython, and since it's not worth including a new header file for just these two defines, we don't need to do anything.
- Radio fixed an issue with the default transmit power not being set correctly (so in v2.1.0 it started very low)
    - The default power used in MicroPython is the same as the DAL, but it's defined internally, so MicroPython was not affected by this bug

We've tested all the features affected, and done a sanity check on the rest of the basic features.

Haven't updated the version in the `module.json` file since we should only edit that when we change the `mpconfigport.h` file.

### DIFF
--- a/module.json
+++ b/module.json
@@ -7,7 +7,7 @@
   "author": "Damien P George <damien.p.george@gmail.com>",
   "homepage": "http://micropython.org",
   "dependencies":{
-    "microbit-dal":"lancaster-university/microbit-dal#v2.1.0"
+    "microbit-dal":"lancaster-university/microbit-dal#v2.1.1"
   },
   "extraIncludes":[
     "inc",


dpgeorge:
I tested this and didn't find any problems with it.  I think the only thing that would affect MicroPython is the delayed errors on accel/compass use (which I couldn't test).

Also the firmware size increased by 56 bytes from 228804 to 228860, but this didn't cross a 1024 page boundary so the filesystem should still be the same size.

So I think it's good to merge!

microbit-carlos:
Great, thanks Damien!

PR #630, opened 2019-01-04
microbit: Add get_analog_period_microseconds method to more pin classes.

microbit-carlos:
I noticed the `get_analog_period_microseconds` method was not present on all the micro:bit pins, turns out it was only added to the `MicroBitDigitalPin` class definition, so this PR adds the method to `MicroBitAnalogDigitalPin` and `MicroBitTouchPin` as well.


dpgeorge:
Looks good, although I didn't test it.

dpgeorge:
Tested and merged in aa42114b279dfa7f3f0775dda83b43a98553e701

PR #634, opened 2019-01-19
Use full yotta command (not yt) during build

knowledgejunkie:
yotta is now packaged for Debian testing/unstable, and should be available to install via apt in the forthcoming stable releases of Debian 10 "Buster" and Ubuntu 19.04 "Disco Dingo".

When yotta is installed via the apt package manager on Debian/Ubuntu, the `yt` shorthand alternative for the yotta command is unavailable due to a clash with an identically-named command from the python3-yt package.

This update ensures the default yotta command is used when building MicroPython via the Makefile, and updates the README to use the full yotta command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ $(HEX_FINAL): yotta $(VER_ADDR_FILE)
 	@size $(HEX_SRC:.hex=)
 
 yotta: $(MBIT_VER_FILE)
-	@yt build
+	@yotta build
 
 $(MBIT_VER_FILE): FORCE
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution, it looks good to me.  I tested it on Arch Linux and it works fine.

dpgeorge:
Merged in 3923ae7de9e905bcfa0e923a3495b89b2e309978 and 76cc06158fdbf913569972f0bc3bcbe85c4c6378, with the latter commit updated to also change instances of "yt" in the docs.

PR #645, opened 2019-03-19
Docs : Delete python maillist info

selmen2004:
this is because it was deleted from python website
it returns error : No such list microbit


dpgeorge:
Thanks for the contribution, merged in a16d82fab1b92026f5b9b3b4b680bec20a9fed2e with a comment in the commit about the archive of this mailing list.

PR #648, opened 2019-03-26
Corrected comments so instructions  are for microbit

rhubarbdog:
The instructions in the comments were for 'The PyBoard' (pyb).
I've changed them so they use microbit code

### DIFF
--- a/tools/pyboard.py
+++ b/tools/pyboard.py
@@ -18,7 +18,8 @@ Or:
 Then:
 
     pyb.enter_raw_repl()
-    pyb.exec('pyb.LED(1).on()')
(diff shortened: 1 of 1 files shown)


rhubarbdog:
This fixes <a href="https://github.com/bbcmicrobit/micropython/issues/649">#649</a>

dpgeorge:
Thanks for the patch.  Even though the file is called "pyboard.py" I agree it'd be nice to include a working microbit example here.  How about also changing the first paragraph so it says something like "This module provides the Pyboard class, used to communicate with and control a micro:bit or pyboard over a serial USB connection."?

mytechnotalent:

PR #653, opened 2019-03-31
Incorrect SI unit?

objarni:
The "Technical Details" section says 7800 kHz samples are consumed per second, but the overview states 7812.5 samples per second.

Either it is 7812.5 Hz, or 7.8125 kHz, I presume?


dpgeorge:
You are right, thanks for the patch!  Merged in 7383c2cd6db7940b23b447b19bcdfcaaeaf393fe

Issue #657, opened 2019-04-04
Makefile 'create version file' target commands need disambiguating


WayneKeenan:
When running `Makefile` in my environment (macOS) the `$(VER_ADDR_FILE): yotta` target fails for 2 reasons:

1. The shell `echo` built-in doesn't support the `-n` flag so a newline is not generated resulting in a badly formatted `build/veraddr.txt`, the `/bin/echo` command needs calling.
2. the 'vanilla' `objdump` command picks up the XCode utility and not the arm-gcc utility in the PATH.

I have a PR ready to be raised.

The error output:
```
[2/2] Linking CXX executable source/microbit-micropython
objdump: Unknown command line argument '-x'.  Try: '/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/objdump -help'
objdump: Did you mean '-D'?
tools/adduicr.py build/bbc-microbit-classic-gcc-nosd/source/microbit-micropython.hex $(cat build/veraddr.txt) -o build/firmware.hex
usage: adduicr.py [-h] [-o OUTPUT] firmware address
adduicr.py: error: argument address: invalid <lambda> value: '0x'
make: *** [build/firmware.hex] Error 2

```

dpgeorge:
Fixed by dc92faf91be4cdf74130ec9144a5e5ba02d3836d

Issue #656, opened 2019-04-04
Firmware build fails on MacOS at make echo command


diaena:
`make` generates an incorrect `VER_ADDR_FILE`: 

<img width="719" alt="image" src="https://user-images.githubusercontent.com/10000430/55552626-cf73b400-56d5-11e9-95c4-2d7e28868b20.png">

<img width="303" alt="image" src="https://user-images.githubusercontent.com/10000430/55552585-aa7f4100-56d5-11e9-9695-b0c47c2ab7e1.png">

(It should be of the form `0x00036df2`)

This is because make uses `sh` to execute commands by default, but MacOS and Linux implement `sh` differently.
On MacOS, `sh` functionality is implemented by bash which on MacOS is compiled with `--enable-strict-posix-default` and as a result of this flag is that the `echo` command does not understand the `-n` flag.

Ways of fixing:

1. use make inbuilt echo command
2. use printf constructs instead of echo (probably nicest and most portable)
3. use the full path to `/bin/echo`

WayneKeenan:
I just raised issue #657 and PR #658, which also fixes an `objdump` command issue,

Frederic-soft:
There is also an issue with ```size```, which should be ```arm-none-eabi-size``` on line 16 of the Makefile.

dpgeorge:
Fixed by dc92faf91be4cdf74130ec9144a5e5ba02d3836d and 4cd8ad10529eeec9b58b9d9ad9f39a2b177a1a35

PR #658, opened 2019-04-04
Make shell command explicit for version file creation

WayneKeenan:
Fix for #657

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,8 @@ $(MBIT_VER_FILE): FORCE
 	python tools/makeversionhdr.py $(MBIT_VER_FILE)
 
 $(VER_ADDR_FILE): yotta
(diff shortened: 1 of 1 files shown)


WayneKeenan:
also applies for #656

dpgeorge:
Thanks for the PR.  I think we can eliminate completely the need for the "echo -n" by calling adduicr.py as follows:
```
tools/adduicr.py $(HEX_SRC) 0x$$(cat $(VER_ADDR_FILE)) -o $(HEX_FINAL)
```
Note the addition of "0x" before cat'ing VER_ADDR_FILE.  I'd prefer this approach because it's simpler, can you please update the PR?

dpgeorge:
Updated as above and merged in dc92faf91be4cdf74130ec9144a5e5ba02d3836d

PR #659, opened 2019-04-08
docs: fix the table in pin.rst to match the diagram

mattheww:
The pin table in `pins.rst` disagrees with the diagram:

 - it says three of the edge connector pins are for the LED rows, rather than columns 7, 8, and 9;
 - it has the three SPI pins in the opposite order

I think the diagram is right and the table is wrong:

 - [pinmap.csv] says the three LED pins are internal GPIOs 10,11,12 which are definitely columns;
 - spi.rst has sclk=pin13, mosi=pin15, miso=pin14 .

[pinmap.csv]: https://tech.microbit.org/docs/hardware/pinmap.csv

### DIFF
--- a/docs/pin.rst
+++ b/docs/pin.rst
@@ -43,13 +43,13 @@ module:``microbit.pin0`` - ``microbit.pin20``.
 +-----+---------+----------+
 |  5  | Digital | Button A |
 +-----+---------+----------+
-|  6  | Digital | Row 2    |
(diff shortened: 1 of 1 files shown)


dpgeorge:
Thanks for the contribution, you are correct!

The fix here also matches the code, source/microbit/microbitspi.cpp, and the display pin mappings in source/microbit/microbitdisplay.cpp

PR #660, opened 2019-04-23
Update uart.rst

fcorthay:
provided info about what the serial port does and doesn't


dpgeorge:
Thanks for the contribution.

But the text added here is a bit confusing and out of place.  A user wanting to use the UART should already know what they want it for, so no need to explicitly mention it.  And there is already a note/warning about losing the USB REPL when using the UART, just below the `init()` method.

mytechnotalent:

PR #632, opened 2019-01-07
Docs : corrected and improved compile information

rhubarbdog:
This is corrected information which became apparent during issue <a href="https://github.com/bbcmicrobit/micropython/issues/619">#619</a>

Issue #619, opened 2018-12-12
Can't install compilation packages - ubuntu 16.04


rhubarbdog:
Hi,
I want to be able to compile my own firmware, but i am using a ubuntu 16.04 clone. I've been following the discussion <a href="https://github.com/bbcmicrobit/micropython/issues/604">#604</a> but it seem's that the OP has dropped out of the discussion.
I follow the reasoning that `ppa/ppamiller` is no longer required and procede to follow the instructions in README.md.
when i type `sudo apt-get install cmake ninja-build gcc-arm-none-eabi srecord libssl-dev` it terminates with the error
```
Errors were encountered while processing:
 /var/cache/apt/archives/binutils-arm-none-eabi_2.26-4ubuntu1+8_amd64.deb
 /var/cache/apt/archives/gcc-arm-none-eabi_15%3a4.9.3+svn231177-1_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

carlosperate:
Could you try to run this and see if it works on its own?

```
sudo add-apt-repository -y ppa:team-gcc-arm-embedded/ppa
sudo apt-get update -qq
sudo apt-get install -y gcc-arm-embedded
```

rhubarbdog:
> Could you try to run this and see if it works on its own?
>
> ```
> sudo add-apt-repository -y ppa:team-gcc-arm-embedded/ppa
> sudo apt-get update -qq
> sudo apt-get install -y gcc-arm-embedded
> ```

all worked with no errors to reports

when i try `sudo apt-get install cmake ninja-build gcc-arm-none-eabi srecord libssl-dev` it reports

```
Errors were encountered while processing:
 /var/cache/apt/archives/binutils-arm-none-eabi_2.26-4ubuntu1+8_amd64.deb
 /var/cache/apt/archives/gcc-arm-none-eabi_15%3a4.9.3+svn231177-1_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

carlosperate:
Hi @rhubarbdog,

Doing `apt-get install gcc-arm-embedded` should get you the GCC ARM compiler already, so there is no need to include `gcc-arm-none-eabi` in the next command. Can you try instead:

```
sudo apt-get install cmake ninja-build srecord libssl-dev
```

So, in summary, it would be:

```
sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa
sudo apt-get update
[...]

PR #667, opened 2019-09-04
Updated instructions for Ubuntu 18.04.

PropGit:
Existing build system instructions do not work for a fresh install of Ubuntu 18.04.

Found Issue #619 to be helpful but still stumbled along the way.  PR #632 is the closest to what I came up with after many attempts and scouring the web (and it should still be considered for 16.04.

This PR includes instruction updates created and tested multiple times on fresh Ubuntu 18.04 VM systems.


mytechnotalent:

Issue #655, opened 2019-04-01
Missing modules in the docs


microbit-carlos:
The following MicroPython modules are not referenced in the docs:

* array
    - [Upstream docs](https://docs.micropython.org/en/latest/library/array.html)
* builtins
    - [Upstream docs](https://docs.micropython.org/en/latest/library/builtins.html)
* ucollections
    - [Upstream docs](https://docs.micropython.org/en/latest/library/ucollections.html)
* gc
    - [Upstream docs](https://docs.micropython.org/en/latest/library/gc.html)
* math
    - [Upstream docs](https://docs.micropython.org/en/latest/library/math.html)
* ustruct
    - [Upstream docs](https://docs.micropython.org/en/latest/library/ustruct.html)
    - Related issue: https://github.com/bbcmicrobit/micropython/issues/652
* sys
    - [Upstream docs](https://docs.micropython.org/en/latest/library/sys.html)

We should either port the upstream docs (note that the module might only be partially enabled in the micro:bit port) or add an entry and link to the upstream docs.
Taking in consideration we have already ported `utime`, `micropython`, and `machine`, we should probably do the same for these.

mofada:
Running `help('modules')` in REPL mode shows that there is a utime module, but I cannot find it in the editor, what should I do?

![Y1Y$ `D%EXM37KB }_1~AR](https://user-images.githubusercontent.com/22286246/97841143-3331e900-1d20-11eb-9bd6-e807c3177c54.png)

![~BNG8K9Y(UJY%88F_4G _61](https://user-images.githubusercontent.com/22286246/97841202-4cd33080-1d20-11eb-82a0-01ea65b6e71a.png)

mofada:
Oh, sorry, I think my problem is a plug-in problem

kopp:
I guess, that you are using this package https://pypi.org/project/pseudo-microbit/ which does not provide `utime`. Maybe you can open an issue in this project?

mofada:
@kopp Yes, I am using this plugin, but it seems that the most recent update was three years ago and I am not sure if it is still being maintained.

mofada:
[intellij-micropython](https://github.com/vlasovskikh/intellij-micropython)  Or this plug-in is also very good, but it also lacks those few files. I filed an issue and I don’t know when it will be updated.

PR #670, opened 2019-10-03
Add missing modules in the docs #655

aitzol:
Added raw content from https://docs.micropython.org


mytechnotalent:

PR #671, opened 2019-11-28
Docs: remove reference to windows driver

microbit-mark:
as per Foundation ticket microbit-foundation/platform-software-issue-tracker#582 Windows 7 reaches EOL in Jan 2020 and we no longer want to recommend  installing the mBED driver as it conflicts with later versions of  Windows


mytechnotalent:

microbit-carlos (approved):
Good call, this was only really useful for Windows 7 (and technically 8.0), and installing it in newer Windows versions can break WebUSB, so it's much better to remove this.
Thanks @microbit-mark!

PR #691, opened 2020-11-08
Support RGBW WS2812 neopixels

ssav7912:
This patch adds support for RGB-W WS2812 neopixel components. 

NeoPixel constructor now takes 3 arguments (pin, num_leds, num_channels). num_channels is used to determine whether an extra white channel needs to be sent to the LED chip. 

neopixelsend.s assembly also now takes num_channels as a parameter in R3 to use as "parallel - serial conversion mask" in R6 instead of hardcoding value.

Commenting was largely for my own benefit deciphering the micropython internal structure and so may be untidy/uncertain.

### DIFF
--- a/source/lib/neopixelsend.s
+++ b/source/lib/neopixelsend.s
@@ -27,7 +27,7 @@
 
 .global sendNeopixelBuffer
 
- /* declared as extern void sendBuffer(uint32_t pin, uint8_t* data_address, uint16_t num_leds) */
+ /* declared as extern void sendBuffer(uint32_t pin, uint8_t* data_address, uint16_t num_leds, uint16_t num_channels) */
 
 sendNeopixelBuffer:
     push {r0, r1, r2, r3, r4, r5, r6}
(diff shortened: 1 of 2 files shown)


dpgeorge:
Thanks for the contribution.  This is a good feature to have.

I've actually implemented this feature (RGBW pixel support) in 8d3078846c0129c2a1db20bd115b1501b5c07cd8 and 652116dd094d241807b0e16b8b51847c5eb469fa which:
- retains backwards compatibility with the existing RGB API
- is compatible with micro:bit v2 (which already supports RGBW)
- is a bit more concise on code size (which needs to be minimised on this project)

source/lib/neopixelsend.s
```diff
@@ -27,7 +27,7 @@
 
 .global sendNeopixelBuffer
 
- /* declared as extern void sendBuffer(uint32_t pin, uint8_t* data_address, uint16_t num_leds) */
+ /* declared as extern void sendBuffer(uint32_t pin, uint8_t* data_address, uint16_t num_leds, uint16_t num_channels) */
```
dpgeorge, line 30: can just change the `num_leds` parameter to be `num_bytes`, then the multiplication of channels*LEDs can be done outside this routine

source/microbit/modneopixel.cpp
```diff
@@ -34,45 +34,57 @@ extern "C" {
...
 
 static inline void neopixel_clear_data(neopixel_obj_t *self) {
-    memset(&self->data[0], 0, COL_NUM_COMPONENTS * self->num_pixels);
+    memset(&self->data[0], 0, self->num_channels * self->num_pixels);
 }
 
 STATIC mp_obj_t neopixel_make_new(const mp_obj_type_t *type_in, mp_uint_t n_args, mp_uint_t n_kw, const mp_obj_t *args) {
     (void)type_in;
-    mp_arg_check_num(n_args, n_kw, 2, 2, false);
+    mp_arg_check_num(n_args, n_kw, 3, 3, false);
```
dpgeorge, line 67: this change would break existing use of this class, because it now requires the `num_channels` arg

PR #694, opened 2020-12-15
Docs: update for RGBW neopixel

microbit-mark:
Doc update to include support for RGBW Neopixels. I haven't tested this on hardware, so making an assumption that the values are in the order written eg `np[0]= (R, G, B, W)` as opposed to GRBW which I think is the default for NeoPixelsand we order them here https://github.com/microbit-foundation/micropython-microbit-v2/blob/6a4b97ffce97a41a593e77553a82e24e59b7107c/src/codal_port/modules/neopixel.py#L19

### DIFF
--- a/docs/neopixel.rst
+++ b/docs/neopixel.rst
@@ -14,7 +14,7 @@ need to import it separately with::
     From our tests, the micro:bit NeoPixel module can drive up to around 256
     NeoPixels. Anything above that and you may experience weird bugs and
     issues. The micro:bit can only supply 90mA **V1** or 190mA **V2**  to
-    external devices,larger numbers of NeoPixels require an external power
+    external devices, larger numbers of NeoPixels require an external power
     supply with common ground.
 
     NeoPixels are designed to work at 5V, but luckily they still function using
--- a/docs/neopixel.rst
+++ b/docs/neopixel.rst
@@ -11,10 +11,10 @@ need to import it separately with::
 
 .. note::
 
-    From our tests, the micro:bit NeoPixel module can drive up to around 256
+    From our tests, the Microbit NeoPixel module can drive up to around 256
     NeoPixels. Anything above that and you may experience weird bugs and
     issues. The micro:bit can only supply 90mA **V1** or 190mA **V2**  to
-    external devices, larger numbers of NeoPixels require an external power
+    external devices,larger numbers of NeoPixels require an external power
     supply with common ground.
 
     NeoPixels are designed to work at 5V, but luckily they still function using
@@ -54,9 +54,10 @@ Classes
     NeoPixel(pin, n, bpp)
 
     Initialise a new strip of ``n`` number of neopixel LEDs controlled via pin
-    ``pin``. For **V2** micro:bit boards that support RGBW neopixels, a third
-    value can be passed to ``NeoPixel`` to indicate the number of bytes per
-    pixel (bpp) is 4 rather than the default of 3.
+    ``pin``. The **V2** micro:bit can also support RGBW neopixels, so a third
+    argument can be passed to ``NeoPixel`` to indicate the number of bytes per
+    pixel (bpp). For RGBW, this is is ``4`` rather than the default of 3 for
+    RGB and GRB.
     
     Each pixel is addressed by a position (starting from 0). Neopixels
     are given RGB (red, green, blue) / RGBW (red, green, blue, white) **V2**
--- a/docs/neopixel.rst
+++ b/docs/neopixel.rst
@@ -4,7 +4,7 @@ NeoPixel
 .. py:module:: neopixel
 
 The ``neopixel`` module lets you use NeoPixel (WS2812) individually addressable
-RGB LED strips with the micro:bit. Note to use the ``neopixel`` module, you
+RGB and RGBW **V2** LED strips with the micro:bit. Note to use the ``neopixel`` module, you
 need to import it separately with::
 
     import neopixel
@@ -13,9 +13,9 @@ need to import it separately with::
 
     From our tests, the Microbit NeoPixel module can drive up to around 256
     NeoPixels. Anything above that and you may experience weird bugs and
-    issues. As the micro:bit can only supply 90mA to external devices,
-    larger numbers of NeoPixels require an external power supply with common
-    ground.
+    issues. The micro:bit can only supply 90mA **V1** or 190mA **V2**  to
+    external devices,larger numbers of NeoPixels require an external power
+    supply with common ground.
 
     NeoPixels are designed to work at 5V, but luckily they still function using
     the 3V supply of the BBC micro:bit. Please note that the micro:bit edge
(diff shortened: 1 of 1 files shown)


docs/neopixel.rst
```diff
@@ -51,11 +51,17 @@ Classes
 
 .. py:class::
     NeoPixel(pin, n)
+    NeoPixel(pin, n, 4)
 
     Initialise a new strip of ``n`` number of neopixel LEDs controlled via pin
-    ``pin``. Each pixel is addressed by a position (starting from 0). Neopixels
-    are given RGB (red, green, blue) values between 0-255 as a tuple. For
-    example, ``(255,255,255)`` is white.
+    ``pin``. For **V2** micro:bit boards that support RGBW neopixels, a third
```
microbit-carlos, line 57: "For **V2** micro:bit boards that support RGBW neopixels" sounds like some V2 board do not support them?

docs/neopixel.rst
```diff
@@ -51,11 +51,17 @@ Classes
...
+    value can be passed to ``NeoPixel`` to indicate the number of bytes per
```
microbit-carlos, line 58: A third "argument"

docs/neopixel.rst
```diff
@@ -51,11 +51,17 @@ Classes
...
+    pixel is 4 rather than the default of 3.
```
microbit-carlos, line 59: I think we can select an unspecified number of "pixels"? So while the main two options are 3 or 4 values, technically more could be sent if a new class of neopixel was released in the future.

I probably wouldn't mention it like that here as that could be confusing, but something along the lines of " number of bytes per pixel, where the default value of 3 is used for RGB and GRB, and 4 can be used for RGBW".

docs/neopixel.rst
```diff
@@ -51,11 +51,17 @@ Classes
 
 .. py:class::
     NeoPixel(pin, n)
+    NeoPixel(pin, n, 4)
```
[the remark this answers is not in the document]
microbit-mark: it is 3, I believe. perhaps `bpp` is better as long as the acronym is explained.

docs/neopixel.rst
```diff
@@ -13,9 +13,9 @@ need to import it separately with::
 
     From our tests, the Microbit NeoPixel module can drive up to around 256
     NeoPixels. Anything above that and you may experience weird bugs and
-    issues. As the micro:bit can only supply 90mA to external devices,
-    larger numbers of NeoPixels require an external power supply with common
-    ground.
+    issues. The micro:bit can only supply 90mA **V1** or 190mA **V2**  to
+    external devices,larger numbers of NeoPixels require an external power
+    supply with common ground.
```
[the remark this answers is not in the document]
microbit-mark: The code content in the learning system is MIT licensed https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/master/LICENSE but the image assets are not included here, so I have contacted adafruit to see if we are able to use the material. For now, I have added attribution https://github.com/bbcmicrobit/micropython/pull/694/commits/7599420f7dedb1d42e8187a2132d95b106775602

docs/neopixel.rst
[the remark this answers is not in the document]
microbit-mark: https://github.com/bbcmicrobit/micropython/pull/694/commits/40a0c50bc2569fa34670796e1b8dd200b37aab1c

docs/neopixel.rst
microbit-mark: Updated in https://github.com/bbcmicrobit/micropython/pull/694/commits/474f3cc7442f540d016ee946f14475666bcb78a6

docs/neopixel.rst
microbit-mark: Done in https://github.com/bbcmicrobit/micropython/pull/694/commits/474f3cc7442f540d016ee946f14475666bcb78a6

docs/neopixel.rst
microbit-mark: UPdated in https://github.com/bbcmicrobit/micropython/pull/694/commits/474f3cc7442f540d016ee946f14475666bcb78a6

docs/neopixel.rst
microbit-mark: UPdated in https://github.com/bbcmicrobit/micropython/pull/694/commits/474f3cc7442f540d016ee946f14475666bcb78a6

docs/neopixel.rst
```diff
@@ -51,12 +53,13 @@ Classes
...
-    NeoPixel(pin, n, 4)
+    NeoPixel(pin, n, bpp)
 
     Initialise a new strip of ``n`` number of neopixel LEDs controlled via pin
-    ``pin``. For **V2** micro:bit boards that support RGBW neopixels, a third
-    value can be passed to ``NeoPixel`` to indicate the number of bytes per
-    pixel is 4 rather than the default of 3.
+    ``pin``. The **V2** micro:bit can also support RGBW neopixels, so a third
+    argument can be passed to ``NeoPixel`` to indicate the number of bytes per
+    pixel (bpp). For RGBW, this is is ``4`` rather than the default of 3 for
```
microbit-carlos, line 61:
```suggestion
    pixel (bpp). For RGBW, this is is ``4`` rather than the default of ``3`` for
```

microbit-carlos (commented):
V2 also has `fill` and `clear`, could you check if those are available for V1 as well?
https://github.com/microbit-foundation/micropython-microbit-v2/blob/6a4b97ffce97a41a593e77553a82e24e59b7107c/src/codal_port/modules/neopixel.py

docs/neopixel.rst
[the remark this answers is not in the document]
microbit-mark: adafruit have confirmed we can use the image as we wish

microbit-mark:
V1 has `clear` and `show`. V2 has `clear`,` write`, `fill`, `show`

microbit-mark:
I've updated to add the V2 methods in https://github.com/bbcmicrobit/micropython/pull/694/commits/964185346ea4ca37afb5a33bc5e49444f82257a3

Not sure on how best to present `write()` as I think it is equivalent to `show()`.

Also not sure on how to present arguments for `fill()`. Is `fill((r,g,b))` okay or should it be `n,n,n` or better to show the format in the description?

docs/neopixel.rst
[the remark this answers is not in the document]
microbit-carlos: Great, thanks Mark!

microbit-carlos:
> Not sure on how best to present write() as I think it is equivalent to show().

Yeah, they are equivalent. The current text is good, although if you need to edit the file anyway I might remove `is available as part of the RGBW implementation and`, just mentioning that they are equivalent is good enough.

> Also not sure on how to present arguments for fill(). Is fill((r,g,b)) okay or should it be n,n,n or better to show the format in the description?

Showing the tuple would be difficult, as the length depends on how many bytes per pixel. I think it's fine to just name the argument `color`, as done in the [python source code](https://github.com/microbit-foundation/micropython-microbit-v2/blob/6a4b97ffce97a41a593e77553a82e24e59b7107c/src/codal_port/modules/neopixel.py), and then in the description mention that it has to be a tuple of the same length as the number of colours per pixel, as configured in the constructor.

microbit-carlos:
Perfect, thanks Mark!
Ready for merge 🎉

PR #696, opened 2020-12-21
V2 docs: Remove is_on() from speaker.

microbit-carlos:


microbit-carlos:
Render looks good: https://microbit-micropython--696.org.readthedocs.build/en/696/speaker.html
We can merge.

PR #706, opened 2021-02-08
Update README to point to target repositories

jaustin:
As the yotta registry is being deprecated, this commit explicitly specifies that the yotta target for MicroPython should be fetched from GitHub rather than from the registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,10 +27,11 @@ Once all packages are installed, use yotta and the provided Makefile to build.
 You might need need an Arm Mbed account to complete some of the yotta commands,
 if so, you could be prompted to create one as a part of the process.
 
-- Use target bbc-microbit-classic-gcc-nosd:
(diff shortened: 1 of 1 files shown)


rneacy:
This builds as intended with ``make all`` on Linux (Ubuntu 20.04.1).

I'm not able to build micropython at all on Windows regardless of the target link due to some strange C++ versioning errors - but this may be on my end.

jaustin:
Thanks @rneacy - @dpgeorge are you happy to take this one for now to get the instructions in before the deprecation?

carlosperate:
@jaustin since we are updating the README could we also add a note about the V2 repo being a different place?

dpgeorge:
Merged in e9dc17f37398c52ed7c4fa5922a8ebec2e1d378c

PR #715, opened 2021-03-30
update pin_speaker context

microbit-mark:
Let people know that `pin_speaker` is designed only for PWM use.

There is also a reference to pin_speaker in https://github.com/bbcmicrobit/micropython/blob/v2-docs/docs/microbit_micropython_api.rst#pins but I think this is still valid


microbit-carlos (approved):
Thanks Mark!

PR #723, opened 2021-08-18
Docs: Update info on python.microbit.org

microbit-mark:
Attempts to ease some of the confusion we have on support tickets where people are trying to open MicroPython projects in MakeCode.

- Change to make the Python Editor the default way of programming
- Mention that the code won't work in MakeCode

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -25,20 +25,21 @@ you enjoy developing for the BBC micro:bit using MicroPython.
     **V2**.
 
 If you're a new programmer, teacher or unsure where to start, begin with the
-tutorials.
-
-.. image:: comic.png
+:ref:`Tutorials <Tutorials>` and use the `micro:bit Python Editor <https://python.microbit.org>`_
+to program the micro:bit. 
 
 .. note::
 
-    This project is under active development. Please help other
-    developers by adding tips, how-tos, and Q&A to this document.
-    Thanks!
+   MicroPython will not work in the MakeCode editor, as this
+   uses a `different version of Python. <https://support.microbit.org/support/solutions/articles/19000111744-makecode-python-and-micropython>`_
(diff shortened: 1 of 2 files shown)


docs/tutorials/introduction.rst
```diff
@@ -1,18 +1,17 @@
 Introduction
 ------------
+.. _Tutorials:
 
-We suggest you download and use the `mu editor <http://codewith.mu/>`_ when
-working through these tutorials. Instructions for downloading and installing
-Mu are on its website. You may need to install a driver, depending on your
-platform (instruction are on the website).
+We suggest you use the `micro:bit Python Editor <https://python.microbit.org>`_ when
+working through these tutorials.
```
microbit-carlos, line 6: No need to remove Mu from this tutorial, we can recommend both.

docs/index.rst
```diff
@@ -25,20 +25,21 @@ you enjoy developing for the BBC micro:bit using MicroPython.
...
-.. image:: comic.png
+:ref:`Tutorials <Tutorials>` and use the `micro:bit Python Editor <https://python.microbit.org>`_
+to program the micro:bit. 
 
 .. note::
 
-    This project is under active development. Please help other
-    developers by adding tips, how-tos, and Q&A to this document.
-    Thanks!
+   MicroPython will not work in the MakeCode editor, as this
```
microbit-carlos, line 33: Should we say the "MicroPython API will" to clarify it's the same language, but the modules/API is different?

microbit-carlos (commented):
Thanks for the clarifications @microbit-mark!
I think it might be best to cherry-pick this commit onto the master branch and then merge that here, as this is not V2 specific, it would also be beneficial to have this in the V1 docs.

PR #725, opened 2021-09-08
docs: fix missing pin methods and improve docs

microbit-matt-hillsdon:
- fix missing space in directive that caused methods to be omitted.
- recast the get_mode() return values as their Python values rather than
  C names which aren't defined in Python.
- remove duplicate class definition for MicroBitAnalogDigitalPin.

### DIFF
--- a/docs/pin.rst
+++ b/docs/pin.rst
@@ -144,28 +144,27 @@ its own to that.
 
         Set the pin to high if ``value`` is 1, or to low, if it is 0.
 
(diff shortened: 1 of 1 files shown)


microbit-matt-hillsdon:
get_mode return values taken from https://github.com/microbit-foundation/micropython-microbit-v2/blob/eba8995843ebc7246765b364710543c9ffee344a/src/codal_port/microbit_pinmode.c#L97

PR #727, opened 2021-09-23
Fix typo in neopixel docs: mumber -> number

microbit-matt-hillsdon:


microbit-carlos (approved):
Nice catch, thanks Matt!

PR #730, opened 2022-03-21
gitignore: Add Sphinx build dir, venv dirs, and macOS files.

microbit-carlos:


dpgeorge:
In general I say that only project-specific entries should go in `.gitignore`, like `docs/_build`.  Entries like `.DS_Store` and `venv` should go in the users `.gitignore_global` (and depend on their OS and how they like to set up virtual env dirs).

That said, this PR looks fine to me.

microbit-carlos:
Great, thanks Damien!

PR #731, opened 2022-03-25
docs: Remove radio.RATE_250KBIT as it has been dropped from V2.

microbit-carlos:
The constant is still present in the micro:bit V1 MicroPython port, but it is not in the V2 MicroPython port.
The value from the constant (`2`) can be used as an argument in the V2 MicroPython `radio.config(data_rate=2)`, but because it is not officially supported in the V2 nRF52833 microcontroller we cannot guarantee it will always work.

To keep the documentation simple it might be better to remove the mention of `radio.RATE_250KBIT`, rather than adding more content to a fairly long function description.

Discussion in https://github.com/microbit-foundation/micropython-microbit-v2/issues/87.

### DIFF
--- a/docs/radio.rst
+++ b/docs/radio.rst
@@ -28,17 +28,19 @@ We assume you have done this for the examples below.
 Constants
 =========
 
-.. py:attribute:: RATE_250KBIT
-
-    Constant used to indicate a throughput of 256 Kbit a second.
-
 .. py:attribute:: RATE_1MBIT
 
-    Constant used to indicate a throughput of 1 MBit a second.
+    Constant used to indicate a throughput of 1 Mbit a second.
 
 .. py:attribute:: RATE_2MBIT
 
-    Constant used to indicate a throughput of 2 MBit a second.
+    Constant used to indicate a throughput of 2 Mbit a second.
+
+.. py:attribute:: RATE_250KBIT
+
+    **Deprecated**.
+    This rate is possible with micro:bit V1, but it is not guaranteed to work
+    on V2, so it has been deprecated for compatibility reasons.
 
 
 Functions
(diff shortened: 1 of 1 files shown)


dpgeorge:
I agree, removing this const from the docs would be a good idea (so it's not used in new code).

But I also think there should be an addition to the docs explaining to some extent the situation with this constant and the 250kb data rate.  Eg "A datarate of 250kbit/sec is possible with micro:bit v1, and may be possible with micro:bit v2.  To access this hidden feature pass "2" to the `data_rate` config argument.  This was previously accessible via `RATE_250KBIT` but has been removed because it's not guaranteed to work on all v2 devices."

microbit-carlos:
Sounds good 👍 
I'll add it to the PR.

microbit-carlos:
As this PR will go into the main branch, which is for the V1 docs I think it might be better to separate this message for each target.

This PR contains the following note in the constants section:

```rst
.. note::

    A lower data rate of 250 kbit/sec is possible with micro:bit V1 via
    the ``RATE_250KBIT`` constant (or using its value ``2`` directly). However
    as this rate is not guaranteed to work on V2 it has been deprecated for
    compatibility reasons.
```

And this note below the `data_rate` argument in the V2 docs:

```rst
    .. note::

        A lower data rate of of 250kbit/sec is supported in micro:bit V1, and
        may be possible with micro:bit V2, but it is not guaranteed to work on
        all devices. To access this hidden feature for compatibility with V1
        pass ``2`` to the ``data_rate`` argument.
```

microbit-carlos:
@dpgeorge if the wording looks good we can merge this and create the v2-docs PR after rebasing that branch.

dpgeorge:
This looks fine to me.

microbit-carlos:
Thanks Damien, as discussed in the call, I've changed it back as a constant with a deprecation notice here and will create a different update specifically for the V2 docs.

image

dpgeorge:
> I've changed it back as a constant with a deprecation notice

Very good!

PR #733, opened 2022-04-01
docs: Update accelerometer page with better range info.

microbit-carlos:
Correct old notes about not being able to change the range.
Remove note already mentioned in the header, and move to header
note about needing to call an accelerometer function regulartly.

Indicate in the header the default range so that it can be removed
from each function description.

Use `:param:` and `:return:` in the function descriptions.

### DIFF
--- a/docs/accelerometer.rst
+++ b/docs/accelerometer.rst
@@ -3,89 +3,85 @@ Accelerometer
 
 .. py:module:: microbit.accelerometer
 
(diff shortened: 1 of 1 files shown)


microbit-carlos:
@microbit-mark could you give this a quick review when you get a chance?

Review URL: https://microbit-micropython--733.org.readthedocs.build/en/733/accelerometer.html

microbit-mark:
LGTM Thanks @microbit-carlos

PR #738, opened 2022-04-14
docs: Fix parameter names for display show() and scroll().

microbit-carlos:
Thanks @microbit-matt-hillsdon for noticing!


microbit-carlos (commented):
I've tested this on V1 and V2, and as this is a straightforward fix to the docs I'll merge this as it is.

PR #740, opened 2022-04-21
docs: Add do docstring notation to audio & add it's also in microbit module.

microbit-carlos:
The `audio` module now is also available via `microbit.audio` in V1 as it is in V2.

The PR also updates the content to use docstring notation like `:para xxx:`.


microbit-carlos:
@microbit-mark could you review once you get a chance?

I'll have to rebase v2-docs again after merging this and will make sure the V2 content is properly merged with this conflicting changes.

Thanks!

docs/audio.rst
```diff
@@ -38,6 +41,14 @@ Classes
 
     It takes just over 4 ms to play a single frame.
 
+    .. py:function:: copyfrom(other)
+
+        Overwrite the data in this ``AudioFrame`` with the data from another
+        ``AudioFrame`` instance.
+
+        :param other: ``AudioFrame`` instance to copy the data from.
```
microbit-carlos, line 49: @microbit-mark do you have a suggestion on how to rephrase this in a way that it doesn't end with a preposition?

microbit-carlos:
Also backported the tweaks in the "Technical Details" section from the v2-docs branch, commit https://github.com/bbcmicrobit/micropython/commit/38403c07fb3b10781c1df615263fcd69dfb98860#diff-1eb5832ad8abd778154545f6342a1327ba414d6f61c082d12352b56c511757e1R90

docs/audio.rst
microbit-mark: Would this make more sense?

```suggestion
        :param other: ``AudioFrame`` instance from which to copy the data.
```

microbit-carlos: Perfect, thanks!

PR #743, opened 2022-04-25
docs: Branch with the V2 documentation

microbit-carlos:
This draft PR shows the doc updates for micro:bit V2, which is deployed to ReadTheDocs as the default entry point, which expands to: https://microbit-micropython.readthedocs.io/en/v2-docs/

Many of the changes to the `v2-docs` branch that were not specific to V2 have been backported to the V1 docs, and this branch will be constantly rebased.

Having this as opened as PR makes GitHub track the changes between rebases (it adds a timeline entry linking to the old tree), so it can be used for traceability in case some merge/rebase goes wrong.


microbit-carlos:
As I had the same `v2-docs` branch in my fork I must have accidentally opened this PR pointing to my fork instead of the branch on this repo, so I've opened https://github.com/bbcmicrobit/micropython/pull/746 with the right branch.

PR #744, opened 2022-04-25
docs: Add note about possible 250kbit/sec radio rate in V2.

microbit-carlos:
Follow up from https://github.com/bbcmicrobit/micropython/pull/731

In this case removing the `RATE_250KBIT` constant, that is no longer availabe in V2, and adding a note on how to configure the data_rate in the `config()` function.

Review deployment: https://microbit-micropython--744.org.readthedocs.build/en/744/radio.html

image

### DIFF
--- a/docs/radio.rst
+++ b/docs/radio.rst
@@ -36,12 +36,6 @@ Constants
 
     Constant used to indicate a throughput of 2 Mbit a second.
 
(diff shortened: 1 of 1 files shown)


microbit-mark:
LGTM thanks @microbit-carlos

microbit-carlos:
Thanks Mark!

PR #753, opened 2022-06-20
docs: Add Sound Effects documentation.

microbit-carlos:
## Proposal

The current proposal adds a new class in the audio module named `SoundEffect`, which can be played using the `audio.play()` function.

The user can then create new instances, configuring them via constructor arguments, and after creation they can be modified via attributes.

A collection of a Sound Effects could be played by grouping them in a list (or something more advanced like a generator, so essentially an iterable).

Preview build of the docs: https://microbit-micropython--753.org.readthedocs.build/en/753/audio.html

### In progress

- [x] The default values have not been considered yet, the current values are more or less random
    - More info in this comment: https://github.com/bbcmicrobit/micropython/pull/753#issuecomment-1172629876
- [ ] The descriptions and explanations could use some work, as the document is getting long a bit hard to digest, but we can update that later
- [ ] The order of the constructor arguments could be modified if it makes more sense to group them differently
- [ ] The built-in effects (if they are not dropped) are not defined yet. MakeCode has these examples, which we could reuse or adapt/modify
    - https://github.com/microsoft/pxt/blob/v8.1.1/webapp/src/components/soundEffectEditor/soundUtil.ts#L52-L182
    - We should reduce the total number of provided Sound Effects, as their usefulness is limited (compared with their availability in the editors) and take additional flash space
    - We are currently planning to play with this feature at europython and attendees can then submit their Sounds Effects

## Things Still Under Consideration

1. `Effect()` might not be the best name for this feature
    - We might want to use the word "effect" for other future features, like filters than could be applied to recorded sound before it is played back
    - `SoundEffect()` is better but also longer (and we prefer shorter names that are easier to type for young learners)
    - This PR description has been updated already to reflect the `SoundEffect` name as described in comment https://github.com/bbcmicrobit/micropython/pull/753#issuecomment-1172629876
2. The parameter name `interpolation` is too long, we are still looking for alternatives.
    - `curve` is being considered, but one of the options is actually `INTER_CURVE`, so we might have to rename that to something else 
    - `ramp`?
3. The idea to provide built-in pre-created Sound Effects was to let the user easily create and modify an Effect. In the MakeCode Editor and Python Editor the user can select one from a list, and the code/block is automatically generated showing all the values used to create that sound.

### A new function to play an effect instead of creating a class instance

We know that some beginner programmers struggle with the concept of instantiating classes in variables, and then using attributes to modify them. An example of this would be NeoPixels, although on the other hand things like `Images` where a new instance is immediately used in a function has generally been okay. So:

```python
# This has generally been more difficult
np = neopixel.NeoPixel(pin0, 8)
np[0] = (255, 0, 128) 
np.show()

# But this has been okay, which is similar to audio.play(Effect(...))
display.show(Image('00300:'
                   '03630:'
                   '36963:'
                   '03630:'
                   '00300'))
```

So we are considering providing a function to play a single Sound Effect:

```python
# Configure the effect on the accelerometer data
my_effect = audio.play_effect(
    freq_start=400,
    freq_end=2000,
    duration=500,
    vol_start=100,
    vol_end=255,
    wave=audio.WAVE_TRIANGLE,
    fx=audio.FX_VIBRATO,
    interpolation=audio.INTER_LOG
)

# And using sensor data
while True:
   audio.play_effect(
        freq_start=accelerometer.get_x(),
        freq_end=accelerometer.get_y(),
        duration=500,
        vol_start=100,
        vol_end=255,
        wave=audio.WAVE_TRIANGLE,
        fx=audio.FX_VIBRATO,
        interpolation=audio.INTER_LOG
    )
```

If we adopt this we might need a better name that `play_effect(...)` as it is not that different than `play(Effect(...))`.

## Implementation details not shown in the docs

- Classes normally have a longer "internal" name (e.g. `Image`-> `MicroBitImage`, so the `Effect` class could have the longer `MicroBitSoundEffect` internal name
- Built-in pre-made Effects should be immutable
- Serialisation of a Sound Effect, for example to be sent via radio, could be done in a string format that could be eval'ed directly
    - The provided constants depend on the way the audio module is imported (`import audio` vs `from audio import *` vs `from microbit import *`), so their value would have to be used directly
        - For these values to be human readable we might need to use strings
    - i.e. `print(audio.Effect(vol_start=100)` -> `Effect(freq_start=400, freq_end=2000, duration=500, vol_start=100, vol_end=255, wave="triangle", fx="vibrato", interpolation="log")`
    - If the worse case scenario doesn't fit in the default radio packet size (or the max size), maybe we could use positional arguments for `__repr__` and keyword arguments for `__str__`?
        - The example from before would become `Effect(None, 400, 2000, 500, 100, 255, "triangle", "vibrato", "log")` 
- Not all the CODAL effect parameters are exposed to the user, for example, the `fx` has parameters that can be tweaked.
    - The hardcoded parameters can be found in the MakeCode implementation: https://github.com/microsoft/pxt-microbit/blob/v5.1.1/libs/core/soundexpressions.ts#L370
- By default `audio.play()` sets `wait=False` so it's a blocking function. If the CODAL call async version of play is called, and multiple calls are done inmediately one after another, they will not overlap.
    - So these three sound effects will still play one after another
      ```
      audio.play(audio.Effect.CHIRP, wait=False)
      audio.play(audio.Effect.CHIRP, wait=False)
      audio.play(audio.Effect.CHIRP, wait=False)
      ```
      But this will show the image while playing the sound
      ```
      audio.play(audio.Effect.CHIRP, wait=False)
      display.show(Image.HAPPY
      ```
    - We'll need to figure out how to document it, or if there is anything that can be done about this.

### DIFF
--- a/docs/audio.rst
+++ b/docs/audio.rst
@@ -209,6 +209,8 @@ But a new one can be created like this::
 Built in Sound Effects
 ----------------------
 
+⚠️ WARNING: These have not been created/implemented yet
+
 Some pre-created Sound Effects are already available as examples. These can
 be played directly ``audio.play(audio.SoundEffect.SQUEAK)``,
 or used as a base to create new effects
@@ -237,9 +239,9 @@ Sound Effects Example
         duration=500,
         vol_start=100,
         vol_end=255,
-        wave=audio.WAVE_TRIANGLE,
-        fx=audio.FX_VIBRATO,
-        interpolation=audio.LOG
+        wave=audio.SoundEffect.WAVE_TRIANGLE,
+        fx=audio.SoundEffect.FX_VIBRATO,
+        interpolation=audio.SoundEffect.INTER_LOG
     ))
 
     # Play a Sound Effect instance, modify an attribute, and play it again
@@ -315,7 +317,7 @@ samples. When reading reaches the start or the mid-point of the buffer, it
 triggers a callback to fetch the next ``AudioFrame`` which is then copied into
 the buffer. This means that a sound source has under 4ms to compute the next
 ``AudioFrame``, and for reliable operation needs to take less 2ms (which is
-32k cycles in micro:bit V1 or 128k in V2, so should be plenty).
+32000 cycles in micro:bit V1 or 128000 in V2, so should be plenty).
 
 
 AudioFrame Example
--- a/docs/audio.rst
+++ b/docs/audio.rst
@@ -23,7 +23,7 @@ There are three different kinds of audio sources that can be played using the
     my_effect = audio.SoundEffect(freq_start=400, freq_end=2500, duration=500)
     audio.play(my_effect)
 
-3. `Audio Frames <##audioframe>`_, an iterable (like a list or a generator)
+3. `Audio Frames <#audioframe>`_, an iterable (like a list or a generator)
    of Audio Frames, which are lists of 32 samples with values from 0 to 255::
 
     square_wave = audio.AudioFrame()
@@ -114,7 +114,7 @@ Sound Effects **V2**
 ====================
 
 .. py:class::
-    SoundEffect(preset=None, freq_start=500, freq_end=2500, duration=500, vol_start=255, vol_end=0, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LOG)
+    SoundEffect(freq_start=500, freq_end=2500, duration=500, vol_start=255, vol_end=0, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LOG)
 
     An ``SoundEffect`` instance represents a sound effect, composed by a set of
     parameters configured via the constructor or attributes.
@@ -124,9 +124,6 @@ Sound Effects **V2**
     can first create an effect ``my_effect = SoundEffect(duration=1000)``,
     and then change its attributes ``my_effect.duration = 500``.
 
-    :param preset: An existing SoundEffect instance to use as a base, its values
-        are cloned in the new instance, and any additional arguments provided
-        overwrite the base values.
     :param freq_start: Start Frequency in Hertz (Hz), eg: ``400``
     :param freq_end: End Frequency in Hertz (Hz), eg: ``2000``
     :param duration: Duration of the sound (ms), eg: ``500``
@@ -141,6 +138,10 @@ Sound Effects **V2**
         frequencies, different wave shapes have different rates of change
         in frequency. One of the following values: ``INTER_LINEAR``,
         ``INTER_CURVE``, ``INTER_LOG``.
+
+    .. py:function:: copy()
+
+        :returns: A copy of the SoundEffect.
 
     .. py:attribute:: freq_start
 
@@ -189,7 +190,7 @@ function, or by using a function that does the conversion automatically like
 For example, with the :doc:`REPL </devguide/repl>` you can inspect the built
 in Effects::
 
-    >>> audio.SoundEffect.CROAK
+    >>> print(audio.SoundEffect.CROAK)
     SoundEffect(freq_start=..., freq_end=..., duration=..., vol_start=..., vol_end=..., wave=..., fx=..., interpolation=...)
 
 The built in Effects are immutable, so they cannot be changed. Trying to modify
@@ -198,11 +199,11 @@ a built in SoundEffect will throw an exception::
     >>> audio.SoundEffect.CLICK.duration = 1000
     Traceback (most recent call last):
         File "<stdin>", line 1, in <module>
-    TypeError: effect cannot be modified
+    TypeError: SoundEffect cannot be modified
 
-But a new one can be created like this::
+But a new one can be created using the ``copy()`` method::
 
-    >>> click_clone = SoundEffect(audio.SoundEffect.CLICK)
+    >>> click_clone = audio.SoundEffect.CLICK.copy()
     >>> click_clone.duration = 1000
     >>>
 
@@ -213,8 +214,7 @@ Built in Sound Effects
 
 Some pre-created Sound Effects are already available as examples. These can
 be played directly ``audio.play(audio.SoundEffect.SQUEAK)``,
-or used as a base to create new effects
-``audio.SoundEffect(audio.SoundEffect.SQUEAK, duration=2000)``.
+or they can be cloned as a base to create new effects.
 
 * ``audio.SoundEffect.SQUEAK``
 * ``audio.SoundEffect.WARBLE``
@@ -257,7 +257,8 @@ Sound Effects Example
     # You can also create a new effect based on an existing one, and modify
     # any of its characteristics via arguments
     audio.play(audio.SoundEffect.WARBLE)
-    my_modified_effect = SoundEffect(audio.SoundEffect.WARBLE, duration=1000)
+    my_modified_effect = audio.SoundEffect.WARBLE.copy()
+    my_modified_effect.duration=1000
     audio.play(my_modified_effect)
 
     # Use sensor data to modify and play the existing Sound Effect instance
--- a/docs/audio.rst
+++ b/docs/audio.rst
@@ -114,7 +114,7 @@ Sound Effects **V2**
 ====================
 
 .. py:class::
-    SoundEffect(preset=None, freq_start=400, freq_end=200, duration=500, vol_start=100, vol_end=255, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LINEAR)
+    SoundEffect(preset=None, freq_start=500, freq_end=2500, duration=500, vol_start=255, vol_end=0, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LOG)
 
     An ``SoundEffect`` instance represents a sound effect, composed by a set of
     parameters configured via the constructor or attributes.
--- a/docs/audio.rst
+++ b/docs/audio.rst
@@ -20,7 +20,7 @@ There are three different kinds of audio sources that can be played using the
 2. `Sound Effects <#sound-effects-v2>`_ (**V2**), a way to create custom sounds
    by configuring its parameters::
 
-    my_effect = audio.Effect(freq_start=400, freq_end=2500, duration=500)
+    my_effect = audio.SoundEffect(freq_start=400, freq_end=2500, duration=500)
     audio.play(my_effect)
 
 3. `Audio Frames <##audioframe>`_, an iterable (like a list or a generator)
@@ -45,8 +45,8 @@ Functions
         - ``Sound``: The ``microbit`` module contains a list of
           built-in sounds, e.g. ``audio.play(Sound.TWINKLE)``. A full list can
           be found in the `Built in sounds <#built-in-sounds-v2>`_ section.
-        - ``Effect``: A sound effect, or an iterable of sound effects, created
-          via the :py:meth:`audio.Effect` class
+        - ``SoundEffect``: A sound effect, or an iterable of sound effects,
+          created via the :py:meth:`audio.SoundEffect` class
         - ``AudioFrame``: An iterable of ``AudioFrame`` instances as described
           in the `AudioFrame Technical Details <#id2>`_ section
 
@@ -114,17 +114,17 @@ Sound Effects **V2**
 ====================
 
 .. py:class::
-    Effect(preset=None, freq_start=400, freq_end=200, duration=500, vol_start=100, vol_end=255, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LINEAR)
+    SoundEffect(preset=None, freq_start=400, freq_end=200, duration=500, vol_start=100, vol_end=255, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LINEAR)
 
-    An ``Effect`` instance represents a sound effect, composed by a set of
+    An ``SoundEffect`` instance represents a sound effect, composed by a set of
     parameters configured via the constructor or attributes.
 
     All the parameters are optional, with default values as shown above, and
     they can all be modified via attributes of the same name. For example, we
-    can first create an effect ``my_effect = Effect(duration=1000)``, and then
-    change its attributes ``my_effect.duration = 500``.
+    can first create an effect ``my_effect = SoundEffect(duration=1000)``,
+    and then change its attributes ``my_effect.duration = 500``.
 
-    :param preset: An existing Effect instance to use as a base, its values
+    :param preset: An existing SoundEffect instance to use as a base, its values
         are cloned in the new instance, and any additional arguments provided
         overwrite the base values.
     :param freq_start: Start Frequency in Hertz (Hz), eg: ``400``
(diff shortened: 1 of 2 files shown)


microbit-carlos:
@dpgeorge @jaustin @microbit-giles This PR is ready for review.

As we'd like not to delay the implementation too much it'd be appreciated if any feedback could be provided soonish.

microbit-carlos:
Oh, and @microbit-matt-hillsdon, I think you might be interested to have a look at these as well at some point for the stubs.

microbit-giles:
My main concern is that the name `Effect()` may cause confusion and problems later on, especially if we implement sound 'filters' like reverb, amplification and reverse which are known as 'Effects' in widely-used sound editing software. We also have an 'fx' parameter within `Effect()`. Although longer to type, I think `SoundEffect()` may be clearer and less ambiguous in future.

microbit-matt-hillsdon:
> Oh, and @microbit-matt-hillsdon, I think you might be interested to have a look at these as well at some point for the stubs.

Thanks, yes, I think I'd like to have a stubs branch for this so we can spot any issues typing the API early.

docs/audio.rst
```diff
@@ -83,8 +86,215 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+Some pre-created Sound Effects are already available as examples. These can
+be played directly ``audio.play(audio.Effect.SQUEAK)``, or used as a base to
+create new effects ``audio.Effect(audio.Effect.SQUEAK, duration=2000)``.
+
+* ``audio.Effect.SQUEAK``
+* ``audio.Effect.WARBLE``
+* ``audio.Effect.CHIRP``
+* ``audio.Effect.CROAK``
+* ``audio.Effect.CLICK``
+
```
jaustin, line 221: I was really happy with including these, but I think their relationship to the built-in sounds is curious to explain. I don't know that this means we shouldn't have them, but I'm struggling to work out what to call them. 

`audio.SoundEffect.CROAK` for example is sort of weird to distinguish from `Sound.TWINKLE` - why can one of them be used as a preset and not the other? ("because one is a SoundEffect and one is a Sound"?)

This is one downside to "SoundEffect" over pure "Effect" - it ties the built-in sounds more closely to the SoundEffects - though @microbit-carlos points out that it does make sense that a 'Sound' is made out of SoundEffects...

docs/audio.rst
```diff
@@ -83,8 +86,215 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+        freq_start=400,
+        freq_end=2000,
+        duration=500,
+        vol_start=100,
+        vol_end=255,
+        wave=audio.WAVE_TRIANGLE,
+        fx=audio.FX_VIBRATO,
+        interpolation=audio.LOG
+    ))
+
```
jaustin, line 243: I like this, it's nice and clear, but also think we should give some thought to when to document the positional, not keyword options.

@microbit-matt-hillsdon when autocomplete works well with positional arguments, it can be a really smooth editing experience - are there ways we could document the code here so that we get good positional argument help during autocomplete as long as the user hasn't used any keyword arguments 

(for example, if I start typing `audio.play(Effect(400` will the autocomplete in the editor be showing the parameter help from freq_start (as long as we re-order the parameters so start is first))

docs/audio.rst
```diff
@@ -83,8 +86,215 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+    # any of its characteristics via arguments
+    audio.play(audio.Effect.WARBLE)
+    my_modified_effect = Effect(audio.Effect.WARBLE, duration=1000)
+    audio.play(my_modified_effect)
+
+    # Use sensor data to modify and play the existing Sound Effect instance
+    while True:
+        my_effect.freq_start=accelerometer.get_x()
+        my_effect.freq_end=accelerometer.get_y()
+        audio.play(my_effect)
```
jaustin, line 264: This is really nice and much clearer than also containing *all* the parameters and @microbit-giles I'd like us to make sure we include some nice lean examples in the edu content that encourage people to use this cleaner interface.

docs/audio.rst
```diff
@@ -83,8 +86,215 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+
+        sleep(500)
+        display.clear()
+
+
+Sound Effects **V2**
+====================
+
+.. py:class::
+    Effect(preset=None, freq_start=400, freq_end=200, duration=500, vol_start=100, vol_end=255, wave=WAVE_SQUARE, fx=None, interpolation=INTER_LINEAR)
```
microbit-carlos, line 117: One thing I am not very happy about this ordering is that `interpolation` is directly related to the `freq_start` and `freq_end`, but as it is less important/impactful than the other attributes, positionally is lower down the list.

To mitigate this we can have examples with just just this parameters in this order: `Effect(freq_start=...,  freq_end=..., interpolation=...)`

jaustin (commented):
This is looking great. I think SoundEffect overall is better than Effect, but one minor counter argument amongst the comments (that it confuses it more with the built-in sounds).

Still open to other suggestions for the name. I think if it were a function to create these I'd expect it to be something like `audio.generate(...)` or something related to a synth. It's the act of having to name the object that's hard.

microbit-carlos:
Okay, based on @microbit-giles and @jaustin suggestion I've renamed `Effect()` to `SoundEffect()` in b89aa5d0f53cba0147e665da880c945f229b9246

docs/audio.rst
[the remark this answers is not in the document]
microbit-carlos: I agree that on first impressions is really attractive to duplicate the MakeCode parameter order, as that would technically offer us the advantage of copy/paste arguments from the output of the really nice MakeCode widget:

image


However, as the `wave`, `fx`, and `interpolation` parameters depend on constants defined in MicroPython this is not possible, so I think we should prioritise an order that makes more sense for MicroPython users.
Although it is true, that using keyword arguments we do have the flexibility on how tutorials or the users order them.

microbit-carlos:
@dpgeorge I think we are at a point were we can implement the current version of the docs (minus the addition of the pre-made built-in SoundEffects), so that we can start playing with the feature and get a feeling for it.

For that I think it might be good to keep the implementation in a feature branch, so that we can still tweak it before merging it.

Also we would like to of bringing the Python Editor beta with a MicroPython build with Sound Effects for workshops at EuroPython. I think it would be really cool if we could play with the SoundEffects with people at the conference and if they'd like to contribute their sounds to an example file, we could then pick some of those to use as the pre-build SoundEffects and credit them.
The conference starts on the 11th of July, do you think that is doable? Ideally with a bit of time for us to be able to make sure any workshop material we prepare works.

docs/audio.rst
dpgeorge: The existing built-in sounds could be rewritten as a (constant) tuple of (constant) `SoundEffect` instances.  This is similar to sequences of images, like all the clock faces.

Eg:
```python
>>> audio.Sound.SPRING
(SoundEffect(...), SoundEffect(...))
```
Then users could really see how `Sound`'s are made and tweak them if they like.

dpgeorge:
> The parameter name `interpolation` is too long, we are still looking for alternatives

How about `shape`?  That matches the internal CODAL name.

docs/audio.rst
```diff
@@ -104,11 +315,11 @@ samples. When reading reaches the start or the mid-point of the buffer, it
 triggers a callback to fetch the next ``AudioFrame`` which is then copied into
 the buffer. This means that a sound source has under 4ms to compute the next
 ``AudioFrame``, and for reliable operation needs to take less 2ms (which is
-32000 cycles, so should be plenty).
+32k cycles in micro:bit V1 or 128k in V2, so should be plenty).
```
dpgeorge, line 318: 32k could mean 32768.  I'd leave it as 32000.

docs/audio.rst
```diff
@@ -83,8 +86,216 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+        are cloned in the new instance, and any additional arguments provided
+        overwrite the base values.
+    :param freq_start: Start Frequency in Hertz (Hz), eg: ``400``
+    :param freq_end: End Frequency in Hertz (Hz), eg: ``2000``
+    :param duration: Duration of the sound (ms), eg: ``500``
+    :param vol_start: Start volume value, range 0-255, eg: ``120``
+    :param vol_end: End volume value, range 0-255, eg: ``255``
+    :param wave: Type of wave shape, one of these values: ``WAVE_SINE``,
+        ``WAVE_SAWTOOTH``, ``WAVE_TRIANGLE``, ``WAVE_SQUARE``,
+        ``WAVE_NOISE`` (randomly generated noise).
```
dpgeorge, line 137: Where do these `WAVE_SINE` etc constants live?  I see in examples below they are at the top-level in `audio`, eg `audio.WAVE_SINE`.

The other place would be `SoundEffect.WAVE_SINE`.

dpgeorge: Since built-in sound effects live in `audio.SoundEffect` (like `audio.SoundEffect.CLICK`), it makes sense that these constants live there as well.

docs/audio.rst
```diff
@@ -83,8 +86,216 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+    parameters configured via the constructor or attributes.
+
+    All the parameters are optional, with default values as shown above, and
+    they can all be modified via attributes of the same name. For example, we
+    can first create an effect ``my_effect = SoundEffect(duration=1000)``,
+    and then change its attributes ``my_effect.duration = 500``.
+
+    :param preset: An existing SoundEffect instance to use as a base, its values
+        are cloned in the new instance, and any additional arguments provided
+        overwrite the base values.
```
dpgeorge, line 129: The interaction with a preset and default keyword arguments is problematic.

Consider `SoundEffect()`.  Does that create a new instance with all default values?

Consider `SoundEffect(vol_end=255)`.  Does that create a new instance with all default values except `vol_end` which is overridden to 255?

Consider `SoundEffect(audio.SoundEffect.CLICK)`.  Does that copy `CLICK` and then override everything with the default values?  If not, how do you explain the difference between that and `SoundEffect()` behaviour?

-----

How about changing the constructor signature to:
```python
SoundEffect(preset=SoundEffect.DEFAULT, *, freq_start, freq_end, duration, vol_start, vol_end, wave, fx, interpolation)
```
Then `SoundEffect.DEFAULT` has the default values for the keyword-only args as previously specified above (freq_start=500 etc).

dpgeorge:
The existing `Sound` class lives in `microbit.Sound`, and its `microbit.audio.AudioFrame`.  So where should `SoundEffect` go, in `microbit` or `microbit.audio`?

docs/audio.rst
microbit-carlos: > Consider SoundEffect(). Does that create a new instance with all default values?

Yes, I think this one is straight forward.

> Consider SoundEffect(vol_end=255). Does that create a new instance with all default values except vol_end which is overridden to 255?

Yes, as it would if the `preset` argument didn't exist.


> Consider `SoundEffect(audio.SoundEffect.CLICK)`. Does that copy `CLICK` and then override everything with the default values? If not, how do you explain the difference between that and `SoundEffect()` behaviour?

Okay, I could see how maybe that is a bit less obvious. 
The way I would see it is that `SoundEffect(audio.SoundEffect.CLICK)` is like a copy constructor, so it ignores any of the `__init__` default values.
I think `SoundEffect(audio.SoundEffect.CLICK, duration=2000)` could be easily read as "copy all `audio.SoundEffect.CLICK` values, but overwrite `duration` to `2000`, but I could see how that might not be as obvious if written as  `SoundEffect(duration=2000, preset=audio.SoundEffect.CLICK)`.

>Then SoundEffect.DEFAULT has the default values for the keyword-only args as previously specified above (freq_start=500 etc).

The disadvantage of that is we wouldn't be able to created shorter strings for serialisation with positional arguments.
So this shorter string wouldn't eval:
```
SoundEffect(988, 440, 190, 255, 255, 0, 1, 8)
```

If the confusion arising from having a `preset` argument in the constructor, what if the only way to duplicate and modify would be to use a `copy()` method?
```
my_effect = audio.SoundEffect.CLICK.copy()
my.effect.duration = 2000
```
It's an extra line, but might be easier to understand?

docs/audio.rst
microbit-carlos: Yeah, good call, it'll change I'll use the full figure. 👍

docs/audio.rst
microbit-carlos: Yeah, it makes sense to have them inside `audio.SoundEffect`, my worry was that it become too verbose:

```
audio.play(audio.SoundEffect(interpolation=audio.SoundEffect.INTER_LOG))
```

But to be fair, the difference is not that large, and it also helps with the editor autocompletion to not have that much directly inside the audio module.

```
audio.play(audio.SoundEffect(interpolation=audio.INTER_LOG))
```

docs/audio.rst
```diff
@@ -83,8 +86,216 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+
+        Effect to add on the sound, one of the following values:
+        ``FX_TREMOLO``, ``FX_VIBRATO``, ``FX_WARBLE``, or ``None``.
+
+    .. py:attribute:: interpolation
+
+        The type of curve between the start and end
+        frequencies, different wave shapes have different rates of change
+        in frequency. One of the following values: ``INTER_LINEAR``,
+        ``INTER_CURVE``, ``INTER_LOG``.
```
microbit-carlos, line 181: Creating a comment here to be able to keep this discussion thread in parallel.

From the PR description:
> The parameter name `interpolation` is too long, we are still looking for alternatives.
> * `curve` is being considered, but one of the options is actually `INTER_CURVE`, so we might have to rename that to something else
> * `ramp`?

microbit-carlos: Adding @dpgeorge's comment from https://github.com/bbcmicrobit/micropython/pull/753#issuecomment-1179641170 here as well:
> > The parameter name `interpolation` is too long, we are still looking for alternatives
> 
> How about `shape`? That matches the internal CODAL name.

microbit-carlos: `shape` could work! It's probably better than `ramp` as well, although maybe it's a matter of opinion.

@microbit-giles For the MicroPython SoundEffects, what are your thoughts about changing the name of `interpolation` (long and not a well known word for young coders) to `shape`?

docs/audio.rst
dpgeorge: > SoundEffect(988, 440, 190, 255, 255, 0, 1, 8)

This won't work because 988 is not a valid preset.  It would need to be `SoundEffect(None, 988, 440, 190, 255, 255, 0, 1, 8)`.

Actually, the way it's implemented at the moment is that if you pass `None` as the preset argument then the preset is the default set of values.  And any extra keyword args override that default /None preset value.

The signature for that is:
```python
SoundEffect(preset=None, *, freq_start, freq_end, duration, vol_start, vol_end, wave, fx, interpolation)
```

docs/audio.rst
dpgeorge: I have added these constants to `audio.SoundEffect`.

microbit-carlos:
> The existing `Sound` class lives in `microbit.Sound`, and its `microbit.audio.AudioFrame`. So where should `SoundEffect` go, in `microbit` or `microbit.audio`?

The original proposal was to have these inside the `audio` module,  as it's related to things like the `audio.AufioFrame`. So it ends up looking like this: `audio.play(audio.SoundEffect())`

The reason we ended up putting `Sound` in the `microbit` module, instead of the `audio` module was:
1) We considered these sounds to be micro:bit specific, to be part of the micro:bit personality, i.e. playing the "micro:bit HAPPY sound".
2) Shorter (`audio.Sound.HAPPY` vs `Sound.HAPPY`) always helps

For now the test branch has it as `audio.SoundEffect`, but we should reconsider if there is a good use case for having it as `microbit.SoundEffect`.

docs/audio.rst
microbit-carlos: > > SoundEffect(988, 440, 190, 255, 255, 0, 1, 8)
> 
> This won't work because 988 is not a valid preset. It would need to be `SoundEffect(None, 988, 440, 190, 255, 255, 0, 1, 8)`.

Yes, sorry, that would only work like that if we moved `preset` to be the last parameter, as it was suggested in https://github.com/bbcmicrobit/micropython/pull/753#discussion_r911437405. I thought I had added a comment somewhere to indicate that having it as the last parameter would help with `__repr__`, but I couldn't find it, so I probably forgot to write it down.

-----

The more I think about it, the better the `copy()` method sounds. Do you have any thoughts about it?

docs/audio.rst
microbit-carlos: Okay, I'll modify the docs to reflect that 👍

docs/audio.rst
microbit-carlos: Yes, this is something that was quite desirable, our only concern is all the extra flash usage, specially if the strings are already in CODAL and probably cannot be excluded as they are.
Would be good to get an estimation of how much extra flash space this would consume?

docs/audio.rst
microbit-carlos: Done in c0a2909120496aa33117d07573c5cd54d18fad60.

docs/audio.rst
microbit-carlos: It's been agreed to remove the `preset` constructor parameter, and add a `copy()` method to be able to clone SoundEffects.

docs/audio.rst
```diff
@@ -83,8 +86,215 @@ The built-in sounds can be called using ``audio.play(Sound.NAME)``.
...
+
+    .. py:attribute:: freq_start
+
+        Start Frequency in Hertz (Hz)
+
+    .. py:attribute:: freq_end
+
+        End Frequency in Hertz (Hz)
+
+    .. py:attribute:: duration
```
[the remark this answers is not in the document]
microbit-carlos: Already using duration, so let's keep it.

docs/audio.rst
microbit-carlos: We'll keep this order.

microbit-carlos:
Merging this as it is the current state of SoundEffect in the v2.1.0-beta.1 release.
Will create a new PR to include proposals on predefined sounds.

microbit-carlos:
- https://github.com/bbcmicrobit/micropython/pull/765

PR #754, opened 2022-06-20
docs: Add Power Management documentation.

microbit-carlos:
## Proposal

The current proposal has been simplified to only have two sleep functions `power.deep_sleep()` and `power.off()` (as there is already `microbit.sleep()` we had to use different wording) and a function to configure the wake up sources `power.wake_source()`.

Preview build of the docs: https://microbit-micropython--754.org.readthedocs.build/en/754/power.html

## Things Still Under Consideration

- The MakeCode implementation, which is a relatively thin wrapper around CODAL, had to offer the option to "request sleep" which goes to sleep when the current fiber yields, or to "request sleep and block the fiber until it goes to sleep". For MicroPython I think we could have only the second option, as the usage of fibers is limited
    - MakeCode example for reference: https://makecode.microbit.org/_EoqLgCDgr9Vj
- CODAL also offers `lowPowerEnable()` and `lowPowerIsEnabled()` as barriers to disable sleep in important portions of code. With the current implementation I don't think is required to expose these to the MicroPython users, but if we find a use case that needs it we might have to reconsider.

### Wake up sources as `deep_sleep()` arguments

The reason the wake up sources have been separated to a different function call was allow setting the sources once during startup, so that programmes with multiple `deep_sleep()` calls didn't have to replicate these values around.

```python
from microbit import *
import radio

power.wake_source(pins=(pin0, pin1), buttons=button_a)

@run_every(s=1)
def check_pin_every_second():
    if pin2.read_digital():
        power.deep_sleep()

radio.on()
while True:
    if button_b.is_pressed():
        power.deep_sleep()
    if radio.receive() == 'sleep':
        power.deep_sleep()
```

However, this is the kind of thing that could be resolved by the user creating a wrapper function:

```python
from microbit import *
import radio

def go_to_sleep():
    power.deep_sleep(pins=(pin0, pin1), buttons=button_a)

@run_every(s=1)
def check_pin_every_second():
    if pin2.read_digital():
        go_to_sleep()

radio.on()
while True:
    if button_b.is_pressed():
        go_to_sleep()
    if radio.receive() == 'sleep':
        go_to_sleep()
```

One of the advantages of having the wake up sources as arguments is that it removes any ambiguity about the sources also being applicable to `power.off()`, as it only applies for `deep_sleep()`.

At this point I'm leaning towards this simplification, unless somebody can think of a good reason to keep them separated.

## Implementation details not shown in the docs

- The pin classes might need to be updated to be able to set them as wake up sources? It's an implementation detail, but might be worth discussing if there are trade offs.
- MakeCode extension implementation for reference: https://github.com/microbit-foundation/pxt-microbit-v2-power/blob/master/power.cpp
- We are providing a way to detect the wake up source


microbit-carlos:
@dpgeorge this is ready to do an initial implementation (although technically is lower priority than the Sound Effects), mostly to see if there are any technical reasons that might affect the functions as they are defined here.
The MakeCode implementation turned out a bit more complex than we anticipated, mostly to be able to deal with how all fibers go to sleep, but I think that should not affect MicroPython.

docs/power.rst
```diff
@@ -0,0 +1,161 @@
...
+    LED near the USB connector.
+
+    The micro:bit will only wake if the reset button is pressed, or if on
+    battery power, when a USB cable is connected.
+
+    The the board wakes up it will start for a reset state, so your programme
+    will start running from the beginning.
+
+
+.. py:function:: deep_sleep(ms=None, pins=None, buttons=None, run_every=False)
```
dpgeorge, line 84: `pins` and `buttons` could actually be combined into, eg, `events`.  Because the code can check the type of the objects passed in.  Eg `deep_sleep(events=(button_a, pin0))`.

docs/power.rst
```diff
@@ -0,0 +1,161 @@
...
+    operation where it left off.
+
+    The Deep Sleep mode will consume more battery power than Off mode.
+
+    The wake up sources are configured via arguments.
+    If no wake up sources have been configured it will sleep indefinitely.
+
+    :param ms: A time in milliseconds to wait before it wakes up.
+    :param pins: A single instance or a tuple of pins, e.g.
+        ``wake_source(pins=(pin0, pin2))``.
```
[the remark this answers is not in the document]
microbit-carlos: Good catch, I'll remove this mention.

docs/power.rst
microbit-carlos: Yeah, having the buttons for only 2 options doesn't feel quite right. Although I'm not 100% sure about the word `events` though, it doesn't feel very obvious `deep_sleep(events=(pin1,pin2))`, but it's warming up on me.

Do you think using `pins` and include buttons in the same argument would be confusing? `deep_sleep(pins=(pin1, button_a))`

dpgeorge: How about `wake_on` or `wake_by`?  Eg `deep_sleep(wake_on=(pin1, button_a))`.

microbit-carlos: `wake_on` sounds good 👍 
Added in 082574457fd3ec725e50730132f69bddbf279827.

docs/power.rst
```diff
@@ -0,0 +1,162 @@
...
+
+    Set the micro:bit into the Deep Sleep mode.
+
+    The programme state is preserved and on wake it will resume operation
+    where it left off.
+
+    The wake up sources are configured with the ``wake_source()`` function.
+    If no wake up sources have been configured it will sleep indefinitely.
+
+.. py:function:: wake_source(pins=None, buttons=None, ms=None, run_every=False)
```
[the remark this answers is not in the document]
microbit-carlos: This is resolved by combining `pins` and `buttons` into a single `wake_on` parameter: https://github.com/bbcmicrobit/micropython/pull/754#discussion_r950850595

docs/power.rst
dpgeorge: OK, now implemented.

dpgeorge:
As discussed elsewhere, my suggestion is that `run_every=True` does not end the `deep_sleep()` call, but rather wakes briefly to execute the run every and then resumes the deep sleep.  The deep sleep only ends when the given timeout expires, or a `wake_on` event occurs.  In particular, `power.deep_sleep(run_every=True)` will never return.  (And maybe we can then make `run_every=True` the default.)

microbit-carlos:
This has been shipped in v2.1.0-beta.1 in the current form, so I'll merge the docs, and any changes to sleep will need a new PR here as well.

Issue #761, opened 2022-08-08
New Feature: Function to scale values with different ranges


microbit-carlos:
To match the V2 implementation discussed in https://github.com/microbit-foundation/micropython-microbit-v2/issues/102

microbit-carlos:
Docs PR:
- https://github.com/bbcmicrobit/micropython/pull/762

dpgeorge:
Implemented in 256520fbc02a3004eb269a165d94e856016861f5

PR #762, opened 2022-08-17
docs: Add `microbit.scale(value, from_, to)` function.

microbit-carlos:
As discussed in:
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/102
- https://github.com/bbcmicrobit/micropython/issues/761

Preview URL: https://microbit-micropython--762.org.readthedocs.build/en/762/microbit.html#microbit.scale

### DIFF
--- a/docs/microbit.rst
+++ b/docs/microbit.rst
@@ -31,6 +31,25 @@ Functions
     :returns: The number of milliseconds since the board was switched on or
         restarted.
 
+.. py:function:: scale(value, from_, to)
+
+    Converts a value from a range to another range.
+
+    For example, to convert 30 degrees from Celsius to Fahrenheit::
+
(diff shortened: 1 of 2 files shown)


docs/microbit.rst
```diff
@@ -31,6 +31,22 @@ Functions
     :returns: The number of milliseconds since the board was switched on or
         restarted.
 
+.. py:function:: scale(value, from, to)
```
dpgeorge, line 34: Unfortunately `from` is a keyword in Python so cannot be used as an argument name.

How about `from_`?

microbit-carlos: Yeah, I can't think of a better word, so `from_` looks like it might be the best option. 👍

dpgeorge: OK, I'll use `from_`.

Issue #760, opened 2022-08-08
Automatically turn on radio on import


microbit-carlos:
To match V2 implementation as discussed in https://github.com/microbit-foundation/micropython-microbit-v2/issues/97.

dpgeorge:
Implemented by 8d2138720fa2cdadbeebba8a8f0a72563b891f20

PR #763, opened 2022-08-19
docs: Radio is now turned on during import.

microbit-carlos:
As discussed in:
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/97
- https://github.com/bbcmicrobit/micropython/issues/760

And implemented in https://github.com/bbcmicrobit/micropython/commit/8d2138720fa2cdadbeebba8a8f0a72563b891f20 and https://github.com/microbit-foundation/micropython-microbit-v2/commit/ca108e7d609c14b1c2a6c2e895eb019056a3e92d.

Preview URLs:
- https://microbit-micropython--763.org.readthedocs.build/en/763/radio.html
- https://microbit-micropython--763.org.readthedocs.build/en/763/tutorials/radio.html

### DIFF
--- a/docs/radio.rst
+++ b/docs/radio.rst
@@ -48,8 +48,13 @@ Functions
 
 .. py:function:: on()
 
-    Turns the radio on. This needs to be explicitly called since the radio
-    draws power and takes up memory that you may otherwise need.
+    Turns the radio on.
+
+    Since MicroPython-on-micro:bit v1.1
+    the radio is turned on by default when the radio module is imported.
+    In earlier releases, to reduce power consumption until needed, this
+    function had to be explicitly called.
+    For those cases ``radio.off()`` can be called after import.
 
 .. py:function:: off()
 
--- a/docs/tutorials/radio.rst
+++ b/docs/tutorials/radio.rst
@@ -107,10 +107,6 @@ We're going to use the radio module to create something akin to a swarm of
 fireflies signalling to each other.
 
 First ``import radio`` to make the functions available to your Python program.
-Then call the ``radio.on()`` function to turn the radio on. Since
-the radio draws power and takes up memory we've made it so *you* decide
-when it is enabled (there is, of course a ``radio.off()`` function).
-
 At this point the radio module is configured to sensible defaults that make
 it compatible with other platforms that may target the BBC micro:bit. It is
 possible to control many of the features discussed above (such as channel and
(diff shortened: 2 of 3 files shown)


docs/radio.rst
```diff
@@ -48,8 +48,13 @@ Functions
 
 .. py:function:: on()
 
-    Turns the radio on. This needs to be explicitly called since the radio
-    draws power and takes up memory that you may otherwise need.
+    Turns the radio on.
```
dpgeorge, line 51: Reading this now, I realise we did not consider the point that enabling the radio takes up RAM (as well as power), which is in very short supply on micro:bit v1.

Turning the radio on at import may break some code and be impossible to fix.  It could be the case that the user's program is very tight on memory and they do something like this:
```python
import radio
radio.config(length=10, queue=1)
radio.on()  # this is where the memory is allocated for radio buffers
```
In the approach now where it's turned on at import, it will allocate the default buffer space, which is length=32 and queue=3.  That default allocates (32+6)*(3+1) = 152 bytes.  Then in the above code it will free this and reallocate to the smaller size, which is (10+6)*(1+1) = 32 bytes.  The import may fail if the micro:bit doesn't have much free RAM left.

I'm not sure how much of an issue this is but it needs to be considered.

docs/radio.rst
```diff
@@ -48,8 +48,13 @@ Functions
...
+
+    Since MicroPython v1.1
```
dpgeorge, line 53: It's not really MicroPython v1.1, it's MicroPython-on-micro:bit v1.1.

microbit-carlos: Updated in [5fae7a5](https://github.com/bbcmicrobit/micropython/compare/05c30a665f0e3f71cb93b54a96add87501ad7550..5fae7a55cb9bdef994ea31ec155503dcf140ce1a)

docs/radio.rst
microbit-carlos: Yeah, printing the free memory size before and after `radio.on()` aligns to your calculation and shows a decrease of about 160 bytes.

You are right that this can cause issues, but I think most users will likely configure the radio very soon after import, as your example shows.

In those cases, hopefully, there is more than 120 bytes available at that point, as you'd expect the main.py file to contain at least some code to use memory as well.

We'll release this as a beta release in the Python Editor beta, and that way we can consider revert it if there are issues before the main release.

dpgeorge: I think it's unlikely to cause an issue using a little more memory like this.  I doubt many users realise they can reclaim a few 10's of bytes of RAM by reconfiguring the radio like this, so it's probably never done in practice (what's more likely is that they *increase* the length/queue size to get better radio performance).

Issue #713, opened 2021-03-22
music.play(pin=None) causes typeError, expecting a pin


martinwork:
Testing with https://python.microbit.org/v/beta
Editor Version: 2.2.0-beta.5;
MicroPython Version: 1.0.1/2.0.0-beta.5

An error occurs if music.play specifies pin=None, either directly or through a variable.

```python
import microbit
from microbit import *
import music

myPin=pin0

while True:
    music.play(music.JUMP_UP)
    sleep(1000)
    
    music.play(music.JUMP_UP,pin=myPin)
    sleep(1000)

    myPin=None
```

microbit-carlos:
Good catch, thanks Martin!
For compatibility with V2 we should to accept `None` as a `pin` value in V1 as well.

dpgeorge:
Fixed by 1252f887ddc790676bf9314a136bd17650b9c36c

PR #766, opened 2022-09-07
Allow music pin to be `None`

dpgeorge:
This is to address #713.

It required a refactor of the music code to more closely align with the v2 implementation.

The following functions now accept `None` for the pin argument:
- `music.stop()`
- `music.play()`
- `music.pitch()`


dpgeorge:
This increases the firmware size by 32 bytes, which is OK.

microbit-carlos (approved):
Looks good and works a treat, thanks Damien! 👍

PR #769, opened 2022-09-12
docs: Update Power Mgm to change run_every behaviour.

microbit-carlos:
From waking up the board when the next function is scheduled, to allow the scheduled functions to continue running indefinitely while the boards sleeps.

Review URL: https://microbit-micropython--769.org.readthedocs.build/en/769/power.html

### DIFF
--- a/docs/power.rst
+++ b/docs/power.rst
@@ -9,7 +9,7 @@ There are two micro:bit board low power modes that can be requested from
 MicroPython:
 
 - **Deep Sleep**: Low power mode where the board can be woken up via
-  multiple sources (pins, button presses, or a timer) and resume
+  multiple sources (pins, button presses, uart data, or a timer) and resume
   operation.
 - **Off**: The power mode with the lowest power consumption, the only way to
   wake up the board is via the reset button, or by plugging the USB cable while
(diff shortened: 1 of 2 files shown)


microbit-carlos:
@microbit-matt-hillsdon I think with the changes in this PR the only impact to the stubs and translation would be on the `run_every` parameter description, is that right?

examples/datalog-sleep.py
```diff
@@ -3,13 +3,14 @@
...
 
 while True:
+    # Display the temperature when button A is pressed
     if button_a.is_pressed():
-        # Display the temperature when button A is pressed
         display.scroll(temperature())
-    # To go sleep and wake up with run_every or button A
+    # To go sleep, wake up when button A is pressed, and ensure the
+    # function scheduled with run_every still executes in the background
     power.deep_sleep(wake_on=button_a, run_every=True)
```
dpgeorge, line 16: since `run_every=True` is the default, does it need to be explicitly specified here?

docs/power.rst
```diff
@@ -53,12 +53,18 @@ Functions
     button is pressed (which resets the board) or, in battery power,
     when the USB cable is inserted.
 
+    When the ``run_every`` parameter is set to ``True`` (the default), any
+    function scheduled with :py:meth:`microbit.run_every<microbit.run_every>`
+    will still run while the board sleeps. When the scheduled time is reached
+    the micro:bit will momentarily wake up to run the scheduled function
+    and then automatically go back to sleep.
```
dpgeorge, line 60: The second sentence here is more of an implementation detail, so I suggest making that clear, and giving more detail.  Eg:
```
Implementation note: To make sure the `run_every`'s continue to run, the
micro:bit will wake up from deep sleep at the correct time to run the next
scheduled `run_every`, and then go back to deep sleep mode as soon as
that `run_every` completes.  It will continue to do this until the deep sleep
finishes due to the `ms` timeout being reached, or a `wake_on` event
occurs.
```

dpgeorge:
> could you confirm this aligns with what we discussed

Yes!

examples/datalog-sleep.py
microbit-carlos: No, it's no longer necessary, but I though it'd be good to keep it to be explicit and also so that is clear when reading the example code that the `run_every` function will continue to run while it sleeps.

docs/power.rst
microbit-carlos: I think you are right, that paragraph probably contained a bit too much detail, so I tried to simplify it a bit (it says almost the same, but hopefully it's less technical), and then added the technical detail to the "Detailed Information" section in https://github.com/bbcmicrobit/micropython/pull/769/commits/18dd86d95a04e644cb0aa43beb4cba11acb942dc

microbit-carlos:
This change was shipped in v2.1.0, so merging to the docs.

PR #771, opened 2022-09-26
docs: Indicate when `microbit.scale()` returns float or int.

microbit-carlos:


microbit-carlos:
I'll merge this now, so that it is present on the v1.1.0 tag from this repo that will trigger a tag with the same name in readthedocs.

microbit-carlos (commented):
I'll merge this now, so that it is present on the v1.1.0 tag from this repo that will trigger a tag with the same name in readthedocs.

PR #787, opened 2022-12-02
docs: Update URL to download arm-none-eabi-gcc.

microbit-carlos:
The old URL still works, but it is not updated with the latest versions anymore.
Reported by @cleoqc in Mastodon, thanks!


CleoQc:
Thanks for the support on Mastodon!

PR #791, opened 2023-05-15
docs: Add initial proposal for V2 recording & playback API. (WIP)

microbit-carlos:
Docs preview:
- https://microbit-micropython--791.org.readthedocs.build/en/791/audio.html
- https://microbit-micropython--791.org.readthedocs.build/en/791/microphone.html

This initial proposal has been discussed in:
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/49

But we have some open question that will likely result and a rework of some of this.

## Initial proposal

The initial proposal in this PR was to create a new `AudioBuffer` class to contain the audio data and sampling rate.
The `AudioBuffer.rate` property could then be used by `microphone.record()` and `audio.play()`  to configure recording and playback rates.
This was done to avoid introducing a new parameter to `audio.play()` to configure the sampling rate, when it could only work with a single type of sound input (as it might not be possible to change the rate of the SoundExpressions or AudioFrames).

### Disadvantages

However, changing the rate in a buffer type to change the playback rate in real-time is a bit awkward:

```python
my_recording = audio.AudioBuffer(duration=5000, rate=5500)
microphone.record_into(my_recording)
audio.play(my_recording, wait=False)
while audio.is_playing():
    x = accelerometer.get_x()
    my_recording.rate = scale(x, (-1000, 1000), (2250, 11000))
    sleep(50)
```

An alternative we considered was to have the playback sampling rate modified via the audio module itself:

```python
audio.play(my_recording, wait=False)
while audio.is_playing():
    x = accelerometer.get_x()
    audio.set_rate(scale(x, (-1000, 1000), (2250, 11000)))
    sleep(50)
```

However, this would have to set the same rate to everything played via the audio module, and Sound Expression have a different default rate (44K) than recordings (11K). So `audio.set_rate(22000)` should slow down Sound Expression and speed up recordings.

Alternatively, if we wanted to change the playback rate via the audio module, we could set a ratio instead. Something equivalent to `audio.set_speed(100%)` (with different semantics). But a disadvantage would be that it's removing some of math/physics learning opportunity to directly relate the sampling rate value with the effects that it has in playback speed. 

## Alternative proposal: bytearray as the buffer type

In this case a byte array would be returned by `microphone.record()` and used with`microphone.record_into()`.

As this data type does not include info about the rate, we depend on the `audio.play()` adding an extra argument that might not work with other sound types like Sound Expressions and Audio Frames.

However, we still have the issue of updating the playback rate in real time during playback, which means we might would have to use use a similar approach to the previously mentioned `audio.set_speed(100%)`:

```python
sound_in_byte_array = microphone.record(duration=3000, rate=5500)
audio.play(sound_in_byte_array, rate=5500 wait=False)
while audio.is_playing():
    x = accelerometer.get_x()
    audio.set_speed(scale(x, (-1000, 1000), (50, 200)))
    sleep(50)
```

```python
DURATION_SECONDS = 3
SAMPLE_RATE = 5500
recording = bytearray(DURATION_SECONDS * SAMPLE_RATE)
microphone.record_into(recording, rate=SAMPLE_RATE)
audio.play(recording, rate=SAMPLE_RATE)
```

## Alternative proposal: AudioFrames as the buffer type

This would be the same as the bytearray proposal, but using the existing AudioFrames instead.

We might need to tweak the AudioFrame class to let us user larger buffers, as the default is 32 samples. As `audio.play()` can consume an iterable as well, we would need to figure out a good balance between AudioFrame size and number of AudioFrames in a recording buffer.

### DIFF
--- a/docs/audio.rst
+++ b/docs/audio.rst
@@ -12,7 +12,7 @@ a speaker to pin 0 and GND on the edge connector to hear the sounds.
 The ``audio`` module can be imported as ``import audio`` or accessed via
 the ``microbit`` module as ``microbit.audio``.
 
-There are four different kinds of audio sources that can be played using the
+There are three different kinds of audio sources that can be played using the
 :py:meth:`audio.play` function:
 
 1. `Built in sounds <#built-in-sounds-v2>`_ (**V2**),
@@ -24,14 +24,9 @@ There are four different kinds of audio sources that can be played using the
     my_effect = audio.SoundEffect(freq_start=400, freq_end=2500, duration=500)
     audio.play(my_effect)
 
-3. `AudioBuffer <#audiobuffer>`_ (**V2**), a generic buffer for audio that can
-   be used to record sound from the micro:bit V2 built-in microphone::
-
-    my_audio_buffer = microphone.record()
-    audio.play(my_audio_buffer)
-
-4. `Audio Frames <#audioframe>`_, an iterable (like a list or a generator)
-   of Audio Frames, which are lists of 32 samples with values from 0 to 255::
+3. `Audio Frames <#audioframe>`_, an instance or an iterable (like a list or
+   generator) of Audio Frames, which are lists of samples with values
+   from 0 to 255::
 
     square_wave = audio.AudioFrame()
     for i in range(16):
@@ -47,18 +42,16 @@ Functions
 
     Play the audio source to completion.
 
-    :param source: There are four types of data that can be used as a source:
+    :param source: There are three types of data that can be used as a source:
 
         - ``Sound``: The ``microbit`` module contains a list of
           built-in sounds, e.g. ``audio.play(Sound.TWINKLE)``. A full list can
           be found in the `Built in sounds <#built-in-sounds-v2>`_ section.
         - ``SoundEffect``: A sound effect, or an iterable of sound effects,
           created via the :py:meth:`audio.SoundEffect` class
-        - ``AudioBuffer``: An audio buffer, or an iterable of audio buffers,
-          created via the :py:meth:`audio.AudioBuffer` class or
-          :doc:`microphone.record() <microphone>` function
-        - ``AudioFrame``: An iterable of ``AudioFrame`` instances as described
-          in the `AudioFrame Technical Details <#id2>`_ section
+        - ``AudioFrame``: An instance or an iterable of ``AudioFrame``
+          instances as described in the
+          `AudioFrame Technical Details <#technical-details>`_ section
 
     :param wait: If ``wait`` is ``True``, this function will block until the
         source is exhausted.
@@ -78,6 +71,13 @@ Functions
 .. py:function:: stop()
 
     Stops all audio playback.
+
+.. py:function:: set_rate(sample_rate)
+
+    Changes the sampling rate of ``AudioFrame`` playback.
+    The default playback rate is 7812 samples per second.
+    Decreasing the playback sampling rate results in slowed down sound, and
+    increasing it speeds it up.
 
 
 Built-in sounds **V2**
@@ -226,70 +226,16 @@ Sound Effects Example
     :code: python
 
 
-Audio Buffer **V2**
-===================
-
-.. py:class::
-    AudioBuffer(duration=3000, rate=11000)
-
-    Create a buffer to contain audio data and its sampling rate.
-
-    The sampling rate is configured via constructor or instance attribute,
-    and is used by the ``microphone.record_into()`` and
-    ``audio.play()`` functions to configure the recording and playback rates.
-
-    For audio recording, reducing the number of samples recorded per second
-    will reduce the size of the data buffer, but also reduce the sound quality.
-    And increasing the sampling rate increases the buffer size and sound
-    quality.
-
-    For audio playback, reducing the sampling rate compared with the recording
-    rate, will slow down the audio. And increasing the playback rate
-    will speed it up.
-
-    The size of the buffer will be determined by the samples per second
-    and the ``duration`` configured when creating a new instance.
-
-    :param duration: Indicates in milliseconds, how much sound data the buffer
-        can contained at the configured ``data_rate``.
-    :param rate: Sampling rate of for the data in the buffer. This value is
-        used for recording and playback, and can be edited as an attribute.
-
-    .. py:function:: copy()
-
-        :returns: A copy of the Audio Buffer.
-
-    .. py:attribute:: rate
-
-        The sampling rate for the data inside the buffer.
-        TODO: Indicate range of valid values here.
-
-Audio Buffer Example
---------------------
-
-::
-
-    my_buffer = audio.AudioBuffer(duration=5000)
-    microphone.record_into(my_buffer)
-    audio.play(my_buffer)
-
-    # A smaller buffer can be generated with the same duration by using
-    # a lower sampling rate
-    smaller_buffer = audio.AudioBuffer(duration=5000, rate=5500)
-    microphone.record_into(my_buffer)
-    audio.play(my_buffer)
-
-
 AudioFrame
 ==========
 
 .. py:class::
-    AudioFrame
+    AudioFrame(size=32)
 
-    An ``AudioFrame`` object is a list of 32 samples each of which is an unsigned byte
-    (whole number between 0 and 255).
+    An ``AudioFrame`` object is a list of samples each of which is an unsigned
+    byte (whole number between 0 and 255).
 
-    It takes just over 4 ms to play a single frame.
+    :param size: How many samples to contain in this instance.
 
     .. py:function:: copyfrom(other)
 
@@ -305,13 +251,19 @@ Technical Details
     You don't need to understand this section to use the ``audio`` module.
     It is just here in case you wanted to know how it works.
 
-The ``audio`` module can consumes an iterable (sequence, like list or tuple, or
-generator) of ``AudioFrame`` instances, each 32 samples at 7812.5 Hz, and uses
-linear interpolation to output a PWM signal at 32.5 kHz, which gives tolerable
-sound quality.
+The ``audio.play()`` function can consume an instance or iterable
+(sequence, like list or tuple, or generator) of ``AudioFrame`` instances.
+Its default playback rate is 7812 Hz, and uses linear interpolation to output
+a PWM signal at 32.5 kHz.
 
-The function ``play`` fully copies all data from each ``AudioFrame`` before it
-calls ``next()`` for the next frame, so a sound source can use the same
+Each ``AudioFrame`` instance is 32 samples by default, but it can be
+configured to a different size via constructor.
+
+So, for example, playing 32 samples at 7812 Hz takes just over 4 milliseconds
+(1/7812.5 * 32 = 0.004096 = 4096 microseconds).
+
+The function ``play()`` fully copies all data from each ``AudioFrame`` before
+it calls ``next()`` for the next frame, so a sound source can use the same
 ``AudioFrame`` repeatedly.
 
 The ``audio`` module has an internal 64 sample buffer from which it reads
@@ -324,6 +276,9 @@ the buffer. This means that a sound source has under 4ms to compute the next
 
 AudioFrame Example
 ------------------
+
+Creating and populating ``AudioFrame`` iterables and generators with
+different sound waveforms:
 
 .. include:: ../examples/waveforms.py
     :code: python
--- a/docs/audio.rst
+++ b/docs/audio.rst
@@ -24,7 +24,22 @@ There are three different kinds of audio sources that can be played using the
     my_effect = audio.SoundEffect(freq_start=400, freq_end=2500, duration=500)
     audio.play(my_effect)
 
-3. `Audio Frames <#audioframe>`_, an instance or an iterable (like a list or
+3. `Audio Recordings <#audiorecording-audiotrack-v2>`_, an object that can
+   be used to record audio from the microphone::
+
+    recording = audio.AudioRecording(duration=4000)
+    microphone.record_into(recording)
+    audio.play(recording)
+
+4. `Audio Tracks <#audiorecording-audiotrack-v2>`_, a way to point to a portion
+   of the data in an ``AudioRecording`` or a ``bytearray`` and/or modify it::
+
+    recording = audio.AudioRecording(duration=4000)
+    microphone.record(recording)
+    track = AudioTrack(recording)[1000:3000]
+    audio.play(track)
+
+5. `Audio Frames <#audioframe>`_, an instance or an iterable (like a list or
    generator) of Audio Frames, which are lists of samples with values
    from 0 to 255::
 
@@ -49,6 +64,10 @@ Functions
           be found in the `Built in sounds <#built-in-sounds-v2>`_ section.
         - ``SoundEffect``: A sound effect, or an iterable of sound effects,
           created via the :py:meth:`audio.SoundEffect` class
+        - ``AudioRecording``: An instance of ``AudioRecording`` as described
+          in the `AudioRecording <#audiorecording-audiotrack-v2>`_ section
+        - ``AudioTrack``: An instance of ``AudioTrack`` as described in the
+          `AudioTrack <#audiorecording-audiotrack-v2>`_ section
         - ``AudioFrame``: An instance or an iterable of ``AudioFrame``
           instances as described in the
           `AudioFrame Technical Details <#technical-details>`_ section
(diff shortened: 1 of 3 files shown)


microbit-carlos:
Based on the latest discussion we have agreed that we'd prefer to avoid introducing a new data type for his feature.
In that case we have two options:

### Use a byte array and change sampling via function in the audio module

```
sound_in_byte_array = microphone.record(duration=3000, rate=5500)
audio.play(sound_in_byte_array, wait=False)
while audio.is_playing():
    x = accelerometer.get_x()
    playback_sampling_rate = scale(x, (-1000, 1000), (2_200, 11_000))
    audio.set_rate(playback_sampling_rate)
    sleep(50)
```

One previous suggestion was to create a function in the lines of `audio.set_speed()` that used a percentage value (0 to 100) as a the input range. This could solve the issue with different pipelines in the sound mixer having different playback sampling rates, however we believe it's important to use real numbers to be able to directly compare and understand how changing the sampling rate during recording vs playback affects sound.

To be able to implement something like `audio.set_rate()` we would probably need to ensure everything that is inputted to `audio.play()` uses the same default sampling rate. Right now these are the types of input `audio.play()` takes:
- Sound effects via user-created `audio.SoundEffect()` instances
    - CODAL `SoundExpressions`, default sampling rate: 44_100
- Built-in sounds via `microbit.Sound` pre-generated instances
    - CODAL `SoundExpressions`, default sampling rate: 44_100
- `audio.AudioFrame`
    - MicroPython data type, default sampling rate: 8_000?
        - Not sure about this one, the docs list as single frame of 32 samples to take a bit over 4ms.

We'd have to check how it'd affect sound quality, but we could consider decreasing the sampling rate for `SoundExpressions` to 11K. However, we would still have the isse that we won't be able to increase the default `AudioFrame` sampling rate as that would how old programmes sound.

While this would be the cleanest way to do this for the user API, I'm not seeing a way in which it can be achieved with the current requirements? Does anybody have any ideas to overcomes this issues?


### Expand AudioFrame to include sampling rate

This option is similar to the original proposals about creating a new `AudioBuffer` data type, but instead we can expand AudioFrames to be able to have different sizes and different sampling rates.

By default they should still behave as they do in micro:bit V1 (and older MicroPython versions for V2), which is 32 samples at 8K (?) sampling rate.

But this could be changed via constructor parameters to have any size buffer at any sample rate.

Unfortunately, that still leaves us with the awkward case of changing playback sampling rate by changing a variable from the samples class, instead of a method in the audio module.

```python
my_recording = audio.AudioBuffer(length=11000, rate=5500)
my_recording = microphone.record_into(my_recording)
# Or
my_recording = microphone.record(duration=2000, rate=5500)

audio.play(my_recording, wait=False)
while audio.is_playing():
    x = accelerometer.get_x()
    my_recording.rate = scale(x, (-1000, 1000), (2250, 11000))
    sleep(50)
```

microbit-carlos:
@dpgeorge the docs have been updated, let me know if something doesn't match our previous conversation.

microbit-carlos:
@dpgeorge there are couple of issues related to setting the sampling rate, but shouldn't affect the MicroPython implementation and should be fixed (without changes in the API) in the next CODAL release: https://github.com/lancaster-university/codal-microbit-v2/issues?q=is%3Aopen+milestone%3Av0.2.60+label%3Ap0

docs/audio.rst
```diff
@@ -69,6 +72,13 @@ Functions
 
     Stops all audio playback.
 
+.. py:function:: set_rate(sample_rate)
```
dpgeorge, line 75: Perhaps this could be made into a static-method of the `AudioFrame` class?  That way the scope of it is clear, it only acts on these objects.  And it would then be possible to write:
```py
AudioFrame.set_rate(8000)
...
my_frame = AudioFrame(...)
my_frame.set_rate(8000)
``

File: docs/audio.rst [hunk 2]
```diff
@@ -215,16 +225,17 @@ Sound Effects Example
 .. include:: ../examples/soundeffects.py
     :code: python
 
+
 AudioFrame
 ==========
 
 .. py:class::
-    AudioFrame
+    AudioFrame(size=32)
```
dpgeorge, line 233: Perhaps this could be extended to take optional keyword arguments `duration` and `rate` (with a default), to make it easy to create an `AudioFrame` of a given duration.

docs/audio.rst
```diff
@@ -240,13 +251,19 @@ Technical Details
...
     It is just here in case you wanted to know how it works.
 
-The ``audio`` module can consumes an iterable (sequence, like list or tuple, or
-generator) of ``AudioFrame`` instances, each 32 samples at 7812.5 Hz, and uses
-linear interpolation to output a PWM signal at 32.5 kHz, which gives tolerable
-sound quality.
+The ``audio.play()`` function can consume an instance or iterable
+(sequence, like list or tuple, or generator) of ``AudioFrame`` instances.
+Its default playback rate is 7812 Hz, and uses linear interpolation to output
+a PWM signal at 32.5 kHz.
```
dpgeorge, line 257: I think we should get rid of this linear interpolation, and just output the 7812 Hz signal directly.  I did some tests, and the audio quality is better *without* the interpolation.

docs/microphone.rst
```diff
@@ -70,11 +119,61 @@ Functions
     * **return**: a representation of the sound pressure level in the range 0 to
       255.
 
+.. py:function:: record(duration=3000, rate=7812, wait=True)
```
dpgeorge, line 122: Maybe we can remove this function altogether, and just have `record_into()`.  That way memory management is explicit, it avoids situations where there's a lot of heap fragmentation, and also avoids any difficulties with this function returning an `AudioFrame` object that is growing over time.

docs/microphone.rst
```diff
@@ -70,11 +119,61 @@ Functions
...
+
+    :param duration: How much time to record in milliseconds.
+    :param rate: Number of samples to capture per second.
+    :param wait: When set to ``True`` it blocks until the recording is
+        done, if it is set to ``False`` it will run in the background.
+    :returns: An ``AudioFrame`` with the sound samples.
 
-Example
-=======
+.. py:function:: record_into(buffer, rate=7812, wait=True)
```
dpgeorge, line 142: Perhaps this function can set some new internal state in an `AudioFrame` object which indicates the length of the recording (as opposed to the total allocated length of the buffer).  Then `audio.play()` would use this state to only play the amount that was recorded.

Could then add a method to get/set this length, eg `AudioFrame.get_recording_length()`, and `AudioFrame.set_recording_length()`.

docs/audio.rst
microbit-carlos: So, as a static method that would affect all AudioFrame playback instead of individual instances?

I agree that having it as part of `AudioFrame` makes it a lot more clear that it only affects playback of this type, but might be confusing when accessing the function via instances:

```python
frame_one = AudioFrame(...)
frame_two = AudioFrame(...)
frame_one.set_rate(8000)
frame_two.set_rate(8000)
...
while audio.is_playing():
    frame_one.set_rate(scale(accelerometer values...))

# At this point, I'd expect playback for frame_two to still be 8000?
audio.play(frame_two) 
```

Would it'd be more intuitive if each instance hold its playback rate?

docs/audio.rst
microbit-carlos: I think in this case, similar to the previous comment about having an `AudioFrame.set_rate()`, if `rate` was added to the constructor then we'd need to have `set_rate()` as a method changing each instance value, instead of a class static function.

I do like the idea to be able to make the arguments of the `AudioFrame` constructor simpler to work out if the user just wants a specific "size in time".

What would happen if the constructor is provided an argument for `size` and for `duration`? Would that throw an exception?

docs/audio.rst
microbit-carlos: Sounds good, in that case let's do that and remove this from the docs 👍

docs/microphone.rst
microbit-carlos: Yes, I completely agree here, playing with the current branch and recording a couple of seconds into a 5 second "buffer" results in a few seconds of "silent playback".

Would `record_into()` then also use this marker to continue recording where it left off?

docs/microphone.rst
microbit-carlos: Sorry, I forgot the context in which AudioFrame would be growing. I take it that was the case when `wait=False`, but if we have to have a known/default value for `duration` and `rate`, wouldn't the AudioFrame be allocated at the start before the recording occurs?

I still quite like the idea of being able to do `audio.play(microphone.record(duration=3000))`.
Would it be simpler if `wait` was removed from this version and only available in `record_into()`?

docs/audio.rst
dpgeorge: > So, as a static method that would affect all AudioFrame playback instead of individual instances?

Yes, as a static method calling it once affects all AudioFrame instances.

If we went this way, probably best to document it as `AudioFrame.set_rate(...)` and never refer to it as `frame.set_rate(...)`.  Then it's clear it sets the global rate for all audio frames.

> Would it'd be more intuitive if each instance hold its playback rate?

I'm not sure... maybe?  But then it would be an instance method and you would always call `frame.set_rate(...)` to set the playback rate of that instance.

I think that's possible to implement.

docs/audio.rst
dpgeorge: Note that there's a difference between record and play rates.  The rate here is related to the recording rate, although the actual recording rate is set when you call `microphone.record_into()`.  And that's different again to the playback rate set by `.set_rate()`.



> What would happen if the constructor is provided an argument for `size` and for `duration`? Would that throw an exception?

Yes it could throw an exception.

docs/microphone.rst
dpgeorge: > I take it that was the case when `wait=False`, but if we have to have a known/default value for `duration` and `rate`, wouldn't the AudioFrame be allocated at the start before the recording occurs?

When `wait=False` the returned-AudioFrame will be gradually filling up.  If we go with the current implementation where the whole buffer is preallocated at the start of the recording, then I guess it's not too bad, the user just sees blank data that's gradually filling up.

So, we can keep this function.  And even support `wait=False`.  Under the hood it will simply allocate an AudioFrame of the desired size, then pass it to `record_into()`.

docs/microphone.rst
dpgeorge: > Would `record_into()` then also use this marker to continue recording where it left off?

I think that would be confusing.  It would require the user to explicitly reset the marker to the beginning to reuse the buffer.  Maybe instead `record_into()` could have an additional argument which lets the user specify the starting location in the buffer.

docs/microphone.rst
```diff
@@ -70,11 +117,59 @@ Functions
...
+        done, if it is set to ``False`` it will run in the background.
+
+.. py:function:: is_recording()
+
+    :returns: ``True`` if the microphone is currently recording sound, or
+      ``False`` otherwise.
+
+.. py:function:: stop_recording()
+
+    Stops an a recording running in the background.
```
microbit-matt-hillsdon, line 155:
```suggestion
    Stops a recording running in the background.
```

Fixing typo but that also got me wondering what happens if you record more than once with wait=False (before the first has finished). 
Do they all work? If so does this function stop them all?

dpgeorge: Only one can be recorded at once, so a previously running recording is aborted at the point you call `record()` or `record_into()` again (as though you called `stop_recording()` first).

docs/audio.rst
microbit-grace, line 1: Hi, is there an `audio.sound_level()` doc missing?
cc: @microbit-matt-hillsdon

microbit-carlos: sound_level is already in the docs:
https://github.com/bbcmicrobit/micropython/blob/v2-docs/docs/microphone.rst?plain=1#L68
There is also a proposed sound_level_db not yet implemented:
- https://github.com/bbcmicrobit/micropython/pull/803

microbit-grace: Ah yes, there is one for `microphone`, but is there one for `audio` as well? I think I saw it being mentioned here (https://github.com/microbit-foundation/micropython-microbit-v2/pull/163) that there is an `audio.sound_level()`

microbit-matt-hillsdon: microphone.sound_level is already in the docs but audio.sound_level is not. audio.sound_level is new on the audio-recording branch and appears to be the _output_ audio level but is not documented here. I can't see any discussion of it beyond the commits that implement it.

docs/microphone.rst
```diff
@@ -70,11 +117,59 @@ Functions
     * **return**: a representation of the sound pressure level in the range 0 to
       255.
 
+.. py:function:: record(duration=3000, rate=7812)
```
microbit-matt-hillsdon, line 120: Based on the implementation on the audio-recording branch, there isn't a default for duration. The examples also all pass it explicitly. If we're happy with that then this should likely be amended to:

```suggestion
.. py:function:: record(duration, rate=7812)
```

docs/microphone.rst
microbit-carlos: Discussion:
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/197

docs/audio.rst
microbit-carlos: Yes, I missed this one, thanks!
Added in 87d7f75327d84afe790fdea8a88f349fdd373fe7.

docs/microphone.rst
microbit-carlos: This issue to discuss some implementation enhancements would cover this as well:
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/198

docs/audio.rst
microbit-carlos: Discussion about specifying the size of an `AudioFrame`:
- https://github.com/microbit-foundation/micropython-microbit-v2/issues/193

microbit-carlos:
@dpgeorge the docs have been updated with the conclusion from https://github.com/microbit-foundation/micropython-microbit-v2/issues/205#issuecomment-2221716206.

One thing I've changed that we haven't discussed before was to remove the `rate` argument from `microphone.record_into()`, as the rate can be set in the input AudioRecording/AudioTrack, and without the argument there isn't any ambiguity as to what takes precedence.

docs/audio.rst
```diff
@@ -215,6 +244,153 @@ Sound Effects Example
...
+  type like a ``bytearray``. It's similar to a
+  `memoryview <https://docs.micropython.org/en/v1.9.3/pyboard/reference/speed_python.html#arrays>`_
+  and it can be used to easily chop a portion of the audio data in the
+  ``AudioRecording`` or to modify its contents.
+
+AudioRecording
+--------------
+
+.. py:class::
+    AudioRecording(duration, rate=7812)
```
dpgeorge, line 268: Shall we try with a default rate of 11000?

docs/audio.rst
```diff
@@ -215,6 +244,153 @@ Sound Effects Example
...
+        :param sample_rate: The sample rate to set.
+
+    .. py:function:: get_rate()
+
+        Return the configured sampling rate for this
+        ``AudioRecording`` instance.
+
+        :return: The configured sample rate.
+
+    .. py:function:: copy()
```
dpgeorge, line 295: Do we need this method?  The Python way would be to use the constructor:
```py
new_recording = AudioRecording(old_recording)
```
(I can't remember if we discussed this or not.)

docs/audio.rst
```diff
@@ -215,6 +244,153 @@ Sound Effects Example
...
+
+        :returns: a copy of the ``AudioRecording``.
+
+    .. py:function:: track(start_ms=0, end_ms=-1)
```
dpgeorge, line 299: Are these keyword-only arguments, or positional?

Ie do we allow `recording.track(123)` or require `recording.track(start_ms=123)`?

docs/audio.rst
```diff
@@ -215,6 +244,153 @@ Sound Effects Example
...
+    The ``AudioTrack`` object points to the data provided by the input buffer,
+    which can be an ``AudioRecording``, or a buffer-like object like a
+    ``bytearray``.
+    It also contains its own sampling rate, so multiple ``AudioTrack``
+    instances pointing to the same buffer can have different rates and won't
+    affect the rate of the original buffer.
+
+    :param buffer: The buffer containing the audio data.
+    :param rate: The sampling rate at which data will be stored
+        via the microphone, or played via the ``audio.play()`` function.
```
dpgeorge, line 338: If you pass in an AudioRecording as the buffer and don't specify the rate, should it take the rate of the AudioRecording?

dpgeorge: If the answer to that is "yes" then the signature would need to be:
```py
AudioTrack(buffer, rate=None)
```
and `None` means:
- if buffer is an AudioRecording (or AudioTrack?) then take the rate from that
- otherwise rate is default 7812 (or 11000?)

docs/audio.rst
```diff
@@ -215,6 +244,153 @@ Sound Effects Example
...
+
+        :return: The configured sample rate.
+
+    .. py:function:: copyfrom(other)
+
+        Overwrite the data in this ``AudioTrack`` with the data from another
+        ``AudioTrack``, ``AudioRecording`` or buffer-like object like
+        a ``bytes`` or ``bytearray`` instance.
+
+        :param other: Buffer-like instance from which to copy the data.
```
dpgeorge, line 360: If the other is smaller than self, should the remaining bytes be left untouched?  Probably.

docs/audio.rst
```diff
@@ -215,6 +244,153 @@ Sound Effects Example
...
+
```
dpgeorge, line 361: Will need to document access via subscript and slicing, and `len(audio_track)` operator.

docs/audio.rst
microbit-carlos: Yes, we looked at picking between a copy constructor and a `copy()` method, in the end we decided to go with the method because the micro:bit API hasn't used copy constructors before and the `copy()` method was already a pattern used in classes like `Image`.

docs/audio.rst
microbit-carlos: I think it's fine to leave them as positional, users can still use keywords for clarity.

docs/audio.rst
microbit-carlos: Yes, I'll add to the docs as well.

docs/audio.rst
microbit-carlos: Is that commonly documented via the dunder methods?
That might  be a bit hard to understand for some of our audience, maybe we should add an explanation before or after the class documentation, or the in the class description?

docs/audio.rst
microbit-carlos: Right, originally I was thinking to force the default rate, as it's also a simpler implementation and explanation. But after some thought I think most users would expected the rate to be copied.  I'll update the docs to the `rate=None` method.

docs/audio.rst
microbit-carlos: Sounds good, will update the docs.

docs/audio.rst
dpgeorge: OK, I've implemented `AudioRecording.copy()`.

(There's actually also the existing `SoundEffect.copy()`, so that's nice and consistent.)

docs/audio.rst
microbit-matt-hillsdon: I don't think they are positional at the moment:

```python
>>> audio.AudioRecording(100).track(0)
TypeError: extra positional arguments given
```

docs/audio.rst
microbit-matt-hillsdon: Would be great to document these somehow as otherwise it's trial and error/reading the C.

I think for AudioTrack we have:

```
    @overload
    def __getitem__(self, i: int) -> int: ...
    @overload
    def __getitem__(self, s: slice) -> AudioTrack: ...
    def __setitem__(self, i: int, x: int) -> None: ...

    def __add__(self, v: AudioTrack) -> AudioTrack: ...
    def __iadd__(self, v: AudioTrack) -> AudioTrack: ...
    def __sub__(self, v: AudioTrack) -> AudioTrack: ...
    def __isub__(self, v: AudioTrack) -> AudioTrack: ...
    def __mul__(self, v: float) -> AudioTrack: ...
    def __imul__(self, v: float) -> AudioTrack: ...
 ```

So slice access but not slice assignment.

File: docs/audio.rst [hunk 14]
```diff
@@ -17,14 +17,31 @@ There are three different kinds of audio sources that can be played using the
 
 1. `Built in sounds <#built-in-sounds-v2>`_ (**V2**),
```
microbit-matt-hillsdon, line 18: Just above this list of 5 things we say it's a list of 3 things!

docs/audio.rst
dpgeorge: I've now made them positional.

docs/audio.rst
microbit-carlos: Thanks Matt! Resolved in b129d161192a640b0d034e9071e3b94af6731601.

PR #803, opened 2024-03-21
docs: Add `microphone.sound_level_db()` function.

microbit-carlos:
https://microbit-micropython--803.org.readthedocs.build/en/803/microphone.html#microbit.microphone.sound_level_db

### DIFF
--- a/docs/microphone.rst
+++ b/docs/microphone.rst
@@ -91,6 +91,10 @@ Functions
 
     :return: A representation of the sound pressure level in the range 0 to 255.
 
+.. py:function:: sound_level_db()
+
+    :return: A representation of the sound pressure level in decibels (dB)
+        in the range of 52.0 to 110.0 dB.
 
 Example
 =======
--- a/docs/microbit_micropython_api.rst
+++ b/docs/microbit_micropython_api.rst
@@ -108,6 +108,8 @@ The Microphone is accessed via the `microphone` object::
     set_threshold(128)
     # Returns a representation of the sound pressure level in the range 0 to 255.
     sound_level()
+    # Returns a representation of the sound pressure level in decibels (dB).
+    sound_level_db()
 
 Pins
 ----


docs/microphone.rst
```diff
@@ -70,6 +70,9 @@ Functions
     * **return**: a representation of the sound pressure level in the range 0 to
       255.
 
+.. py:function:: sound_level_db()
+
+    * **return**: a representation of the sound pressure level in decibels (dB).
```
dpgeorge, line 75: Maybe it's worth specifying the range of dB values that will be returned?  That could be useful to know how loud/quite sounds it can respond to, and also to perform some scaling on the number to make it suitable, eg, for plotting to the display.

microbit-carlos: @JohnVidler do you know what the min and max value for LevelDetectorSPL are for dB?

JohnVidler: 35-100dB, by default, although they are defined in `LEVEL_DETECTOR_SPL_8BIT_000_POINT` and `LEVEL_DETECTOR_SPL_8BIT_255_POINT` if you want to future proof against wider ranges on future devices

microbit-carlos: I thought those were used only to calculate the 8-bit conversion. Are the min and max values from  `LevelDetectorSPL::getValue(LEVEL_DETECTOR_SPL_DB)` also clamped to those?

microbit-carlos: A quick test printing `uBit.audio.levelSPL->getValue(LEVEL_DETECTOR_SPL_DB)` shows we can get values over 100. @JohnVidler could you check what are the theoretical min and max?

microbit-carlos: Added in https://github.com/bbcmicrobit/micropython/compare/751913410f141d536d9fc33912e0eef502636afe..477c214ed952927cb8ca3e1a331144b461f28036.

